PR #5, opened 2016-01-20
Datashader Notebook Example

brendancol:
- [x] integrate `ImageRGBA`
- [x] update with notebook kernal callbacks
- [x] fix `dh=[ymax - ymin]` issue

### DIFF
--- a/examples/datashader.ipynb
+++ b/examples/datashader.ipynb
@@ -1,0 +1,87 @@
+{
+ "cells": [
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {
+    "collapsed": false
+   },
+   "outputs": [],
+   "source": [
+    "import numpy as np\n",
+    "\n",
+    "from bokeh.plotting import figure, output_notebook, show\n",
+    "from bokeh.models import ColumnDataSource, CustomJS\n",
+    "from bokeh.io import push_notebook\n",
+    "\n",
+    "output_notebook()\n",
+    "        \n",
+    "jscode=\"\"\"\n",
+    "        // Define a callback to capture errors on the Python side\n",
+    "        function callback(msg){\n",
+    "            console.log(\"Python callback returned unexpected message:\", msg)\n",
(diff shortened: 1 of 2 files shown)


jbednar:
Looks good.  I'm eager to see the packaged-up version, with the JS tucked away inside a class to leave what the user actually does need to mess with.  Looking at the code for create_image and relating that to the Bokeh code above, it looks like there is lots of stuff to do with x and y ranges.  Since create_image is something that we know is going to be called by Bokeh, can we put some bokeh-specific code that will get the ranges out of the Bokeh plot (instead of explicitly putting in p.height, etc. inside create_image)?  It Seems like there ought to be a way to simplify that...

jcrist:
Is there a way to not embed bokehjs in the notebook, but instead have it pull from a cdn? This notebook is 78,000 over lines long!

jcrist:
```python
from bokeh.io import output_notebook
from bokeh.resources import CDN
output_notebook(CDN)
```

Could you add this, rerun the notebook, and squash out the previous commits? Trying to keep down repo bloat from large blobs. Not a super big deal, just a preference.

jbednar:
Changing it to use CDN by default sounds reasonable, but my personal preference is just not to commit any notebooks that have not been cleared, so that diffs work across github versions.  So I'd vote for clearing, committing, and *then* squashing previous commits.

brendancol:
I've add the notebook example.  The goal of the notebook is to tell the story of datashader starting with showing the problem of over-saturation when plotting too many points:

A few points...
screen shot 2016-01-28 at 9 55 50 pm

Too many points..
screen shot 2016-01-28 at 9 56 00 pm


Datashader to the rescue...
screen shot 2016-01-28 at 9 56 12 pm

jbednar:
Cool, thanks!  I'll merge when it's on a new PR; should be easy to make one since it's just one, new file.  Or else just add the new file directly, with no PR.

brendancol:
merged...

Issue #11, opened 2016-01-21
Map labels, map opacity in dashboard.py


jbednar:
It would be good to have a toggle button to enable or disable a map labels tile layer, to help people get oriented.

Having an opacity slider for the map tiles would also be good, since the map isn't always useful.

brendancol:
I've superseded the other @jbednar PR with this one: https://github.com/bokeh/datashader/pull/47

jcrist:
Fixed by #47, closing.

PR #17, opened 2016-01-21
Added basemap opacity slider

jbednar:
To get some experience with the new dashboard.py, I added a basemap opacity slider alongside the data opacity slider (as requested in issue #11).  This was relatively simple to do, though I wish it could be more atomic, with changes required at only a single location in the code, instead of three locations spread out.  I think that making that work would require some fundamental changes to Bokeh's architecture, which I'd be happy to discuss with Bokeh devs at some point, but not as a priority right now.  In any case, it's already much cleaner than having to switch to JavaScript! 

The result works fine, but it's very wasteful of screen space, even after packing the two opacity sliders vertically so that they would fit on the page:

image

Is there any way to change the CSS (presumably) to make the packing much tighter?  There are other controls that need to be added, such as a checkbox for turning labels on and off, and the controls are already taking up 1/3 of the page, mostly with whitespace!

### DIFF
--- a/examples/dashboard.py
+++ b/examples/dashboard.py
@@ -168,10 +168,14 @@ class AppView(object):
         basemap_select = Select.create(name='Basemap', value='Toner', options=self.model.basemaps)
         basemap_select.on_change('value', self.on_basemap_change)
 
-        opacity_slider = Slider(title="Opacity", value=100, start=0, end=100, step=1)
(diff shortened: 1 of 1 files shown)


brendancol:
Closing the PR because it has been superseded by https://github.com/bokeh/datashader/pull/47

PR #25, opened 2016-01-26
Task/dashboard example config

brendancol:
adds yml config file for `nyc_taxi` dataset. 

To run: `python dashboard.py --config nyc_taxi.yml`


jbednar:
Looks good; I'm happy to merge it whenever you say it's ready.  I think it's much more readable and maintainable with the dataset-specific stuff moved to the .yml file. Two small points:

- Should the "plots:" section of the config file be changed to "projections:"?  That bit only specifies which fields will be used to make the projection, not other things that affect the plot, so I'd think projection would be more accurate.  That said, for plots it's clear that there will be multiple choices set up that each result in different plots, whereas that may not be clear for projections.

- It's a shame that Pandas apparently requires the filetype to specified in the read call (pd.read_csv).  Seems like with Pandas having so many different file reading functions, there would be a wrapper that dispatches to the correct one based on the filetype, but I can't find one. :-(  So this example will only work with CSV files for now.  Extending it to work with others is not difficult, of course, but it would add complexity that we really want to put off onto some other library, not this example!

jbednar:
Brendan and I just discussed this, and agree that "axes:" is much clearer than either "plots:" or "projections:".

jbednar:
Wouldn't the initial_extent depend on the value of "axes:"?  In this particular case, both pickups and dropoffs cover the same geographical region, but in the general case with arbitrary columns projected onto the plot, the axis ranges should be different for each plot...

jcrist:
Can you move the download script back? I was planning on having that be for all example data, not just this example.

jbednar:
Do you mean, you would have that file be general, handling different downloads via e.g. passing a URL, or do you mean that information about all downloads would be put into that single file?  The current download script seems heavily tied to Taxi, doing various specific processing, so it might be better to have a high-level script that the user calls once that simply calls specific scripts like this, one per dataset...

jcrist:
The current one is heavily tied to taxi, just because that's the only dataset we have. Only 16 of those lines are specific to the taxi dataset though. I think having one download script that is just appended to with more example datasets as needed is a fine solution. It also makes more sense, as we could use the taxi dataset for more than one example (and we should, rather than pollute the users disk with too many datasets). Bokeh does the same thing.

jbednar:
Well, there are only 37 non-whitespace lines in the download script, which leaves 21 *not* having to do with the taxi data, and 16 having to do with it, by your count.  So it's largely a wash.  But I do agree that we'll want that dataset for use in other examples.

brendancol:
moved download script and repointed yml config appropriately

brendancol:
regarding `initial_extent`, these value ARE dependent on the first set of axes in the `axes` list. Putting extents there seem nice because i could then zoom to different location when axes are changed...not sure if that would be a desired feature, but seems an appropriate property for each set of axes.  May that could come in a separate PR

jbednar:
Sounds good.  A separate PR would be fine for that.

PR #29, opened 2016-01-29
Aggregate redesign & general improvements

jcrist:
Previously we were outputting raw dynd arrays after the aggregation step. This was not a good design decision for a few reasons:

1. Enforced array storage for aggregates. Currently all aggregates are arrays, but in general that's limiting.
2. Dynd is currently buggy (but at least in predictable ways). Without encapsulating these arrays in our own objects, we exposed users to issues in dynd.
3. Prevented being able to (easily) attach metadata and relevant methods to the aggregates.

Downsides are that we have to reimplement operations to make aggregates feel natural to work with. However, as most aggregate level operations are simple, aligned elementwise operations, these are fairly trivial to write.

In the act of making this change, I found lots of other things that were related that needed fixing. This PR:

- Defines 3 new aggregate types, and provides common operations on them, and updates the library to work with them. 
- Removes the previous requirement of having aggregates be fields in a larger enclosing record aggregate. Now you can do multiple aggregations by passing in a `summary` object explicitly, or a single aggregate alone (e.g. `mean('field')`).
- Adds `Axis` types, initially defining both `LinearAxis` and `LogAxis`. These are attached to an aggregate upon creation, and serve as relevant metadata describing its location in space.

Note that pretty much all of this is still undocumented. However, I'd like to handle the documentation for the whole library in a separate PR.

### DIFF
--- a/datashader/__init__.py
+++ b/datashader/__init__.py
@@ -3,7 +3,8 @@ from __future__ import absolute_import
 __version__ = '0.0.1'
 
 from .core import Canvas
-from .aggregates import count, sum, min, max, mean, std, var, count_cat
+from .reductions import (count, sum, min, max, mean, std, var, count_cat,
+                         summary)
 
 # Needed to build the backend dispatch
 from .pandas import *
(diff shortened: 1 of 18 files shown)


jbednar:
Ok, good.  I'll go ahead and merge so that I'll have these changes when working on the .ipynb file today; further fixes can be done separately if needed.

PR #31, opened 2016-01-29
Renamed pickup,dropoff_longitude,latitude to _x,_y

jbednar:
Was confusing, because they are now in Web Mercator meters-based coordinates, not latlon.


jbednar:
Note that anyone who has downloaded the taxi data will need to download again (or munge their own copy to do the renaming manually).

PR #33, opened 2016-01-29
Implement `where` for aggregates

jcrist:
Implements `where` for both scalar and categorical aggregates. Syntax:

```python
# `agg` where cond is True, NA everywhere else
agg.where(cond)
# `agg` where cond is True, 2 everywhere else
agg.where(cond, 2)
# `agg` where cond is True, other_agg everywhere else
agg.where(cond, other_agg)
```

Allows for filtering based on elementwise conditionals in aggregates.

### DIFF
--- a/datashader/aggregates.py
+++ b/datashader/aggregates.py
@@ -127,33 +127,7 @@ class ScalarAggregate(Aggregate):
         """
         if not isinstance(cond, ScalarAggregate):
             raise TypeError("cond must be instance of ScalarAggregate")
-        elif not isboolean(cond.dshape):
-            raise TypeError("cond must be a boolean aggregate")
-        _validate_aligned(self, cond)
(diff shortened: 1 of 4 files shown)


jbednar:
Seems very useful.  There's a lot of duplication between the two where() implementations; can that be avoided?  And it's a shame that something boiling down to np.where would take so much sanitizing code to work, but that's presumably unavoidable.

jcrist:
Reduced code duplication. In general, a lot of the ugliness of all this is due to dynd lack of features/stability. I may nuke it from the codebase soon in favor of some carefully managed numpy buffers. That's really all dynd is used for at this point - a buffer of bytes that sometimes crashes if you try to do stuff with it.

jbednar:
That helps, thanks!  One copy of ugly code is better than two! :-)

PR #36, opened 2016-01-30
Created module for IPython kernel callbacks

philippjfr:
As per request by @jbednar I used his initial attempt as a starting point to move the IPython kernel callback in the ``nyc_taxi`` demo notebook into a module so the notebook itself isn't littered with JS code. 

The example from the nyc_taxi demo looks like this now:

```python
from datashader.callbacks import IPythonKernelCallback

def create_image(p, ranges, agg_fn=ds.count):
    x_range, y_range = ranges['x_range'], ranges['y_range']
    h, w = ranges['h'], ranges['w']
    cvs = ds.Canvas(plot_width=w, plot_height=h, x_range=x_range, y_range=y_range)
    agg = cvs.points(df, 'pickup_x', 'pickup_y', agg_fn('passenger_count'))
    pix = tf.interpolate(agg, (255, 204, 204), 'red', how='log')
    dh = y_range[1] - y_range[0]
    dw = x_range[1] - x_range[0]
    p.image_rgba(image=[pix.img], x=x_range[0], y=y_range[0], dw=dw, dh=dh, dilate=False)

p = base_plot()
IPythonKernelCallback(p, create_image, throttle=500, agg_fn=ds.count)
```

We pass the ``Figure`` (p), the Python call, an optional throttling value and an agg_fn to the IPythonKernelCallback. Internally the following happens:

1. The IPythonKernelCallback instance registers itself on the class with a uuid as the key.
2. We generate a string command to execute on the Python side, which a) imports the class and b) looks up the callback instance on the class using the uuid and c) calls the ``update_image`` method on the callback instance with the stringified json of the current ranges.
3. Using a string format we interpolate the plot_id, and the cmd into a template and wrap it in ``CustomJS``.

The Python callback itself accepts two args and kwargs:

1. The plot to render to.
2. The dictionary of stringified json data.
3. Any additional keyword arguments that were passed to the IPythonKernelCallback instance.

This works pretty well in the notebook, but it does have several limitations/drawbacks:

1. This still creates a new image_rgba glyph on every callback, it might be more efficient to update the datasource on an existing glyph.
2. Currently I bypass the ``show`` function, handle the comms object manually and use ``_repr_html`` to display it. I thought this would be fixed by https://github.com/bokeh/bokeh/pull/3681?
3. I've wrapped the kernel callbacks in ``_.throttle`` and added them to ``Bokeh._throttle``, a new object I created. They should presumably be stored somewhere more appropriate.
4. This only works for standalone figures currently, any layouts likes vplot, hplot or gridplot will have issues when attempting to look them up via ``Bokeh.index``, instead the id of the root plot should be sent and the plot looked up on the document.


jbednar:
That looks great!  Is there any way to eliminate the special call to create_image above?

philippjfr:
Yes, I could look up the initial ranges and plot width/height on the plot and execute the callback once on instantiation.

jbednar:
That would be great; much less confusing for users.

jbednar:
Also, it looks like "dsplot =" can be eliminated, as that attribute isn't actually used.  That way it will be clear that instantiating the IPythonKernelCallback class is what installs it.

jbednar:
BTW, should this work when there are multiple such cells?  It doesn't seem to; if I duplicate the last cell in the notebook, only the first such plot seems to regenerate the image_rgba.

philippjfr:
As I noted above:

```
Currently only the most recently executed cell gets updated although old plots still create kernel events on zooming, I think this could be improved. I thought this would be fixed by bokeh/bokeh#3681, maybe that hasn't made it's way into the latest dev release yet?
```

I just fixed a bug that would have prevented this from working by passing the plot instance into the callback but it still doesn't seem to do anything. Any idea when the last bokeh dev release was?

jbednar:
According to https://github.com/bokeh/bokeh/releases, 0.11.1dev3 was 2 days ago, at rev 1673e48.

>Currently only the most recently executed cell gets updated although old plots still create kernel events on zooming, I think this could be improved. 

Ah, you were talking about Shift-enter executing, i.e. executing Jupyter cells?  That's not the behavior I was seeing; I duplicated the last cell, and the first one was the one updating, not the second, which was only executed once I duplicated them.  Actually, now that I think about it, I don't know *which* one was executed last, since they look identical; maybe the one lower on the page was actually the original, not the copy.  

So ok, the last one executed is the one that gets updates?    But that's not how HoloViews works, right?  How did you get it to work for any cell in HoloViews?

philippjfr:
>So ok, the last one executed is the one that gets updates? But that's not how HoloViews works, right? How did you get it to work for any cell in HoloViews?

In HoloViews I manually handle the Document creation, can do the same here I think.

jbednar:
That sounds good; it would be awkward if only the last cell could be interacted with (e.g. Run All wouldn't be very useful then!)

jbednar:
And the new commits look great, thanks!

jbednar:
Although I'm really surprised -- I thought that removing the explicit call to create_image would fix the problem with the incorrect cropping on the initial plot (which is aligned properly, just covers too little area), since the first image would then be generated by the callback just like the rest of the images are.  But that doesn't seem to be the case; the initialization behavior hasn't changed (though the code looks much better!).

philippjfr:
I'm not quite sure what's happening there either. I'm assuming the tile renderer or something else ends up setting a different range to what was specified initially. I've also run into problems with supporting multiple plots at once, in HoloViews we don't use ``show`` and all the magic that entails but I thought that they fixed this with the PR referenced above. Will have to dig into what's happening there tomorrow, going to bed now :)

jbednar:
Thanks so much; this has really helped already!  Brendan mentioned that putting the tiles in or out does have an effect, though I think he said it actually improved something he was looking at, surprisingly.  I can actually remove the tiles from this example, for now; they aren't crucial to include (though they do help to verify that we're doing scaling properly, since the data aligns closely with the geographic tiles).

philippjfr:
I've added manual handling of the comms channels, which means I've had to hook into some of the low level private classes in bokeh because I couldn't get it working using ``show``. This means I've used ``_repr_html_`` to display the plot and am bypassing the usual display machinery entirely. In future it would be better to use ``show`` directly but unless I'm missing something that's not currently possible.

philippjfr:
Also some random feedback, datashader should also have some way of setting or automatically determing a minimum bin size. When zooming in all the way to the street level the bins appear sparse but still appear in a regular, presumably the GPS data already has some fixed sampling and it would be nice if datashader didn't oversample. I've already got some code like that for rasterization callback in the Pepsi project.

jbednar:
Thanks!  While you still remember it, can you explain what the problems with show() were, either here or in a bokeh issue?

jbednar:
Is it ok to merge?

philippjfr:
The recent debounce fix I introduced makes interaction a lot more smooth but it sometimes goes out of sync and ignores the last event. Could you give it a try with and without these lines:

```
if self.debounce:
            self.debounce = False
            return
        else:
            self.debounce = True
```

and see what you think?

philippjfr:
The issue with ``show`` is that although they fixed it recently I don't seem to be able update multiple plots even if I attach them to separate documents.

philippjfr:
>And is your code to avoid oversampling automatic or a manual setting?

Just checked and I ended up going with a fixed sampling because the peak finding code because in that application we had worked out the sampling density.

jbednar:
I tried it, and it does seem to be smoother with the debouncing.  But in general this code should work with non-fixed-aspect-ratio plots, where someone could e.g. scale one axis rather than the other, which would only generate one event.  And presumably a box zoom would only generate one event?  So I'm not sure it's safe to debounce like this.

 I wonder if there is some more subtle way to drop duplicate events other than simply skipping every other one, e.g. to ignore one if there was one just before.  What we've done in other cases is to have the most recent one interrupt the one before it, so that it's always only processing one event, though of course that has issues with never terminating if events keep coming in (which seemed unlikely).  But that's something we did at the back end, i.e. within the callback,  not at the level of the machinery doing the callback.

philippjfr:
>What we've done in other cases is to have the most recent one interrupt the one before it, so that it's always only processing one event, though of course that has issues with never terminating if events keep coming in (which seemed unlikely).

I tried that, it still seemed to be generating too many events for some reason. I've replaced that with a call to ``_.throttle``, which ensures events are only sent every x milliseconds. underscore.js also has a ``_.debounce`` function, which could be used to capture only the final event. [Here](http://drupalmotion.com/article/debounce-and-throttle-visual-explanation)'s a good visual breakdown of the two.

philippjfr:
It's pretty responsive now and no longer swallows events like the previous approach, you can merge at any time.

jbednar:
Great!  Merging now.  I'll try to discuss this with Fabio and Damian to see if it makes sense to use in Bokeh.

damianavila:
Interesting concept :wink:

PR #37, opened 2016-02-01
Add flatten method for categorical aggregate.

jcrist:
Allows converting CategoricalAggregate to ScalarAggregate. Stopgap api, probably will change in the future to something more consistent with numpy, but works fine for now.

### DIFF
--- a/datashader/aggregates.py
+++ b/datashader/aggregates.py
@@ -33,6 +33,14 @@ def _validate_aligned(a, b):
                                   "non-matching axis or shape")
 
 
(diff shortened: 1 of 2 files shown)


jbednar:
Looks good; I'm happy to have it merged (though with the usual sighing about the level of type-related verbiage required).  Is there a way to abstract the type-munging stuff into a base class, method, or separate object so that individual aggregation functions become more readable?  Right now I don't think a user would be able to look at one and have any confidence that they could write their own custom one.

jcrist:
Closing in favor of #48.

PR #38, opened 2016-02-01
Added DatashaderPipeline and updated examples

jbednar:
Adds a configurable pipeline object and examples of using it in the Jupyter notebook.  Should be ready to merge, based on the discussion today, but it would be nice if someone looked at how the agg_fn is specified and think about whether that's the most general way to do it.

### DIFF
--- a/datashader/pipeline.py
+++ b/datashader/pipeline.py
@@ -29,7 +29,7 @@ class Interpolate(param.Parameterized):
         Color string or tuple specifying the ending point for interpolation.""")
 
     how = param.Parameter(default="log", doc="""
-        Function object or string specifying how to map from a scalar into color space.""")
+        Function object or string specifying how to map a scalar into color space.""")
 
     def __call__(self, agg):
         return tf.interpolate(agg, self.low, self.high, self.how)
@@ -42,12 +42,12 @@ class DatashaderPipeline(param.Parameterized):
     effect of varying that element while keeping the rest of the pipeline
     constant.
 
-    The pipeline is roughly:
+    The supported pipeline is roughly:
 
-    1. create canvas of the requested size
+    1. create canvas of the requested size (in data space) and resolution
     2. aggregate using the specified x and y fields, aggregate field, and agg_fn
-    3. apply the specified transfer_fns, if any, in order
-    4. apply the specified color_fn to translate each resulting aggregate into a color
+    3. apply specified transfer_fns, if any, in order
+    4. apply specified color_fn to translate each resulting aggregate into a color
     5. return the result as an image
     """
 
@@ -86,7 +86,9 @@ class DatashaderPipeline(param.Parameterized):
         x_range, y_range = ranges['x_range'], ranges['y_range']
         h, w = ranges['h'], ranges['w']
 
-        cvs = core.Canvas(plot_width=w, plot_height=h, x_range=x_range, y_range=y_range)
+        cvs = core.Canvas(plot_width=w, plot_height=h,
+                          x_range=x_range, y_range=y_range)
+
         agg = cvs.points(ps.data, ps.x, ps.y, ps.agg_fn(ps.agg))
         for f in ps.transfer_fns:
             agg = f(agg)
@@ -94,4 +96,5 @@ class DatashaderPipeline(param.Parameterized):
 
         dh = y_range[1] - y_range[0]
         dw = x_range[1] - x_range[0]
-        plot.image_rgba(image=[pix.img], x=x_range[0], y=y_range[0], dw=dw, dh=dh, dilate=False)
+        plot.image_rgba(image=[pix.img], x=x_range[0], y=y_range[0],
+                        dw=dw, dh=dh, dilate=False)
--- a/datashader/pipeline.py
+++ b/datashader/pipeline.py
@@ -3,11 +3,11 @@ from __future__ import absolute_import, division, print_function
 """
 Declarative interface to Datashader.
 
-Provides a configurable pipeline that makes it simpler to specify
-individual stages independently from the others, Currently does not
-add any new underlying capabilities, with all functionality also
-available using the default imperative interface provided by other
-files.
+Provides a configurable pipeline that makes it more convenient to
+specify individual stages independently from the others.  Does not
+cover all possible datashader functionality, and does not add any new
+underlying capabilities; all functionality is also available using the
+default imperative interface provided by other files.
 """
 
 import param
@@ -36,12 +36,12 @@ class Interpolate(param.Parameterized):
         return tf.interpolate(agg, self.low, self.high, self.how)
 
 
-class DatashaderPipeline(param.Parameterized):
+class Pipeline(param.Parameterized):
     """
-    Configurable datashading pipeline for use with Bokeh.  Allows each
-    element of the pipeline to be specified independently without code
-    duplication, e.g. to show the effect of varying that element while
-    keeping the rest of the pipeline constant.
+    Configurable datashading pipeline.  Allows each element of the
+    pipeline to be specified independently without code duplication,
+    e.g. to show the effect of varying that element while keeping the
+    rest of the pipeline constant.
 
     Given a dataframe-like object df, the supported pipeline is roughly:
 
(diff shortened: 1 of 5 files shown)


datashader/pipeline.py
```diff
@@ -0,0 +1,97 @@
...
+         Color string or tuple specifying the starting point for interpolation.""")
+
+    high = param.Parameter(default="red", doc="""
+         Color string or tuple specifying the ending point for interpolation.""")
+
+    how = param.Parameter(default="log", doc="""
+         Function object or string specifying how to map from a scalar into color space.""")
+
+    def __call__(self, agg):
+            return tf.interpolate(agg, self.low, self.high, self.how)
```
jcrist: Overindented.

jbednar: Deliberately.

datashader/pipeline.py
```diff
@@ -0,0 +1,97 @@
...
+
+    color_fn = param.Callable(default=Interpolate(), doc="""
+        Function to convert a scalar value into a color.""")
+
+
+    def __call__(self, plot, ranges, **params):
+        """
+        Accepts a Bokeh plot and a viewport specified via a ranges dictionary
+        (which should contain x_range, y_range, h, and w).  Returns an image
+        rendered at the specified location, using the current parameter values.
```
jcrist: This line (and many others) contain trailing whitespace. There should be a way to configure your editor to remove these (I know it's doable in both emacs and vim, not sure what you use).

jbednar: I use emacs, but this code was written in the Jupyter notebook and then pasted into a file, and Jupyter doesn't check for that.  I'll fix it.

datashader/pipeline.py
```diff
@@ -0,0 +1,97 @@
...
+
+        cvs = core.Canvas(plot_width=w, plot_height=h, x_range=x_range, y_range=y_range)
+        agg = cvs.points(ps.data, ps.x, ps.y, ps.agg_fn(ps.agg))
+        for f in ps.transfer_fns:
+            agg = f(agg)
+        pix = ps.color_fn(agg)
+
+        dh = y_range[1] - y_range[0]
+        dw = x_range[1] - x_range[0]
+        plot.image_rgba(image=[pix.img], x=x_range[0], y=y_range[0], dw=dw, dh=dh, dilate=False)
```
jcrist: Please wrap lines longer than ~80 characters. Not super strict on this, but this is 96 characters long, and makes it hard to read on a split terminal.

datashader/pipeline.py
jcrist: Why? There's absolutely no reason to do this.

jbednar: Perhaps a more polite way to phrase your question is, "Oh, why do you do that -- I can't see the reason?".  In this case, the reason is that I find it far more readable to scan through a list of declared parameters that is nicely indented like this:

```python
low = param.Parameter(default="lightpink", doc="""
    Color string or tuple specifying the starting point for interpolation.""")

high = param.Parameter(default="red", doc="""
    Color string or tuple specifying the ending point for interpolation.""")
```

Than one where the doc strings and the code all get run together visually:

```python
low = param.Parameter(default="lightpink", doc="""
Color string or tuple specifying the starting point for interpolation.""")

high = param.Parameter(default="red", doc="""
Color string or tuple specifying the ending point for interpolation.""")
```

This convention is something that we developed gradually from working with declarative interfaces, where code is generally dominated by user-configurable parameters like these, making it important to have good formatting for such docstrings.  In this project, it's fine to remove those extra four spaces if people prefer, but there *is* a reason they are there.

jcrist: Sorry, that was rude, you're correct. Perhaps there's some confusing here; I'm not complaining about the docstrings indentation, those are fine. I'm complaining about the indentation in `__call__` (see the line above the first comment). Python's convention is to consistently indent blocks with 4 spaces (https://www.python.org/dev/peps/pep-0008/#indentation).

jcrist:
Overall this looks clean (besides my formatting nitpicks above). Could still use tests though.

In the future, (once this has gotten some use and a nice api is settled on), there should be a fairly easy way to autogenerate the declarative forms of these functions from the implementation. This is fine for now. One thing I'm *slightly* perturbed about is the addition of another dependency, but if it makes your code easier to work with that's fine. You'll need to add it to `.travis.yml`, `requirements.txt` and the `Readme`.

datashader/pipeline.py
jbednar: Ah, in that case you are correct.  The indentation got all screwed up when pasting from Jupyter, and I didn't catch that one.  Annoying!  I've fixed it now; sounds like we can agree on the whitespace convention.  Yes, there's absolutely no reason for indenting ``__call__`` like that.

jbednar:
I don't think there will be any way to generate the declarative forms from the implementation in general, because the necessary information isn't in the implementation.  The implementation doesn't specify the allowed ranges and types for the parameters, and once you change the implementation to do that, then you have something like the declarative form shown here, one way or another.  So I agree that there should just be one specification, but the only way I can imagine how to do that is by making the original specification for the operation into something that explicitly declares the information we need.

jbednar:
The extra dependency is worth thinking about, but there's no rush.  On the one hand, adding one pure-Python dependency available in conda, on top of the major and complicated dependencies that come from Numba, shouldn't matter much.   On the other, even though Param definitely makes working with this approach simpler, it actually doesn't affect the API, and we can easily simulate the same API just with more verbiage (and no type checking) without it.  So we can remove it if we decide to go with another parameter library or to leave things bare and unchecked.

datashader/pipeline.py
```diff
@@ -0,0 +1,97 @@
...
+        Color string or tuple specifying the ending point for interpolation.""")
+
+    how = param.Parameter(default="log", doc="""
+        Function object or string specifying how to map from a scalar into color space.""")
+
+    def __call__(self, agg):
+        return tf.interpolate(agg, self.low, self.high, self.how)
+
+
+class DatashaderPipeline(param.Parameterized):
```
jcrist: Perhaps just `Pipeline`? This is part of `datashader`, so I'm unsure if the extra qualifier is needed.

jbednar: Depends -- are you ok with importing it into ``__init__.py``?  If you are, then people can refer to it as `ds.Pipeline` or `datashader.Pipeline`, which will be nice and clear about where this object comes from.  Otherwise people are likely to import it all on its own (as currently done in the notebook), making it helpful to have the extra qualifier.  I'd vote for putting it in ``__init__.py`` and removing the qualifier, myself.

datashader/pipeline.py
```diff
@@ -0,0 +1,97 @@
...
+        Name of the field in the supplied data object to use for the x coordinate.""")
+
+    y = param.String("y", doc="""
+        Name of the field in the supplied data object to use for the y coordinate.""")
+
+    agg = param.String("count", doc="""
+        Name of the field in the supplied data object to use for aggregation.""")
+
+    agg_fn = param.Callable(reductions.count, doc="""
+        Function for aggregating pixel-bin contents into a scalar.""")
```
jcrist: After some thought, a more general spelling would possibly be:

```
pipeline = Pipeline(glyph=ds.Point('some_col', 'other_col'), agg=ds.mean('col'), transfer=Interpolate())
```

The reason for this is that both `Glyph` and `Reduction` objects are already declarative, and are used internally with the other api. This would allow generality for a single dataset, and still be declarative. I'm not sure if there's a better way to spell the transfer functions in a declarative fashion though.

datashader/pipeline.py
jcrist: That'd be fine by me.

jcrist:
Another passing design thought, this is basically a parametrized scatter plot. Perhaps this should be designed to be analogous to the `bokeh.charts.Scatter` object (http://bokeh.pydata.org/en/latest/docs/user_guide/charts.html#scatter-plots)? Could even bake in ipython notebook support into it.

jbednar:
Good point.  I had considered trying to make the interface match the scatterplot interface already in the example notebook, but you're right -- the interface I came up with is actually quite close to the Charts version of the Scatterplot.  That's worth looking into, because it would make for a clean transition into data shading, and a simple story.  But I think I'll go ahead and merge this and come back to that idea later, because of needing to have something together by tomorrow.

datashader/pipeline.py
jbednar: Yes, that's the direction I was considering taking it.  So far it just mirrors Brendan and Philipp's examples, but I think that changing it like that would make it possible to swap things out more cleanly at each level, especially once we support other glyphs.

datashader/pipeline.py
jbednar: Ah, good point.  Maybe BokehPipeline, then.  

But it's not just for the notebook callback -- it works fine without the callback:

image

But that plot won't be interactive in the notebook, without the callback.

datashader/pipeline.py
jbednar: I went ahead and made it work like that, in part because I think the previous interface was halfway in between a Charts-like interface and a datashader-like interface.  I think it makes more sense to have an interface that closely mirrors what datashader supports (which it does now), and another (perhaps more limited) interface that closely mirrors Charts.Scatterplot.  There's very little code involved in either approach, and we can then decide which one is more useful in practice and then keep one or both of them as seems appropriate.

jbednar:
I switched it so that the Pipeline object simply returns a PIL image, so that it's independent of Bokeh, and I think that makes for a much nicer interface overall.  Lots of things got simplified that way.  I'll go ahead and merge this; we can do further cleanup on master.

PR #41, opened 2016-02-03
IPythonKernelCallback updates instead of redrawing image

philippjfr:
Small change to the IPythonKernelCallback so it updates an existing image glyph rather than redrawing one each time. Should be a better approach than popping and recreating a new glyph each time.

### DIFF
--- a/datashader/callbacks.py
+++ b/datashader/callbacks.py
@@ -2,7 +2,7 @@ import uuid, json
 
 from bokeh.embed import notebook_div
 from bokeh.document import Document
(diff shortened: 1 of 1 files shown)


jbednar:
That sounds cleaner.  I keep trying to think if there is a way to write this class so that it works with any Bokeh object, using whatever the callback returns as data and parameters for the object.  Perhaps that would be a base class, and then this would be the image-specific class.  I can't quite get my head around how to make it work cleanly, and I guess that means we should just merge this now and come back to it...

PR #42, opened 2016-02-03
Add option for count to accept no column

jcrist:
If no column specifies, just increments on bin-hits. If specified, only counts non `nan` values.

Also updated other aggregates to work appropriately in the presence of `nan` values.


jbednar:
Ok, good!  I was actually wondering about how to achieve that when there was no suitable field.  Happy to have it merged now.

PR #44, opened 2016-02-04
dashboard example: now loads data relative to yml config and not dashboard.py

brendancol:
https://github.com/bokeh/datashader/issues/27


jbednar:
Thanks!  I'll go ahead and merge this, even though the tests aren't passing, because it shouldn't affect the tests anyway.

PR #46, opened 2016-02-05
Notebook cleanup

jbednar:
Ready for comments, but not yet for merge.

- Fixed area to avoid cropping off the suburbs
- Set better defaults:
   * count() just counts by default (no need to specify a field)
   * high-dynamic-range blue colormap everywhere, by default

### DIFF
--- a/datashader/callbacks.py
+++ b/datashader/callbacks.py
@@ -8,8 +8,18 @@ from bokeh.io import _CommsHandle
 from bokeh.util.notebook import get_comms
 
 
-class IPythonKernelCallback(object):
+class InteractiveImage(object):
+    """
+    Bokeh-based interactive image object that updates on pan/zoom events.
 
+    Given a Bokeh plot and a callback function, calls the function
+    whenever the pan or zoom changes the plot's extent, regenerating
+    the image dynamically.  Works in a Jupyter/IPython notebook cell,
+    using the existing notebook kernel Python process (not a separate
+    Bokeh server).  Does not yet support usage outside the notebook,
+    but could be extened to use Bokeh server in that case.  
+    """
+    
     jscode="""
         // Define a callback to capture errors on the Python side
         function callback(msg){{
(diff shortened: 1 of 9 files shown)


jbednar:
Also cleaned up various interface bits, and completely reworked the examples.  There are now three .ipynb files, though only two are mentioned in the README, because osm.ipynb can't be run directly unless one figures out how to make the castra file required.  Should be ready to merge now.

datashader/callbacks.py
```diff
@@ -114,10 +137,13 @@ def redraw_image(self, ranges):
...
+        image = self.callback(x_range=x_range, y_range=y_range, w=ranges['w'], h=ranges['h'], **self.kwargs)
         self.ds.data.update(dict(image=[image.img], x=[x_range[0]], y=[y_range[0]],
                                  dw=[dw], dh=[dh]))
 
 
     def _repr_html_(self):
         return self.div
+
+
+IPythonKernelCallback = InteractiveImage  # Temporary; for backwards compatibility
```
jcrist: No need for backwards compatibility, we have no users yet :) Like the name change though.

datashader/pipeline.py
```diff
@@ -59,14 +42,14 @@ class Pipeline(param.Parameterized):
...
-    agg = param.ClassSelector(reductions.Reduction, default=reductions.count("count"), 
+    agg = param.ClassSelector(reductions.Reduction, default=reductions.count(), 
         doc="""Function for incrementally reducing a bin's values into a scalar.""")
 
     transfer_fns = param.HookList(default=[], doc="""
         Optional function(s) to apply to the aggregated bin values, before
         they each get converted into a color.""")
 
-    color_fn = param.Callable(default=Interpolate(), doc="""
+    color_fn = param.Callable(default=partial(tf.interpolate,low="lightblue",high="darkblue"), doc="""
```
jcrist: Why a different default here, instead of just `tf.interpolate()`?

jbednar: History -- I changed the default here first, and only then on tf.interpolate.  Sure, I should delete it here, for simplicity.

datashader/callbacks.py
jbednar: Right; I'll be removing that shortly.  At the time I had a lot of different notebooks, but I think they are all updated now.

PR #47, opened 2016-02-05
dashboard example: added map labels and basemap opacity

brendancol:
- adds map labels layer with checkbox
- adds basemap opacity slider
- refactors a couple `on_change` callbacks.
- changes to dashboard layout to fit additional controls


jbednar:
Looks great, thanks!

PR #54, opened 2016-02-08
Tests and cleanup for Pipeline

jcrist:
- Use numpydoc standard for documentation
- Add tests
- Remove param dependency
- Update notebooks for slight change in call signature

### DIFF
--- a/datashader/pipeline.py
+++ b/datashader/pipeline.py
@@ -1,77 +1,58 @@
 from __future__ import absolute_import, division, print_function
 
-"""
-Declarative interface to Datashader.
-
-Provides a configurable pipeline that makes it more convenient to
-specify individual stages independently from the others.  Does not
-cover all possible datashader functionality, and does not add any new
-underlying capabilities; all functionality is also available using the
-default imperative interface provided by other files.
-"""
-
-import param
+from toolz import identity
 
 from . import transfer_functions as tf
 from . import reductions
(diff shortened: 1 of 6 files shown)


jcrist:
Ping @jbednar for review. Only spelling change here is `df` and `glyph` aren't keyword arguments. This is because they're required parameters, and so no default makes sense.

datashader/pipeline.py
```diff
@@ -1,77 +1,58 @@
...
+    ----------
+    df : pandas.DataFrame, dask.DataFrame
+    glyph : Glyph
+        The glyph to bin by.
+    agg : Reduction, optional
+        The reduction to compute per-pixel. Default is ``count()``.
+    transform_fn : callable, optional
+        A callable that takes the computed aggregate as an argument, and
+        returns another aggregate. This can be used to do preprocessing before
+        passing to the ``color_fn`` function.
```
jbednar: No big deal, since we don't have any examples using non-color_fn transferfns, but this doesn't make it easy to chain transfer_fns anymore...

jcrist: Having it be a list of 1 argument functions was fine, but a bit limiting. I'd argue that taking in just a single function (which could be composed externally) made more sense.

jcrist:
Got the go ahead to merge externally in conversation. Merging.

PR #56, opened 2016-02-08
Add conda recipe

ahmadia:
This will automatically run development tests on install, and tests are available as `datashader.test()` as well.

### DIFF
--- a/conda.recipe/meta.yaml
+++ b/conda.recipe/meta.yaml
@@ -1,0 +1,34 @@
+package:
+  name: datashader
+  version: 0.1.0
+
+extra:
+  channels:
+    - blaze
+
+requirements:
+  build:
+    - python
+    - setuptools
+
+  run:
+    - python
+    - dask >=0.7.6
+    - datashape >=0.5.1
+    - numba >=0.23.1
+    - odo >=0.4.2
+    - numpy >=1.7
+    - pandas >=0.15.0
+    - pillow >=3.1.1
+    - xarray >=0.7.0
+    - toolz >=0.7.4
+
+test:
+  requires:
+    - pytest >=2.8.5
+  imports:
+    - datashader
+
+about:
+  home: https://github.com/bokeh/datashader
+  license: New BSD
--- a/conda.recipe/run_test.py
+++ b/conda.recipe/run_test.py
@@ -1,0 +1,1 @@
+import datashader; datashader.test()
--- a/datashader/__init__.py
+++ b/datashader/__init__.py
@@ -1,6 +1,6 @@
 from __future__ import absolute_import
 
-__version__ = '0.0.1'
+__version__ = '0.1.0'
 
 from .core import Canvas
 from .reductions import (count, sum, min, max, mean, std, var, count_cat,
@@ -14,3 +14,11 @@ try:
     from .dask import *
 except ImportError:
     pass
+
+def test():
+    try:
+        import os, pytest
+        pytest.main(os.path.dirname(__file__))
+    except ImportError:
+        import sys
+        sys.stderr.write("You need to install py.test to run tests -- conda install py.test")
--- a/setup.py
+++ b/setup.py
@@ -1,12 +1,11 @@
 from setuptools import setup
-import datashader
 
 with open('requirements.txt') as f:
     install_requires = f.read().strip().split('\n')
 
 setup(name='datashader',
-      version=datashader.__version__,
+      version='0.1.0',
       description='Data visualization toolchain based on aggregating into a grid',
       url='http://github.com/bokeh/datashader',
       install_requires=install_requires,
-      packages=['datashader'])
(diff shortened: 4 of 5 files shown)


conda.recipe/meta.yaml
```diff
@@ -0,0 +1,34 @@
...
+    
+  run:
+    - python
+    - dask >=0.7.6
+    - datashape >=0.5.1
+    - numba >=0.23.1
+    - odo >=0.4.2
+    - numpy >=1.7
+    - pandas >=0.15.0
+    - param
```
jcrist: Param is no longer needed (removed in #54).

setup.py
```diff
@@ -1,12 +1,11 @@
 from setuptools import setup
-import datashader
 
 with open('requirements.txt') as f:
     install_requires = f.read().strip().split('\n')
 
 setup(name='datashader',
-      version=datashader.__version__,
```
jcrist: Is there a reason for swapping this out?

datashader/__init__.py
```diff
@@ -14,3 +14,10 @@
     from .dask import *
 except ImportError:
     pass
+
+try:
+    import os, pytest
+    def test():
+        pytest.main(os.path.dirname(__file__))
+except ImportError:
+    pass
```
jcrist: Perhaps:

```python
def test():
    import os
    import pytest
    pytest.main(os.path.dirname(__file__))
```

That way the top level namespace isn't polluted with pytest/os imports. Not a big deal though.

conda.recipe/run_test.py
```diff
@@ -0,0 +1 @@
+import datashader; datashader.test()
```
jcrist: What's the argument for installing the tests as part of the package, instead of how we do it in dask? https://github.com/blaze/dask/blob/master/conda.recipe/run_test.py

conda.recipe/meta.yaml
ahmadia: Okay, I'll amend the PR on this.

setup.py
ahmadia: this doesn't even work if you don't have the dependencies installed.  If you wanted to go this route version should be stored in a single file (I don't know how to reconcile this with meta.yaml)

datashader/__init__.py
ahmadia: Ahh, good catch.  I'll fix that.

conda.recipe/run_test.py
ahmadia: This means the package can be tested on the install side without forcing the user to install from source.

jcrist: Ok, that makes sense. Just to clarify, these are only run automatically on build, not on install, correct?

setup.py
jcrist: Ah, good catch. Updating a version number in 3 places is unfortunate, but acceptable.

conda.recipe/run_test.py
ahmadia: That's correct.

ahmadia:
I've amended the PR.

datashader/__init__.py
```diff
@@ -14,3 +14,10 @@
     from .dask import *
 except ImportError:
     pass
+
+try:
+    def test():
+        import os, pytest
+        pytest.main(os.path.dirname(__file__))
+except ImportError:
+    pass
```
jcrist: Try block no longer needed.

ahmadia: is pytest a hard requirement then?

ahmadia: indeed it is, I'll pull it out.

jcrist: It won't be a hard requirement, the call to `test` will just fail when called with `ImportError`. The `try` block is only testing if a function can be created, which won't ever fail with `ImportError`.

ahmadia:
Alright, I've pulled out the try block (no other changes).

conda.recipe/meta.yaml
```diff
@@ -0,0 +1,33 @@
...
+  run:
+    - python
+    - dask >=0.7.6
+    - datashape >=0.5.1
+    - numba >=0.23.1
+    - odo >=0.4.2
+    - numpy >=1.7
+    - pandas >=0.15.0
+    - pillow >=3.1.1
+    - pytest >=2.8.5
```
jcrist: Sorry, another comment :/. Pytest isn't a hard requirement for running, but should be a build requirement (since tests are run on build). Is this correct?

ahmadia: Yes, but then we'd need to restore the try/catch block.  You can't have it both ways.

jcrist: Dask also isn't a hard requirement (only used if found on the machine). Unsure if it should be part of the build or not. Probably couldn't hurt, idk.

jcrist:
Thanks!

PR #59, opened 2016-02-09
 Improve docstrings, a few tests.

jcrist:
- Add docstrings to most things
- Add tests for some untested functions
- Few small code fixes, mostly to improve consistent behavior.

### DIFF
--- a/datashader/core.py
+++ b/datashader/core.py
@@ -27,7 +27,7 @@ class Expr(object):
 class Axis(object):
     """Interface for implementing axis transformations.
 
-    Instances hold implementations of implementations too and from axis space.
+    Instances hold implementations of transformations to and from axis space.
     The default implementation is equivalent to:
 
     >>> def forward_transform(data_x):
--- a/datashader/core.py
+++ b/datashader/core.py
@@ -7,18 +7,87 @@ from odo import discover
 from .utils import Dispatcher, ngjit
 
 
+class Expr(object):
+    """Base class for expression-like objects.
+
+    Implements hashing and equality checks. Subclasses should implement an
+    ``inputs`` attribute/property, containing a tuple of everything that fully
+    defines that expression.
+    """
+    def __hash__(self):
+        return hash((type(self), self.inputs))
+
+    def __eq__(self, other):
+        return type(self) is type(other) and self.inputs == other.inputs
+
+    def __ne__(self, other):
+        return not self == other
+
+
 class Axis(object):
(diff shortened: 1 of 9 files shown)


datashader/core.py
```diff
@@ -7,18 +7,87 @@
...
+        return type(self) is type(other) and self.inputs == other.inputs
+
+    def __ne__(self, other):
+        return not self == other
+
+
 class Axis(object):
+    """Interface for implementing axis transformations.
+
+    Instances hold implementations of implementations too and from axis space.
```
jbednar: implementations of implementations?

Presumably "to and".

datashader/glyphs.py
```diff
@@ -2,18 +2,31 @@
...
 
 
-class Glyph(object):
+class Glyph(Expr):
+    """Base class for glyphs."""
     pass
 
 
 class Point(Glyph):
+    """A point, with center at ``x`` and ``y``.
```
jbednar: An idealized (zero area) point?

I.e., can you clarify that this is the mathematical definition of point, not a dot that has a specific area (which many plotting programs will call a point)?  Later we'll add a circle, but for now people may think this is meant to be a circle.

datashader/reductions.py
```diff
@@ -116,6 +118,7 @@ def _finalize(bases, **kwargs):
 
 
 class FloatingReduction(Reduction):
+    """Base classes for reductions that always have floating dtype."""
```
jbednar: "floating-point" might sound less mysterious.

datashader/reductions.py
```diff
@@ -265,6 +327,14 @@ def _finalize(bases, **kwargs):
 
 
 class std(Reduction):
+    """Standard Deviation of all elements in ``column``.
+
+    Parameters
+    ----------
+    column : str
+        Name of the column to aggregate over. Must be numeric. ``NaN values in
+        the column are skipped.
```
jbednar: ``NaN``

jbednar:
Looks great; happy to have it merged after the minor typo fixes noted.

jbednar:
BTW, is it possible to make the tests pass before the conda package is built?  I thought it was failing due to dynd, but since that's no longer needed, it must be something else...

jcrist:
If you click on the build failure (red x on the commit), you can see what made the test fail. In this case it was a failure to import bokeh when I added the `--doctest-modules` flag for python 2.7. All tests on master have been passing, and of course we don't merge code with tests failing.

PR #60, opened 2016-02-09
Tweaks for NYC taxi example and update call to pipeline

birdsarah:


birdsarah:
Maybe this should have been two PRs.

The second commit fixes what appears to be a small bug in datashader.

The first commit is just some small bokeh clean-ups based on my experience running through the example for the first time.

jcrist:
Oh, yeah, that bug was my fault. I didn't expect changing the name of a positional arg to break anything, but that's why we have tests :/. Still need to get tests in for this (perhaps @philippjfr, if you'd be interested :)). Thanks for the fix!

philippjfr:
@jcrist Sure, I'll try to put some together tomorrow.

jcrist:
@birdsarah, this looks good to me. Is this good to merge?

jbednar:
Thanks for those updates!  I'd be happy to have it merged, but what did you mean by "the black line makes it a bit of a false comparison on the initial plots"?

jbednar:
@jcrist, those positional args used to be keyword args, when trying to decide on a clean interface.  I think it's good now, with these fixes.

jcrist:
This looks fine to me, and would be nice to get in before tagging a release. Merging.

PR #62, opened 2016-02-09
Install odo from blaze channel.

jcrist:
Odo 0.4.2 is now on the blaze channel, install on travis from there.


jbednar:
Should this be updated in the conda recipe as well?

jcrist:
I'm not sure what you mean by that. The conda recipe is just a list of requirements - those haven't changed. This is just updating where we install odo from on travis.

jbednar:
Ah, true.  What I meant is to wonder how conda will know to get odo from a special place, if Travis has to be told explicitly where to get it.

PR #68, opened 2016-02-10
Add instructions to install off master to readme

### LINKED ISSUE #67: Restore manual installation information to README
Hi,

We just had a user on the bokeh mailing list who couldn't install datashader - there's no win-64 package.

I went to point them to the manual install I used the other day, but it's gone from the README. 

Can you add it back? Other people who, want to hack on datashader for example, will still want the manual install information.

Or perhaps - given that you're already setup on RTD, then, it can just be in there under a dev guide?

jcrist:
Fixes #67.

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,11 +31,20 @@ subject to frequent changes.
 
 ## Installation
 
-Datashader is available using the `conda` package manager, from the `bokeh`
-channel:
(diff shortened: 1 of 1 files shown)


jbednar:
Thanks!

PR #79, opened 2016-02-22
[WIP] Add line glyph

jcrist:
This adds a basic implementation of a line glyph. Lines are represented as a `(x, y)` column pair, similar to points. Consecutive rows are joined by line segments between points. Determining proper aggregate behavior is a bit tricky. For now, the following behavior is used:
- If multiple *consecutive* rows map to the same pixel, only the first row is appended to the aggregate.
- Bins that are incremented by interpolating between two points receive the first point's fields (effectively a zero-order-hold approximation of the underlying data)
- Rows that have `nan` for a coordinate are treated as a discontinuity and skipped.

Algorithmic details
- Cohen-Sutherland is used to clip line segments to within the bounding box
- Bresenhams is used to perform the line interpolation.

Both have been mildly optimized, and perform well enough on my laptop (although will never reach the speed of points, as we're doing more work).

Todo:

- [x] Tests
- [x] Parallelize with dask. Currently, using with dask will result in gaps between dataframe chunks

### DIFF
--- a/datashader/glyphs.py
+++ b/datashader/glyphs.py
@@ -82,7 +82,8 @@ class Line(_PointLike):
     """
     @memoize
     def _build_extend(self, x_mapper, y_mapper, info, append):
-        _extend = _build_line_kernel(append, x_mapper, y_mapper)
+        draw_line = _build_draw_line(append, x_mapper, y_mapper)
+        extend_line = _build_extend_line(draw_line)
         x_name = self.x
         y_name = self.y
 
@@ -90,12 +91,12 @@ class Line(_PointLike):
             xs = df[x_name].values
             ys = df[y_name].values
             cols = aggs + info(df)
-            _extend(vt, bounds, xs, ys, plot_start, *cols)
+            extend_line(vt, bounds, xs, ys, plot_start, *cols)
 
         return extend
 
 
-# -- Helpers for drawing computing geometry --
+# -- Helpers for computing line geometry --
 
 # Outcode constants
 INSIDE = 0b0000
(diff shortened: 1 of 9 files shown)


datashader/glyphs.py
```diff
@@ -66,3 +67,182 @@ def _compute_x_bounds(self, df):
 
     def _compute_y_bounds(self, df):
         return df[self.y].min(), df[self.y].max()
+
+
+class Line(Glyph):
+    """A point, with points defined by ``x`` and ``y``.
+
```
jbednar: A polyline, i.e., a list of connected line segments, with vertexes defined by ``x`` and ``y``.

(but if that's correct, seems like this should be called Polyline or PolyLine.  To me, Line implies Straight Line, and this is more like a piecewise-linear approximation to a curve.)

datashader/glyphs.py
```diff
@@ -66,3 +67,182 @@ def _compute_x_bounds(self, df):
...
+    Parameters
+    ----------
+    x, y : str
+        Column names for the x and y coordinates of center of each point.
+    """
```
jbednar: coordinates of each vertex.

jbednar:
There's a good bit of grotty-looking code in there, but that's what line drawing algorithms are like.   What's odd is to see this kind of code being done in Python; it looks like C!  Without Numba this would be some seriously slow Python code.

In any case, happy to see this merged after the fix above.  Is there a gist/notebook to see it in action?

datashader/glyphs.py
jcrist: While that's *technically* correct, I feel that most users would call it a line. I'm mostly just mirroring bokeh's plotting interface here: http://bokeh.pydata.org/en/latest/docs/user_guide/plotting.html#single-lines

jbednar: Ok.  Matching bokeh is a good enough reason.  Still needs updated docstring, though.

jbednar:
Maybe for one trace of OSM data, e.g. https://www.openstreetmap.org/user/Wimmel/traces/2116153?  I can't see where to get bulk downloads of lots of such traces...

jbednar:
Is this ready to merge now?  I'm happy with it, and very eager to see it folded in.

jcrist:
No, it still needs tests (see checklist at top).

jbednar:
But Peter needs it for a demo tomorrow; can we add tests separately (unless you are working on them now?)  I also need to merge the interactive fixes...

jcrist:
This is ready for review.

jbednar:
The new changes look good; thanks!  Happy to see it merged.

PR #80, opened 2016-02-23
InteractiveImage Improvements

philippjfr:
This PR refactors ``InteractiveImage`` initialization into separate methods and changes how the events sent back to BokehJS are generated resulting in a considerable speedup. Rather than having to debounce events until you stop interacting with the plot it is now responsive enough on the NYC example that a simple throttle set to about ~250 ms let's you pan and zoom fairly fluidly. For larger datasets it may well be too slow again so maybe letting the user toggle between debouncing and simple throttling would be nice.

The performance boost comes from generating a model update event directly rather than converting the entire document to json and then computing a patch for only the changed bits (which should only ever have been the datasource anyway). I do think this is the right approach to show off datashader in the notebook, but it's not a general solution to update plots.

I also promised some unit tests at some point, so I'll add some to this PR in the next few days.

### DIFF
--- a/datashader/callbacks.py
+++ b/datashader/callbacks.py
@@ -1,6 +1,10 @@
-import uuid, json
+import uuid
 
-from bokeh.core.json_encoder import serialize_json
+try:
+    import ujson as json
+except:
+    import json
+
 from bokeh.embed import notebook_div
 from bokeh.document import Document
 from bokeh.models import CustomJS, ColumnDataSource
@@ -64,7 +68,7 @@ class InteractiveImage(object):
 
     _callbacks = {}
 
-    def __init__(self, bokeh_plot, callback, throttle=250, **kwargs):
+    def __init__(self, bokeh_plot, callback, throttle=200, **kwargs):
         """
         The callback function should have the signature:
 
@@ -158,11 +162,13 @@ class InteractiveImage(object):
         """
         Generate an update event json message.
         """
-        return serialize_json({'events': [{'attr': u'data',
-                                           'kind': 'ModelChanged',
-                                           'model': self.ds.ref,
-                                           'new': self.ds.data}],
-                               'references': []})
+        data = dict(self.ds.data)
+        data['image'] = [data['image'][0].tolist()]
+        return json.dumps({'events': [{'attr': u'data',
+                                       'kind': 'ModelChanged',
+                                       'model': self.ds.ref,
+                                       'new': data}],
+                           'references': []})
 
 
     def update_image(self, ranges):
--- a/datashader/callbacks.py
+++ b/datashader/callbacks.py
@@ -15,7 +15,8 @@ from bokeh.util.notebook import get_comms
 
 class InteractiveImage(object):
     """
-    Bokeh-based interactive image object that updates on pan/zoom events.
+    Bokeh-based interactive image object that updates on pan/zoom
+    events.
 
     Given a Bokeh plot and a callback function, calls the function
     whenever the pan or zoom changes the plot's extent, regenerating
@@ -23,6 +24,24 @@ class InteractiveImage(object):
     using the existing notebook kernel Python process (not a separate
     Bokeh server).  Does not yet support usage outside the notebook,
     but could be extened to use Bokeh server in that case.
+
+    Parameters
+    ----------
+    bokeh_plot : plot or figure
+        Bokeh plot the image will be drawn on
+    callback : function
+        Python callback function with the signature::
+
+           fn(x_range=(xmin, xmax), y_range=(ymin, ymax),
+              w, h, **kwargs)
+
+        and returning a PIL image object.
+    throttle : int
+        The throttle parameter specifies how frequently events
+        are generated in milliseconds.
+    **kwargs
+        Any kwargs provided here will be passed to the callback
+        function.
     """
 
     jscode="""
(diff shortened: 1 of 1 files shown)


philippjfr:
I've got another patch that uses optionally uses ultrajson to serialize, which gives another factor of 3-5 speedup. Not going to push that unless you think the optional dependency is worth it but it works fluidly with a 50 ms timeout on the NYC example.

jbednar:
Ok, at least on my phone the changes seem ok, so I'm happy to merge. I guess add the ultrasonic patch  first, though.

jbednar:
Ultrajson!

philippjfr:
Okay I feel like an idiot, I just double checked everything and while there is some speedup, it's nowhere near as dramatic as I thought. What happened is that I split screened the browser with the Chrome developer tools, which resized the plot and therefore the reduced the size of the image. Not sure it's worth merging now. I think there's more to be gained by considering sending the image at a slightly lower resolution.

philippjfr:
Okay, after briefly reverting to the old version it's definitely still worth merging, especially if ultrajson is available. I'll push that now.

jcrist:
Thanks for fixing this. If it's clear to you how to test this, tests would be much appreciated. Also, if you could clean up the class level docstring a bit to follow the [numpydoc standard](https://github.com/numpy/numpy/blob/master/doc/HOWTO_DOCUMENT.rst.txt) for describing parameters, that'd also be nice (just need to reformat the parameter descriptions). Neither is a blocker though.

philippjfr:
Just fixed the docstrings. I wasn't proposing testing the actual JS callback itself but everything up to that point, i.e. initialization of the image, updating and json conversion. Will have to learn about PhantomJS to test the rest, but probably not in the next few days :)

philippjfr:
Did some profiling on the update call by the number of pixels for the old implementation which computes a patch using bokeh functions and the new implementation which directly creates just the required message with and without ultrajson:

image

jbednar:
Ok, thanks for investigating that. Is there any reason not to have *both* debounce and throttle?  I.e., don't trigger an event for at least 100ms, and then throttle them to avoid queuing up too many of them?

jbednar:
As for using a slightly lower resolution, I really can't tell how to select the actual resolution of the plot.  Right now the requested datashading size is independent of the Bokeh plot size, so we may have more or less than the correct number of pixels.  @bryevdv, is there any way to find out the inner plot size, in pixels, when specifying the overall size of a Bokeh figure with axes?

jbednar:
I'll merge now so that we can start having better response times, but we may remove the ultrajson bit later, and there are also a few more fixes to make to docstrings.

philippjfr:
>Ok, thanks for investigating that. Is there any reason not to have both debounce and throttle? I.e., don't trigger an event for at least 100ms, and then throttle them to avoid queuing up too many of them?

That's not what debouncing does or maybe I'm misunderstanding you. Debouncing means that as long as you're generating new events nothing will happen, only when there has been no events for x milliseconds does an event get triggered. So for any reasonable debouncing value that means that as long as you're panning or zooming there is absolutely no events generated until you're done. If you interactions with the are shorter than the actual time taken to compute the event that makes sense, but if the event has time to finish while you're still interacting with the plot simple throttling is more appropriate. I think using regular throttling now is fine especially if it's coupled with a fairly conservative timeout as you have set. Ideally the timeout would adjust depending on how long the callback takes to execute but that's overcomplicating things for now.

Here's a diagram on what debouncing does vs throttling, assuming the timeout is set to 50 ms for both.

```
0ms                           250ms                          
|     |     |     |     |     |
x  x  x     x           x     x   - Events
                  o               - Debounced events  
o     o     o           o     o   - Throttled events
```

x - triggered events
o - executed events

jbednar:
Ah.  What I want is neither of those things, then.  I want events to be ignored for the first 200 milliseconds, because there's no point in re-rendering until the wheel zoom or pan has gotten reasonably far away from the starting point.  But then I want something to be triggered, even if events keep coming in, and to be triggered repeatedly if there are events available, but at a frequency no more than every 500ms.  Translating those both to 50 ms for your diagram:

```
0ms                           250ms                          
|     |     |     |     |     |
x  x  x  x  x  x        x        - Events
i  i  o  i  o  i  o     o        - Events that I want to trigger
```
Maybe that's not causal, though.

philippjfr:
The underscore accepts a {leading: false} argument, which I think will do what you want, it'd be something like this:

```
0ms                           250ms                          
|     |     |     |     |     |
x  x  x  x x   x        xxxx     - Events
i  i  o  i  o     o           o  - Events that I want to trigger
```

jbednar:
Sounds perfect!  So that means not using debouncing at all, but throttling plus a special leading-events skip time?  If so that's two parameters for the user to worry about, but we could probably get away with just saying that there is one parameter X, such that the first X ms of events are discarded, and then at most one event is accepted for every X ms after that.  Sound ok?  Then probably 250 is a reasonable default, and then anyone with a plot that takes longer than 250ms to render should set it to a value longer than the time it takes to render a plot?

philippjfr:
No, it's still just one parameter it's just that it doesn't trigger an event immediately but waits to see if there are any other events in the timeout period and then triggers the latest one once the timeout period is up, triggering at most every timeout period after that.

jbednar:
Sounds perfect!

PR #83, opened 2016-02-24
Improved InteractiveImage throttling behavior

philippjfr:
Disables the leading edge event when interacting with the an InteractiveImage, so the plot isn't unnecessarily updated when the user has just started panning or zooming.


jbednar:
Great, thanks!

PR #88, opened 2016-02-29
added try/except around webbrowser.open() after user had trouble running from server

brendancol:


examples/dashboard/dashboard.py
```diff
@@ -290,7 +290,10 @@ def add_roots(doc):
     server = Server(app, io_loop=IOLoop(),
                     extra_patterns=[(r"/datashader", GetDataset)], port=APP_PORT)
 
-    print('Starting server at http://localhost:{}/...'.format(APP_PORT))
-    webbrowser.open('http://localhost:{}'.format(APP_PORT))
+    try:
+        print('Starting server at http://localhost:{}/...'.format(APP_PORT))
+        webbrowser.open('http://localhost:{}'.format(APP_PORT))
+    except:
+        pass
```
jbednar: I think the print statement should move out of the try/except block, because the message is about the *server* starting, not the web browser starting.  Before doing that, we should probably make a string attribute for the URL we're using here, so that the message and the argument to webbrowser always stay in sync.

After `except:`, it's probably worth printing a message saying what the user should do now that opening the web browser automatically was not successful.  Maybe:

```python
url = 'http://localhost:{}/...'.format(APP_PORT)
server = Server(app, io_loop=IOLoop(), extra_patterns=[(r"/datashader", GetDataset)], port=APP_PORT)
		  
print('Starting server at {}...'.format(url))
try:
    webbrowser.open(url)
except:
    print('Unable to open web browser; please navigate to port {} on the machine where the server is running (which may first need to be forwarded to your local machine if the server is running remotely)'.format(APP_PORT))
```

brendancol:
adding to other PR...

PR #98, opened 2016-03-02
Add `set_background`

### LINKED ISSUE #82: Plots with a black background
I'd like to be able to render images that have a black background, but I don't see how to do it.  The [`PIL.Image.fromarray`](http://pillow.readthedocs.org/en/3.1.x/reference/Image.html#PIL.Image.fromarray) function that we're using in `transfer_functions.py` generates an RGBA image, but the function call doesn't seem to allow the background color that comes through for transparent pixels to be specified.  Presumably we could make a blank black image using `PIL.Image.new` and then add the two images together, but that seems awkward...

jcrist:
Sets the background color of an image. Also cleaned up a few docstrings.

Fixes #82.

### DIFF
--- a/datashader/tests/test_transfer_functions.py
+++ b/datashader/tests/test_transfer_functions.py
@@ -93,6 +93,16 @@ img2 = tf.Image(np.array([[0x00000000, 0x00000000],
                 coords=coords2, dims=dims)
 
 
+def test_set_background():
+    out = tf.set_background(img1)
+    assert out.equals(img1)
+    sol = tf.Image(np.array([[0xff00ffff, 0xff0000ff],
+                             [0xff0000ff, 0xff00ff7d]], dtype='uint32'),
+                   coords=coords2, dims=dims)
+    out = tf.set_background(img1, 'red')
+    assert out.equals(sol)
+
+
 def test_stack():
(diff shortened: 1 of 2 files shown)


jbednar:
Interesting. github now says that I assigned you to fix that issue, since you did fix that issue, which is fair enough, I guess!  Still a bit confusing that it wasn't actually me as a human being who did that; I'd rather the message say that waffle.io assigned you!

Can the name be set_bg rather than set_background, to make it easier to use in notebooks (where set_bg(...) only makes the line a little bit longer, and thus won't cause line breaking as often)?

jcrist:
I much prefer the descriptive name. Characters are cheap, but developer understanding isn't. Of course this is opinion based, but is also a common stylistic convention with python code.

jbednar:
To me `bg` and `background` are equally good at conveying the concept of "background color of an image", to a Python user, but if you really disagree, then just go ahead and merge as-is.  Thanks!

jbednar:
(Same goes for "transfer_fns" vs. "transfer_functions"; I don't think "functions" conveys anything that "fns" doesn't, but maybe I'm just old and have seen those abbreviations so many times in my life so as to make no distinction anymore).

PR #100, opened 2016-03-04
Dashboard fixes

jbednar:
This PR adds:

 - Support for Castra-format files (via dask imported dynamically, thus not adding a required dependency)
 - Support for plotting pure counts (rows in the data, e.g. taxi trips, not associated with any particular field)
 - Support for plotting counts for the census data

It's also close to adding support for colorization by census race categories, but I'm not sure how to add that in a general way.  Below are some trivial diffs that are sufficient to get racial color categories shown, in a hardcoded way that removes support for non-categorical data.  With this PR and those diffs applied, there will initially be an error on startup because the default Field is counts in census.yml, but if the Field is then changed to Race to match the new Count Categories aggregate declared below, it should work.

So, @brendancol, can you take it from here?  Can you make dashboard.py support categorical information where appropriate?  I've added the race colors to the census.yml file already, but it would take me a while to figure out how to look that information up when needed, based on the "cat_colors" pointer I added to census.yml (but which could be changed if needed).   There will also of course need to be some logic to switch between tf.interpolate and tf.colorize.  

If we want to avoid errors for nonsensical combinations, just as for the earliest client.py file from ages ago, we'll presumably need to start declaring datatypes for these objects so that the buttons generally work rather than generally fail...

```python
0045-jbednar:~/datashader/examples> git diff
diff --git a/examples/dashboard/dashboard.py b/examples/dashboard/dashboard.py
index 5d304de..0c32239 100644
--- a/examples/dashboard/dashboard.py
+++ b/examples/dashboard/dashboard.py
@@ -55,8 +55,8 @@ class GetDataset(RequestHandler):
                          self.model.active_axes[1],
                          self.model.active_axes[2],
                          self.model.aggregate_function(self.model.field))
-        pix = tf.interpolate(agg, (255, 204, 204), 'red',
-                             how=self.model.transfer_function)
+        color_field = 'race_colors'
+        pix = tf.colorize(agg, self.model.config[color_field])

         # serialize to image
         img_io = pix.to_bytesio()
@@ -72,6 +72,7 @@ class AppState(object):
         self.load_config_file(config_file)

         self.aggregate_functions = OrderedDict()
+        self.aggregate_functions['Count Categories'] = ds.count_cat
         self.aggregate_functions['Count'] = ds.count
         self.aggregate_functions['Mean'] = ds.mean
         self.aggregate_functions['Sum'] = ds.sum
```

### DIFF
--- a/examples/dashboard/dashboard.py
+++ b/examples/dashboard/dashboard.py
@@ -22,7 +22,10 @@ from bokeh.plotting import Figure
 from bokeh.models import (Range1d, ImageSource, WMTSTileSource, TileRenderer,
                           DynamicImageRenderer, HBox, VBox)
 
-from bokeh.models import Select, Slider, CheckboxGroup, CustomJS, ColumnDataSource, Square, HoverTool
+from bokeh.models import (Select, Slider, CheckboxGroup,
+                          CustomJS, ColumnDataSource,
+                          Square, HoverTool)
+
 from bokeh.models import Plot, Text, Circle
 from bokeh.palettes import GnBu9, OrRd9, PuRd9, YlGnBu9, Greys9
 
--- a/examples/dashboard/dashboard.py
+++ b/examples/dashboard/dashboard.py
@@ -29,9 +29,6 @@ from tornado.web import RequestHandler
 from webargs import fields
 from webargs.tornadoparser import use_args
 
-from pdb import set_trace
-
-
 # http request arguments for datashing HTTP request
 ds_args = {
     'width': fields.Int(missing=800),
--- a/examples/dashboard/dashboard.py
+++ b/examples/dashboard/dashboard.py
@@ -5,7 +5,6 @@ from os import path
 import yaml
 import webbrowser
 import uuid
-import pdb
 
 from collections import OrderedDict
 
--- a/examples/dashboard/dashboard.py
+++ b/examples/dashboard/dashboard.py
@@ -146,8 +146,8 @@ class AppState(object):
         self.field = list(self.fields.values())[0]
 
     def load_datasets(self):
-        print('Loading Data...')
         data_path = self.config['file']
+        print('Loading Data from {}...'.format(data_path))
 
         if not path.isabs(data_path):
             config_dir = path.split(self.config_path)[0]
--- a/examples/dashboard/dashboard.py
+++ b/examples/dashboard/dashboard.py
@@ -142,26 +142,26 @@ class AppState(object):
         # parse summary field
         self.fields = OrderedDict()
         for f in self.config['summary_fields']:
-            self.fields[f['name']] = f['field']
+            self.fields[f['name']] = None if f['field'] == 'None' else f['field']
         self.field = list(self.fields.values())[0]
 
     def load_datasets(self):
         print('Loading Data...')
-        taxi_path = self.config['file']
+        data_path = self.config['file']
 
-        if not path.isabs(taxi_path):
+        if not path.isabs(data_path):
             config_dir = path.split(self.config_path)[0]
-            taxi_path = path.join(config_dir, taxi_path)
+            data_path = path.join(config_dir, data_path)
 
-        if not path.exists(taxi_path):
-            raise IOError('Unable to find input dataset: "{}"'.format(taxi_path))
+        if not path.exists(data_path):
+            raise IOError('Unable to find input dataset: "{}"'.format(data_path))
 
         axes_fields = []
         for f in self.axes.values():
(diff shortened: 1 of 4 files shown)


jbednar:
It would also be good if it wouldn't re-run the entire pipeline if only the transfer fn changes.  That way people can quickly try out different transfer_fns.

jbednar:
@brendancol, while this is all fresh in your mind, there are some other small dashboard-related fixes in the issues list that could be addressed before you move on to other things...

examples/dashboard/dashboard.py
```diff
@@ -28,7 +31,6 @@
 from webargs import fields
```
jbednar: How should we handle the dashboard's dependence on webargs?  Should we just state that in the README?  I don't think we'd necessarily want to make webargs a dependency of datashader, at least not until conda supports optional dependencies.

jcrist: It's just an example, so my opinion is that the dependencies don't matter as long as their listed explicitly somewhere.

jbednar: We do need to make it easy for people to run the example, though.  After getting an error message about it, I briefly looked for webargs on conda, tried some bogus versions from non-main channels that didn't work, and eventually pip-installed webargs (which worked).  Most people probably aren't that dedicated.

jbednar:
Is the area highlighted by the hover tool correct?  It doesn't seem to be.  E.g. there's a hotspot just to the left of the blue area:

image

but if I hover directly over that, seemingly enclosing the hotspot in a blue box, the counts aren't particularly high:

image

Yet if I move the mouse down to the cell below and to the left, I get high counts indicative of a hotspot:

image

Does the displayed blue box need to be moved to accurately reflect the area in the hover information?

jbednar:
The behavior in the corners with a large hover-box size also looks suspicious -- shouldn't the box be the same size throughout the array (with at most a pixel of rounding more or less), not cropped to a quarter the size in the corners?

jbednar:
I guess this may be addressed by the proposed switch to averaging pixel values for reaggregation, but I can't quite make sense of the values for some combinations of Field and Aggregate.  E.g.:

image

What does counting the Fare field mean?  Counting how many non-empty Fare values there are?  If so, presumably it's not in `$`; not sure what to do about that.

jbednar:
Maybe report "Avg Fare ($) Count: xxxx", once it's averaging the values instead of re-aggregation?  I.e., show the aggregate explicitly, not just the field?

jbednar:
I added some commits for some useful bits, including out-of-core operation based on code from jcrist, but it's not quite ready to merge because the link to the census.castra file is blank (because I haven't yet heard back about options for hosting that file). It's also strange that castra has to be downloaded from a special channel; is there any way to get castra from a more public place (@jcrist?)

examples/dashboard/dashboard.py
```diff
@@ -107,21 +116,68 @@ def downsample(aggregate, factor):
...
+                legend_fig = self.model.create_legend(img.values,
+                                                      x=min_val,
+                                                      y=0,
+                                                      dh=18,
+                                                      dw=dw,
+                                                      x_start=min_val,
+                                                      x_end=max_val,
+                                                      y_range=(0,18))
+
+                self.model.legend_vbox.children = [legend_fig]
```
jbednar: Once it settles down a bit, it would be good to move the legend support into a function or a class, with appropriate parameters, with an eye to eventually moving it out of the dashboard.py file and into a datashader library file.

examples/dashboard/dashboard.py
```diff
@@ -135,6 +191,8 @@ class AppState(object):
     def __init__(self, config_file, outofcore, app_port):
 
         self.load_config_file(config_file)
+        self.plot_height = 560
+        self.plot_width = 810
```
jbednar: Can the plot just be set to fill the browser window, or is that hard from bokeh Python?

examples/dashboard/dashboard.py
```diff
@@ -330,6 +312,30 @@ def load_datasets(self,outofcore):
...
+        plot_options['plot_height'] = 120
+        plot_options['plot_width'] = 190
+
+        plot_options['min_border_bottom'] = 0
+        plot_options['min_border_left'] = 0
+        plot_options['min_border_right'] = 0
+        plot_options['min_border_top'] = 0
+        plot_options['outline_line_width'] = 0
+        plot_options['toolbar_location'] = None
+
```
jbednar: Seems like `plot_options = dict( opt=val ...) would be much more concise here...

Issue #93, opened 2016-03-01
Provide way to specify complex colormaps


jbednar:
Because datashader allows arbitrarily nonlinear mappings from data values to plotted colors, it would be very useful to have more precise control over colormaps.  Issue #81 is one important extension, but making it possible to generate novel colormaps that are a good match to a particular pipeline's output is also important.  @jcrist and @jbednar have discussed this previously over email, with @jcrist having details for a simple text-based specification that we could use.

PR #109, opened 2016-03-09
Added cmap option to interpolate.

### LINKED ISSUE #81: Allow usage of external colormaps, e.g. from matplotlib
It would be nice if `interpolate()` had knowledge of palettes to range over, instead of requiring a low and high color value.  This should be a fairly straightforward thing to do, but will just require a brief survey of what currently exists and the best way to incorporate the possibilities.  (e.g. https://bids.github.io/colormap/)

jbednar:
Fixes #81

Code from Peter to allow usage of matplotlib  colormaps, via a new `cmap` option to `interpolate`.  Sample usage:

```python
import pandas as pd
import numpy as np
import xarray as xr
import datashader as ds

n = 100000
data= {}
data['x'] = np.random.normal(0, 0.3, size=n)
data['y'] = np.random.normal(0, 0.3, size=n)
df = pd.DataFrame(data)

cvs = ds.Canvas(plot_width=300,plot_height=300)

tf.interpolate(agg, cmap="viridis", how="linear")
```

image

image

### DIFF
--- a/datashader/transfer_functions.py
+++ b/datashader/transfer_functions.py
@@ -85,9 +85,9 @@ def interpolate(agg, low="lightblue", high="darkblue", how='cbrt', cmap=None):
         'log', and 'linear'. Callables take a 2-dimensional array of
         magnitudes at each pixel, and should return a numeric array of the same
         shape.
-    cmap : str or None, optional
-        If specified, overrides the provided low, high values with the name
-        of a colormap to import from matplotlib.
+    cmap : str, optional
+        If specified, interpolate along the named colormap, as provided by
+        matplotlib.  Overrides low,high.
     """
     if not isinstance(agg, xr.DataArray):
         raise TypeError("agg must be instance of DataArray")
--- a/datashader/transfer_functions.py
+++ b/datashader/transfer_functions.py
@@ -109,10 +109,8 @@ def interpolate(agg, low="lightblue", high="darkblue", how='cbrt', cmap=None):
         a = np.where(np.isnan(data), 0, 255).astype(np.uint8)
         img = np.dstack([r, g, b, a]).view(np.uint32).reshape(a.shape)
     else:
-        from matplotlib import cm
-        if cmap not in dir(cm):
-            raise RuntimeException("Can't find Matplotlib colormap named '%s'" % cmap)
-        mapper = cm.__dict__[cmap]
+        from matplotlib.cm import get_cmap
+        mapper = get_cmap(cmap)
         tmp = mapper(data, bytes=True)
         tmp[:,:,3] = np.where(np.isnan(data), 0, 255).astype(np.uint8)
         img = tmp.view(np.uint32).reshape(data.shape)
--- a/datashader/transfer_functions.py
+++ b/datashader/transfer_functions.py
@@ -70,7 +70,7 @@ def _normalize_interpolate_how(how):
     raise ValueError("Unknown interpolation method: {0}".format(how))
 
 
-def interpolate(agg, low="lightblue", high="darkblue", how='cbrt'):
+def interpolate(agg, low="lightblue", high="darkblue", how='cbrt', cmap=None):    
     """Convert a 2D DataArray to an image.
 
     Parameters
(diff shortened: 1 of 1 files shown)


datashader/transfer_functions.py
```diff
@@ -97,12 +100,22 @@ def interpolate(agg, low="lightblue", high="darkblue", how='cbrt'):
...
-    img = np.dstack([r, g, b, a]).view(np.uint32).reshape(a.shape)
+
+    if cmap is None:
+        rspan, gspan, bspan = zip(rgb(low), rgb(high))
+        r = np.interp(data, span, rspan, left=255).astype(np.uint8)
+        g = np.interp(data, span, gspan, left=255).astype(np.uint8)
+        b = np.interp(data, span, bspan, left=255).astype(np.uint8)
+        a = np.where(np.isnan(data), 0, 255).astype(np.uint8)
+        img = np.dstack([r, g, b, a]).view(np.uint32).reshape(a.shape)
+    else:
```
jcrist: That's some unnecessarily mucky python. How about:

```python
from matplotlib.cm import get_cmap
mapper = get_cmap(cmap)
```

datashader/transfer_functions.py
```diff
@@ -85,6 +85,9 @@ def interpolate(agg, low="lightblue", high="darkblue", how='cbrt'):
         'log', and 'linear'. Callables take a 2-dimensional array of
         magnitudes at each pixel, and should return a numeric array of the same
         shape.
+    cmap : str or None, optional
+        If specified, overrides the provided low, high values with the name
+        of a colormap to import from matplotlib.
```
jcrist: Perhaps:

```
cmap : str, optional
    If specified, interpolate along the named colormap, as provided by matplotlib.
```

jcrist:
I'm not terribly enthused about depending on matplotlib for the colormaps here. Is there a different way these could be provided?

datashader/transfer_functions.py
```diff
@@ -97,12 +100,22 @@ def interpolate(agg, low="lightblue", high="darkblue", how='cbrt'):
...
+        r = np.interp(data, span, rspan, left=255).astype(np.uint8)
+        g = np.interp(data, span, gspan, left=255).astype(np.uint8)
+        b = np.interp(data, span, bspan, left=255).astype(np.uint8)
+        a = np.where(np.isnan(data), 0, 255).astype(np.uint8)
+        img = np.dstack([r, g, b, a]).view(np.uint32).reshape(a.shape)
+    else:
+        from matplotlib import cm
+        if cmap not in dir(cm):
+            raise RuntimeException("Can't find Matplotlib colormap named '%s'" % cmap)
+        mapper = cm.__dict__[cmap]
```
jcrist: How about:

```python
from matplotlib.cm import get_cmap
mapper = get_cmap(cmap)
```

jbednar:
Depending on matplotlib for the cmap is awkward, but all the other alternatives I can think of are also awkward, so we'll have to pick our poison (or pick all the poisons!).  Options:

1.  Use this PR's code as-is.<br />
**Pros:** Very little code, all already written.  Allows cmap names like "winter" that are familiar to many Python users.<br />
**Cons:** Adds dependency on Matplotlib.

2. Do the same as option 1, but depending on `bokeh.palettes` instead of mpl.<br />
**Pros:** Provides cmap names consistent with Bokeh<br />
**Cons:** Doesn't support word-style names provided by matplotlib, like "winter".  Adds dependency on bokeh. Requires implementing #93.

3. Do the same as option 2, but copy palettes.py into datashader, and possibly some cmap definitions from matplotlib
**Pros:** Provides cmap names consistent with Bokeh and potentially some from matplotlib.  No extra dependencies.
**Cons:** Sets up potential for drift or inconsistency between datashader and bokeh and or matplotlib for the definition of the colormaps (as for the existing colors.py).  Requires implementing #93.

My own preference is probably for 3, since I want #93 implemented anyway.

jbednar:
Looks like there must have been some problems in how the colormaps were being used in this PR, because the results with the new #111 PR are very different:

```python
import pandas as pd
import numpy as np
import xarray as xr
import datashader as ds

n = 30000
data= {}
data['x'] = np.random.normal(0, 0.3, size=n)
data['y'] = np.random.normal(0, 0.3, size=n)
df = pd.DataFrame(data)

cvs = ds.Canvas(plot_width=300,plot_height=300)
from matplotlib.cm import viridis, cool
tf.interpolate(agg, cmap=viridis, how="linear")
```
image

image

I'm guessing that the ranges viewed weren't being mapped to the colormap range properly in this old PR, because the new ones use the available color range much more fully.

jcrist:
Matplotlib colormaps expect data to be in `[0, 1.0]`, all other data is clipped. This PR didn't do that, so it clipped to the min/max values.

PR #111, opened 2016-03-10
Add cmap to interpolate

### LINKED ISSUE #81: Allow usage of external colormaps, e.g. from matplotlib

### LINKED ISSUE #93: Provide way to specify complex colormaps

jcrist:
Can now interpolate with a colormap. Colormaps can be either lists of colors, or a matplotlib colormap. Due to how bokeh stores palettes, this means that colormaps can be taken from either bokeh or matplotlib. Note that this *doesn't* add either as a dependency, just means that we understand how to use the colormappers.

Example notebook: https://gist.github.com/jcrist/a9dfc6e8f8bce3b4ae30

Fixes #81, fixes #93, subsumes #109.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,7 @@ install:
   # Install dependencies
   - conda create -n test-environment python=$TRAVIS_PYTHON_VERSION
   - source activate test-environment
-  - conda install dask numba numpy pandas pillow pytest toolz xarray
+  - conda install dask numba numpy pandas pillow pytest toolz xarray matplotlib
   # Need a few fixes to odo and datashape that aren't in the latest release
   - conda install -c blaze datashape odo
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,9 +16,9 @@ install:
   # Install dependencies
   - conda create -n test-environment python=$TRAVIS_PYTHON_VERSION
   - source activate test-environment
-  - conda install dask numba numpy pandas pillow pytest toolz xarray matplotlib
-  # Need a few fixes to odo and datashape that aren't in the latest release
-  - conda install -c blaze datashape odo
+  - conda install dask numba numpy pandas pillow pytest toolz xarray datashape odo
+  # Optional dependencies, for testing only
+  - conda install matplotlib
 
   - python setup.py develop --no-deps
 
--- a/datashader/transfer_functions.py
+++ b/datashader/transfer_functions.py
@@ -70,7 +70,7 @@ def _normalize_interpolate_how(how):
     raise ValueError("Unknown interpolation method: {0}".format(how))
 
 
-def interpolate(agg, low="lightblue", high="darkblue", how='cbrt'):
+def interpolate(agg, low="lightblue", high="darkblue", how='cbrt', cmap=None):
     """Convert a 2D DataArray to an image.
 
     Parameters
@@ -85,6 +85,7 @@ def interpolate(agg, low="lightblue", high="darkblue", how='cbrt'):
         'log', and 'linear'. Callables take a 2-dimensional array of
         magnitudes at each pixel, and should return a numeric array of the same
         shape.
+    cmap : list of colors or matplotlib.colors.Colormap, optional
     """
     if not isinstance(agg, xr.DataArray):
         raise TypeError("agg must be instance of DataArray")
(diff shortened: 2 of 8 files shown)


.travis.yml
```diff
@@ -16,7 +16,7 @@ install:
   # Install dependencies
   - conda create -n test-environment python=$TRAVIS_PYTHON_VERSION
   - source activate test-environment
-  - conda install dask numba numpy pandas pillow pytest toolz xarray
+  - conda install dask numba numpy pandas pillow pytest toolz xarray matplotlib
```
jbednar: Can we make a distinction here, at least via comments, between things we install that are actual dependencies (and thus should match pip and conda's list of dependencies) and things that are optional and are only for testing optional functionality?

It would be even better if there could be different travis targets with different sets of dependencies involved, so that we could verify that our optional dependencies are in fact optional, which is a trap that people often fall into, but I'm not sure how to do that in Travis.

datashader/transfer_functions.py
```diff
@@ -70,26 +70,38 @@ def _normalize_interpolate_how(how):
...
-        values.
+        values. Defaults are "lightblue" and "darkblue", respectively.
     how : str or callable, optional
         The interpolation method to use. Valid strings are 'cbrt' [default],
         'log', and 'linear'. Callables take a 2-dimensional array of
         magnitudes at each pixel, and should return a numeric array of the same
         shape.
+    cmap : list of colors or matplotlib.colors.Colormap, optional
+        The colormap to use. Can be either a list of colors (in any of the
+        formats described above), or a matplotlib colormap object.
```
jbednar: Should clarify that it overrides low,high.

datashader/transfer_functions.py
```diff
@@ -70,26 +70,38 @@ def _normalize_interpolate_how(how):
...
+    `high` color [default], or by a specified colormap.
+
     Parameters
     ----------
     agg : DataArray
     low, high : color name or tuple, optional
         The color for the low and high ends of the scale. Can be specified
         either by name, hexcode, or as a tuple of ``(red, green, blue)``
-        values.
+        values. Defaults are "lightblue" and "darkblue", respectively.
```
jbednar: Should we deprecate low, high, delete it from all our examples, and remove it in the following release, since `cmap=[low,high]` is about as concise and is more powerful?

datashader/transfer_functions.py
```diff
@@ -97,12 +109,22 @@ def interpolate(agg, low="lightblue", high="darkblue", how='cbrt'):
...
+    if isinstance(cmap, list):
+        rspan, gspan, bspan = np.array(list(zip(*map(rgb, cmap))))
+        span = np.linspace(span[0], span[1], len(cmap))
+        r = np.interp(data, span, rspan, left=255).astype(np.uint8)
+        g = np.interp(data, span, gspan, left=255).astype(np.uint8)
+        b = np.interp(data, span, bspan, left=255).astype(np.uint8)
+        a = np.where(np.isnan(data), 0, 255).astype(np.uint8)
+        img = np.dstack([r, g, b, a]).view(np.uint32).reshape(a.shape)
+    elif (getattr(type(cmap), '__module__', '').startswith('matplotlib') and
+          callable(cmap)):
```
jbednar: I'm assuming that you're not doing `issubclass(matplotlib.colors.Colormap)` to avoid importing matplotlib?  If so, why not just check that it's callable and try calling it, expecting some error to be raised if it's not something that works like a Matplotlib Colormap?  The current approach won't allow people to subclass that class or to provide duck-typed workalike objects, and I'm not sure what we gain from disallowing that.  I'd like to avoid mucking around with the module names if we don't have to...

jbednar:
Looks great!  I'll close PR #109, since I think this is a much more powerful approach.

datashader/transfer_functions.py
jcrist: I thought about doing that. Could go either way.

datashader/transfer_functions.py
jcrist: Matplolib colormaps have certain semantics (data is in [0, 1.0], take bytes kwarg), that could be easy to mess up if ducktyped. It just seemed less error prone to assert it was from matplotlib. Happen to change it though.

jbednar: I'll leave it up to you, I guess.

PR #119, opened 2016-03-17
Add histogram equalization

### LINKED ISSUE #85: Add histogram equalization transfer function
If a nonlinear 'how' operator is selected for `interpolate`, the color range covered by some intensity ranges will be increased, and for others it will be decreased (according to the specific function selected).  If such a function is chosen well for a particular dataset, local (in data intensity space) differences in value will map onto visible differences in RGB colors on a monitor, across the full range of data values.

An alternative to choosing a fixed such nonlinear function would be to do [histogram equalization](https://en.wikipedia.org/wiki/Histogram_equalization) on the aggregates, making each of the various commonly occurring data value ranges be mapped onto visible color differences, even for data whose distribution cannot a priori be known to map to a particular nonlinear function, and even if that function has some nonstandard shape (e.g. trimodal).  It should be straightforward to implement this function as a transfer function on aggregates, mapping each bin's magnitude into a histogram-equalized magnitude that spreads out the plotted values nicely across the visible color space.

jcrist:
Implements histogram equalization as an option (`how='eq_hist'`) for `interpolate`. Note that the equalization is performed on the aggregate before interpolation, not on the resulting image. Implementation adapted from that in scikit-image. Fixes #85.

Example:
screen shot 2016-03-17 at 4 45 14 pm

Note that this works best for aggregations with low contrast between all values (the mean here barely varies). Distributions that are highly skewed (such as the census population data) are still best viewed using `'log'` of the intensity.

### DIFF
--- a/datashader/transfer_functions.py
+++ b/datashader/transfer_functions.py
@@ -60,9 +60,45 @@ def stack(*imgs, **kwargs):
     return Image(out, coords=imgs[0].coords, dims=imgs[0].dims)
 
 
+def eq_hist(data, nbins=256):
+    """Return a numpy array after histogram equalization.
+
+    For use in `interpolate`.
+
+    Parameters
+    ----------
+    data : ndarray
+    nbins : int, optional
+        Number of bins to use. Note that this argument is ignored for integer
+        arrays, which bin by the integer values directly.
+
+    Notes
+    -----
+    This function is adapted from the implementation in scikit-image [1]_.
+
+    References
+    ----------
+    .. [1] http://scikit-image.org/docs/stable/api/skimage.exposure.html#equalize-hist
+    """
+    if not isinstance(data, np.ndarray):
+        raise TypeError("data must be np.ndarray")
+    if np.issubdtype(data.dtype, np.integer):
+        hist = np.bincount(data.ravel())
+        bin_centers = np.arange(len(hist))
+        idx = np.nonzero(hist)[0][0]
+        hist, bin_centers = hist[idx:], bin_centers[idx:]
+    else:
+        hist, bin_edges = np.histogram(data[~np.isnan(data)], bins=nbins)
+        bin_centers = (bin_edges[:-1] + bin_edges[1:]) / 2
+    cdf = hist.cumsum()
+    cdf = cdf / float(cdf[-1])
+    return np.interp(data.flat, bin_centers, cdf).reshape(data.shape)
+
+
 _interpolate_lookup = {'log': np.log1p,
                        'cbrt': lambda x: x ** (1/3.),
-                       'linear': lambda x: x}
+                       'linear': lambda x: x,
+                       'eq_hist': eq_hist}
 
 
 def _normalize_interpolate_how(how):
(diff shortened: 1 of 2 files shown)


jbednar:
Thanks!  Can you explain "low contrast between all values" and why log is better for highly skewed distributions?

datashader/transfer_functions.py
```diff
@@ -116,7 +152,7 @@ def interpolate(agg, low=None, high=None, cmap=None, how='cbrt'):
         agg = agg.where(agg > 0)
         offset = agg.min()
     how = _normalize_interpolate_how(how)
-    data = how(agg - offset)
+    data = how(agg.data - offset.data)
```
jbednar: Could you explain the implication of this change?

jbednar:
The implementation looks good to me, so I'm happy to see it merged.  Does it work with colorize() as well?  It seems like it might, but I'm not sure, and the colorize docstring doesn't mention it.

jcrist:
For datasets that have a huge and highly skewed range in intensity, the variations at the low end of the spectrum still all get binned to the same [0, 255] value as they did before, even though the intensity has been spread out evenly. The log scale is really needed here to compact this range a bit. This image might help:

screen shot 2016-03-17 at 4 58 23 pm

datashader/transfer_functions.py
jcrist: The functions take a numpy array instead of a xr.DataArray.

jcrist:
It kind of works with `colorize`, but the semantics of doing this there are a bit confusing (do you equalize per pixel, or for the entire array). For now I don't forbid it, but I also don't encourage it (hence why it's not documented).

datashader/transfer_functions.py
jbednar: But there's no corresponding change in the functions themselves in this PR, right?  So is it that the previous functions worked for either xarray or Numpy objects, but eq_hist only works with Numpy, so you made it always use Numpy?  That's fine; just curious...

jcrist: That's correct.

jbednar:
What are the semantics with colorize(), at the moment?

jcrist:
It does it for the entire array, which doesn't really make sense (hence the "it kind of works").

jbednar:
I can't quite tell what it would mean to equalize per pixel, or why colorize differs from interpolate in this respect...

jcrist:
I can't either, which is why I didn't implement it. One potential way would be to equalize the sum of the values in each pixel, then scale accordingly. Roughly (untested):

```python
factor = eq_hist(data.sum(axis=2))
data = data * factor/factor.max()
```

For now I just didn't implement it at all. You won't get any errors if you specify `how='eq_hist'` in `colorize`, but your output may not be ideal.

The reason this is important here, but not for `log` and friends is that those were *elementwise* computations, and only looked at the individual values, while `eq_hist` is a global operation that depends on the other values around it.

jbednar:
Ok, I guess go ahead and merge; it seems useful, won't mess anything else up, and will take me some playing around with it all to decide on the details.  Thanks!

PR #123, opened 2016-03-24
Add dynspread transfer function

jcrist:
`dynspread` dynamically spreads pixels in an image based on a density heuristic. This is useful for maintaining visibility of points when zooming in.

The heuristic is defined as the normalized mean number of non-empty adjacent pixels for each non-empty pixel. This gives a simple to compute tuning parameter in `[0, 1]`.


jbednar:
This looks great to me; just what I was imagining!  No method like this will ever be perfect, but it seems to work well in both sparse and dense regions.  I'll merge it now and we can tweak it over time.  Thanks!

PR #124, opened 2016-03-28
Census

jbednar:
Ok, the census example is now ready to review.  You can see the run version with images at https://anaconda.org/jbednar/census/notebook .

I'm happy with it, but I'd like to add at least one more cell, with some additional analysis.  E.g I'd like to plot all pixels where there is at least one person in each racial category (except perhaps 'o').   I *think* that should be a straigthforward xarray expression, but I couldn't immediately see the syntax for it with the multidimensional category grid structure.  E.g. for the pipeline:

```python
cvs = ds.Canvas(plot_width=plot_width, plot_height=plot_height)
agg = cvs.points(df, 'meterswest', 'metersnorth', ds.count_cat('race'))
img = tf.colorize(agg, color_key, how='eq_hist')
img
```
what should I replace `agg,` with to limit it to those pixels? 

It would also be great if someone else could read carefully through the notebook, looking for any mistakes, confusing bits, or anything that can be expressed more simply; I'm pretty confident in the fundamental claims I'm making in it, but it never hurts to have another pair of eyeballs on it!

If anyone knows a better way to handle the page background color than the HTML bits I'm using here, I'd love a better solution (e.g. one that doesn't make the .tail() output black, and that works with saving as HTML (nbconvert); the way I've done it works fine in a live notebook but the background color is lost when it's exported to HTML.

Now I'll need to go over the remaining notebooks to update them, but hopefully those will be less work.


jbednar:
Merging because everything's working; we can still make updates afterwards...

PR #128, opened 2016-03-29
InteractiveImage only attempts callback if IPython kernel available

philippjfr:
This avoids issues when the plot is reused outside a notebook context, which resulted in vertical zoom events getting dropped.


jbednar:
Thanks!

PR #131, opened 2016-03-30
Notebook updates

jbednar:
This PR is still somewhat WIP, but it's ready for the individual commits to reviewed:

> 648926c Allowed cmap to accept a reversed() list

Allows cmap=reversed(Greys9), instead of the awkward cmap=list(reversed(Greys9))

> c6162a2 Changed smallest circle shape for spreading from a + to a square

I think a square is less visually distracting, but could be persuaded otherwise.

> d2e3d9b Changed default bin number for eq_hist improve accuracy of float normalization

eq_hist ignores the bin number for integers, which is what we usually use it for, but for floats it's crucial to have enough bins.  256 is clearly not enough, but I'm not sure how high to go.

> 4611376 Added spreading support for Pipeline, now defaulting to dynspread

A Pipeline is meant as a simple object that covers most cases, and in e.g. the non-geo example we definitely need spreading by default, since the datapoints are all very far from each other.

> a55302f Expanded and updated timeseries notebook

Incorporated Peter's edits and clarified why downsampling is dangerous for time series.

> 0a640e9 Changed default interpolate and colorize method to eq_hist

The two defensible choices for the default interpolate method are eq_hist and linear.  eq_hist seems safer since it reveals structure that one might not realize was there otherwise.

> 8305880 Minor census cleanup
> 85792d9 Updated non-geo example to use eq_hist

Minor updates to bring notebooks up to date.

### DIFF
--- a/datashader/transfer_functions.py
+++ b/datashader/transfer_functions.py
@@ -114,7 +114,7 @@ def _normalize_interpolate_how(how):
     raise ValueError("Unknown interpolation method: {0}".format(how))
 
 
-def interpolate(agg, low=None, high=None, cmap=None, how='cbrt'):
+def interpolate(agg, low=None, high=None, cmap=None, how='eq_hist'):
     """Convert a 2D DataArray to an image.
 
     Data is converted to an image either by interpolating between a `low` and
@@ -182,7 +182,7 @@ def interpolate(agg, low=None, high=None, cmap=None, how='cbrt'):
     return Image(img, coords=agg.coords, dims=agg.dims)
 
 
-def colorize(agg, color_key, how='cbrt', min_alpha=20):
+def colorize(agg, color_key, how='eq_hist', min_alpha=20):
     """Color a CategoricalAggregate by field.
 
     Parameters
--- a/datashader/transfer_functions.py
+++ b/datashader/transfer_functions.py
@@ -11,7 +11,7 @@ from PIL.Image import fromarray
 
 
 from .colors import rgb
-from .composite import composite_op_lookup, source
+from .composite import composite_op_lookup, over
 from .utils import ngjit
 
 
@@ -247,7 +247,7 @@ def set_background(img, color=None):
     if color is None:
         return img
     background = np.uint8(rgb(color) + (255,)).view('uint32')[0]
-    data = source(img.data, background)
+    data = over(img.data, background)
     return Image(data, coords=img.coords, dims=img.dims)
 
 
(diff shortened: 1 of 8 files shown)


datashader/transfer_functions.py
```diff
@@ -245,7 +247,7 @@ def set_background(img, color=None):
     if color is None:
         return img
     background = np.uint8(rgb(color) + (255,)).view('uint32')[0]
-    data = source(img.data, background)
+    data = over(img.data, background)
```
jcrist: Why did you change this? This doesn't set the background, it overlays an image over a background of that color. For images that use alpha to indicate magnitude (output of `colorize`), this will make all set pixels have full alpha. I'm unsure if this is desired here.

jbednar: That's correct, and intended.  The problem was that previously, if one tried to set the background to black, it only changed the fully transparent pixels, which had very strange results -- if you take such an image and view it in e.g. Preview on a Mac, it's all garbled: the fully transparent pixels are black, but Preview's default gray background shines through (to a greater or lesser extent, depending on alpha) all the others.  So I don't think that set_background was doing something useful before; changing only the fully transparent pixels while removing their transparency yet leaving other pixels transparent doesn't result in a usable image in any scenario I can think of, and certainly not in the use cases I had in mind for set_background.

datashader/transfer_functions.py
```diff
@@ -161,6 +161,8 @@ def interpolate(agg, low=None, high=None, cmap=None, how='cbrt'):
         offset = agg.data[agg.data > 0].min()
     data = how(agg.data - offset, mask.data)
     span = [np.nanmin(data), np.nanmax(data)]
+    if isinstance(cmap,type(reversed(list()))):
+        cmap = list(cmap)
```
jcrist: `if isinstance(cmap, collections.Iterator)`

datashader/transfer_functions.py
jbednar: On second look, I think I see what you're saying here, but I don't think it's true.  I.e., you're worried that the alpha channel will simply be discarded from the top image?  The code isn't discarding the src alpha, as far as I can see; it's using it to control how the src gets mixed with the background, which is what we want here.   Try comparing census.ipynb with set_background using over and source for comparison, and you should see what I mean...

datashader/transfer_functions.py
jbednar: That's better, thanks.  Done.

PR #143, opened 2016-04-14
added streaming example using bokeh server and nyc_taxi.csv

brendancol:
To run example:
```
cd examples
bokeh serve stream.py --show
```

It will take a little bit for the nyc taxi to load into memory.  From there, you will need to pan or zoom once to kick off streaming.  

screen shot 2016-04-14 at 3 01 31 pm

Things that would make this example better:
- [ ] start streaming immediately instead of need to pan/zoom to start
- [ ] add time slider instead of text and allow user to change

### DIFF
--- a/examples/streaming.py
+++ b/examples/streaming.py
@@ -1,0 +1,119 @@
+import math
+
+from bokeh.io import curdoc
+from bokeh.plotting import Figure
+from bokeh.models import ColumnDataSource, CustomJS
+from bokeh.tile_providers import STAMEN_TONER
+from bokeh.models import VBox, HBox, Paragraph
+from bokeh.palettes import BuGn9
+
+import pandas as pd
+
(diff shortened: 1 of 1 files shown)


jbednar:
Looks great to me!  Nice, simple code, and a cool demo.

Any idea why the data is being clipped across the top of Manhattan in that image?  There are data points above that in the dataset...

brendancol:
I wonder if I'm working with an older nyc_taxi.csv that got cut off.  When I run `wc nyc_taxi.csv` I get `10679308` lines in the file.  Does that sound correct?

jbednar:
I get `11842095`, so yes, probably your nyc_taxi.csv was created using too-tight a bounding box.

jbednar:
That shouldn't affect the code, which I'm happy for you to merge whenever you are ready.

PR #150, opened 2016-05-03
InteractiveImage callback checks if underscore is available

philippjfr:
Small fix to ensure zooming does not break on static pages.

### DIFF
--- a/datashader/bokeh_ext.py
+++ b/datashader/bokeh_ext.py
@@ -85,7 +85,7 @@ class InteractiveImage(object):
         var throttled_cb = Bokeh._throttle['{ref}'];
         if (throttled_cb) {{
             throttled_cb()
-        }} else {{
+        }} else if (_ !== undefined) {{
             Bokeh._throttle['{ref}'] = _.throttle(update_plot, {throttle},
                                                   {{leading: false}});
             Bokeh._throttle['{ref}']()


jbednar:
Thanks!

jbednar:
Unfortunately, I don't see any change in the results, zooming still goes horizontally even after regenerating: https://anaconda.org/jbednar/uber/notebook

philippjfr:
Doesn't look like you've updated, I still see this in the output:

```
        var throttled_cb = Bokeh._throttle['d29af1a4_8064_4f35_a94e_86862076eab2'];
        if (throttled_cb) {
            throttled_cb()
        } else {
            Bokeh._throttle['d29af1a4_8064_4f35_a94e_86862076eab2'] = _.throttle(update_plot, 500,
                                                  {leading: false});
            Bokeh._throttle['d29af1a4_8064_4f35_a94e_86862076eab2']()
        }
```

jbednar:
Ah; I'd updated, but InteractiveImage has recently moved from callbacks.py to bokeh_ext.py. This notebook was still importing from callbacks and (unfortunately) finding it in an old stray callbacks.pyc file.  Damned .pyc files! Still no effect that I can see, though: https://anaconda.org/jbednar/uber/notebook

philippjfr:
You're right, it was still broken. Sorry about that, it obviously can't work but something was masking the bug for me. Pushed a fix and reopening shortly.

PR #151, opened 2016-05-03
Fixed InteractiveImage underscore conditional

philippjfr:
The fix in #150 wasn't quite right, sorry about that. This is the correct fix. I double checked this time:

https://beta.anaconda.org/philippjfr/nyc_taxi/notebook


jbednar:
43 uploads to anaconda cloud later, it finally works!  Thanks!

PR #157, opened 2016-05-16
added initial support for raster datasets

brendancol:
Adds support for raster dataset through integration with `rasterio`

- [x] add spelling for `Canvas.raster()`
- [x] add `raster.py` example.  To run example use `bokeh serve raster.py --show`
- [ ] test when example is outside data bounds
- [ ] add tests

Below is of raster with `colors.Hot` transfer function:
screen shot 2016-05-16 at 4 02 44 pm

### DIFF
--- a/datashader/core.py
+++ b/datashader/core.py
@@ -1,7 +1,5 @@
 from __future__ import absolute_import, division, print_function
 
-import math
-
 import numpy as np
 from datashape.predicates import istabular
 from odo import discover
@@ -173,8 +171,7 @@ class Canvas(object):
     def raster(self,
                source,
                band=1,
-               nodata=np.nan,
-               replace_nodata=None):
+               missing=None):
         """Sample a raster dataset by canvas size and bounds. Note: requires
         `rasterio` and `scikit-image`.
 
@@ -184,14 +181,17 @@ class Canvas(object):
             input datasource most likely obtain from `rasterio.open()`.
         band : int
             source band number : optional default=1
-        nodata : number
-            Value for nodata cells
-        replace_nodata : number
-            Replacement value for nodata cells
+        missing : number, optional
+            Missing flag, default is `None` and missing values are replaced with `NaN`
+            if floats, and 0 if int.
 
         Returns
-        _______
+        -------
         data : xarray.Dataset
+
+        Notes
+        -------
+        requires `rasterio` and `scikit-image`.
         """
 
         try:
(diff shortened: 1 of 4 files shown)


datashader/core.py
```diff
@@ -167,6 +170,61 @@ def line(self, source, x, y, agg=None):
...
+            input datasource most likely obtain from `rasterio.open()`.
+        band : int
+            source band number : optional default=1
+        nodata : number
+            Value for nodata cells
+        replace_nodata : number
+            Replacement value for nodata cells
+
+        Returns
+        _______
```
jcrist: Use `-` instead of `_`.

datashader/core.py
```diff
@@ -167,6 +170,61 @@ def line(self, source, x, y, agg=None):
             agg = any()
         return bypixel(source, self, Line(x, y), agg)
 
+    def raster(self,
+               source,
+               band=1,
+               nodata=np.nan,
+               replace_nodata=None):
+        """Sample a raster dataset by canvas size and bounds. Note: requires
```
jcrist: Another formatting nitpick: top line of docstring should be short (otherwise the sphinx formatting gets wonky). Could put the "Note:..." either in a `Note` section (see this excellent guide to numpydoc standard docstrings: https://github.com/numpy/numpy/blob/master/doc/HOWTO_DOCUMENT.rst.txt), or just in a new paragraph before the parameters bit.

datashader/core.py
```diff
@@ -167,6 +170,61 @@ def line(self, source, x, y, agg=None):
...
+        ymax = min(self.y_range[1], source.bounds.top)
+        
+        dx = self.x_range[1] - self.x_range[0]
+        dy = self.y_range[1] - self.y_range[0]
+
+        width_ratio = (xmax - xmin) / dx
+        height_ratio = (ymax - ymin) / dy
+
+        w = int(math.ceil(self.plot_width * width_ratio))
+        h = int(math.ceil(self.plot_height * height_ratio))
```
jcrist: Could just use `np.ceil` since it's already imported (not that imports are bad).

datashader/core.py
```diff
@@ -167,6 +170,61 @@ def line(self, source, x, y, agg=None):
...
+
+        rmin, cmin = source.index(self.x_range[0], self.y_range[0])
+        rmax, cmax = source.index(self.x_range[1], self.y_range[1])
+
+        data = source.read(1, window=((rmax, rmin), (cmin, cmax)))
+
+        if replace_nodata:
+            data[data == nodata] = replace_nodata
+
+        data = resize(np.flipud(data), (w, h), preserve_range=True)
```
jcrist: I wonder if there's a way to remove the skimage dependency here. Applying a linear scaling to a 2d numpy array (for the small case we need here) may not be that hard to write with just numpy.

examples/get_raster_data.sh
```diff
@@ -0,0 +1,7 @@
+mkdir -p data
+cd data
+wget http://s3.amazonaws.com/bokeh_data/geotiff_example.zip
+unzip geotiff_example.zip
+rm geotiff_example.zip
+cd ..
+
```
jcrist: Is there good reason for adding the dataset here instead of in the other python script (noting that this won't work on windows)? No strong feelings here.

datashader/core.py
```diff
@@ -167,6 +170,61 @@ def line(self, source, x, y, agg=None):
...
+        Parameters
+        ----------
+        source : rasterio.Dataset
+            input datasource most likely obtain from `rasterio.open()`.
+        band : int
+            source band number : optional default=1
+        nodata : number
+            Value for nodata cells
+        replace_nodata : number
+            Replacement value for nodata cells
```
jcrist: I wonder if there's a better api for these keywords. Also, much of the transfer-function stuff (for better or worse) assumes that NaN is missing for floats, and 0 otherwise. One thought would be to have:

```rst
missing : number
    Missing flag, default is `NaN`. Missing values are replaced with `NaN`
    if floats, and 0 otherwise.
```

If the user wants more control they could handle this themselves using Xarray after the load (or this could be entirely removed in favor of that):

```python
agg = cvs.raster(source)
agg2 = agg.where(agg == 0)    # NaN where agg is 0
# or
agg3 = agg.fillna(0)          # 0 where agg is NaN
```

jcrist:
I added several comments. Feel free to ignore some or all of them.

jbednar:
Looks great! Jim's comments all sound good to me. I'm happy for it to be merged when you are ready.

PR #170, opened 2016-06-15
Added progressbar to data download

brendancol:


examples/download_sample_data.py
```diff
@@ -1,8 +1,28 @@
...
+
 import pandas as pd
 import numpy as np
+import sys
+
+try:
+    import requests
+except ImportError:
+    print('Download script required requests package: conda install requests')
+    sys.exit(1)
```
jbednar: requires?

examples/download_sample_data.py
```diff
@@ -1,8 +1,28 @@
 """Download data needed for the examples"""
 from __future__ import print_function
-from os import path, makedirs
+
+from os import path, makedirs, remove
+from progress import bar as progressbar
```
jbednar: from .progress?

examples/progress.py
```diff
@@ -0,0 +1,171 @@
+# -*- coding: utf-8 -*-
+
+"""
+clint.textui.progress
+~~~~~~~~~~~~~~~~~
+
```
jbednar: What's the license for this file?

jbednar:
Works great; thanks so much!

jbednar:
Hmm: 
```
~/datashader/examples> python download_sample_data.py 
Downloading Taxi Data...
[################################] 1939419/1939419 - 00:05:37
Filtering Taxi Data
Reprojecting Taxi Data
 
All data downloaded.
330.635u 41.731s 10:09.79 61.0%	0+0k 334+770io 645pf+0w
```
This seems to indicate that the download took 5:37, but the overall task took 10:10, which makes the progressbar of relatively little help -- it's spending about the same time with and without the progressbar advancing.  And this was on a home connection; on a faster connection the download time will presumably be dwarfed by the reprojecting time.  

Should we add a progressbar for the pandas operations as well, e.g. using tdqm (http://stackoverflow.com/questions/18603270/progress-indicator-during-pandas-operations-python)?

PR #177, opened 2016-06-22
added hdf5 fallback for loading census data; fixed py3 divison bug (p…

brendancol:
…lot_width//400)


jbednar:
Looks good, thanks!

PR #178, opened 2016-06-22
added datasets.yml and updated download script

brendancol:


examples/README.md
```diff
@@ -1,11 +1,17 @@
 # Datashader Examples
 
 Many of the examples rely on the NYC Taxi dataset, which can be
-downloaded by running the `download_sample_data.py` script. This may
+downloaded by running the `datashader-download-data` command which is installed with datashader. This may
 take up to 20 minutes, even on a good network connection. The dataset
```
jbednar: I can update this after merge, but we should mention here that the command actually installs all the example data (apart from OSM).

PR #184, opened 2016-06-22
small refactors to decrease progress bar update rate and move smaller…

brendancol:
… datasets to top of yaml file


jbednar:
Happy for this to be merged if the tests pass.  On the other hand, I'm not sure any of what's listed here is tested explicitly, so maybe there is no need to wait.

PR #192, opened 2016-07-03
Added more responsive event throttling for DynamicImage

philippjfr:
This replaces the previous throttling scheme with a more responsive approach. The new approach is a combination of an initial delay, blocking events until Python has acknowledged completion status and a timeout in case the Python callback doesn't complete. In more detail the approach is as follows:

1. On initiating a new zoom event the callback schedules an event to be executed after the ``delay`` period and puts itself into the queue.
2. Any subsequent will replace the original event on the queue.
3. Once the delay period is up the latest zoom event will be sent to Python.
4. When the Python callback has completed it sends a "Completed" message.
5. When the JS callback catches the "Completed" message it either executes the latest event on the queue ensuring the last zoom level is processed or if the queue is empty unblocks the widget.
6. If the Python callback does not complete after the ``timeout`` period it will unblock automatically.

### DIFF
--- a/datashader/bokeh_ext.py
+++ b/datashader/bokeh_ext.py
@@ -44,9 +44,16 @@ class InteractiveImage(object):
               w, h, **kwargs)
 
         and returning a PIL image object.
-    throttle : int
-        The throttle parameter specifies how frequently events
-        are generated in milliseconds.
+    delay : int
+        Specifies the delay between the first received event
+        and when events are actually processed. Useful to
+        ignore events generated immediately after initiating
+        zooming.
+    timeout: int
+        Determines the timeout after which the callback will
+        process new events without the previous one having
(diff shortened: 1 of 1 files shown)


jbednar:
Sounds fabulous!  I'll merge now so that we can all try it out.

jbednar:
Hmm.  At the moment the taxi and census examples won't complete on my machine, saying "Kernel Restarting  The kernel appears to have died. It will restart automatically."  I don't *think* that's due to this update, as nyc_taxi isn't even getting as far as InteractiveImage.

philippjfr:
Don't think they can be related either. Sounds like something is wrong with your Jupyter server, master is working for me.

jbednar:
Same problem even after rebooting.  Odd.

PR #213, opened 2016-08-03
Removed optional ujson usage in InteractiveImage

philippjfr:
It seems ujson doesn't always work correctly and since the speed gain from using it is fairly minor I think we should just drop support for it. Fixes issue https://github.com/bokeh/datashader/issues/207. Thanks to @apiszcz for finding the issue.


jbednar:
Thanks!

PR #214, opened 2016-08-12
Replaced now-deprecated interpolate() and colorize() with shade().

jbednar:
WIP PR; not yet ready to merge.

Various people have found the names `interpolate()` and `colorize()` confusing, including me.   Both of them could be called "colorize" in some sense, since both of them involve choosing a pixel color for each value in an aggregate array, and it's not clear to users why there are different functions for the two cases.  To avoid confusion with the existing function `colorize()`, it seems better to choose a new name for a combined function, and `shade()` seems fitting given the name of this library and by analogy with 3D rendering/shading pipelines (where shading is the last step, as here). 

The initial commit also weakens checking of the `color_key` required for categorical aggregates.  It now allows passing a long list of colors,  to avoid having to pick precisely as many colors as categorical fields.   I'll now make a list of suitable distinguishable colors to use as a default, so that people can easily visualize their aggregates using just "shade(agg)" regardless of whether the agg is 2D or 3D (categorical).  There has already been a good bit of work on making nice [lists of distinguishable colors](http://graphicdesign.stackexchange.com/questions/3682/where-can-i-find-a-large-palette-set-of-contrasting-colors-for-coloring-many-d), but what we need here is a list of distinguishable *hues*, and all the lists I've seen are using value to help distinguish the colors, which we can't do here because we are using value to indicate the underlying data value (e.g. counts, in the census race example).  So making a good list of hues isn't as simple as one might think, but it's definitely better to do it once as a default than to force every user to pick such a list themselves!

### DIFF
--- a/datashader/transfer_functions.py
+++ b/datashader/transfer_functions.py
@@ -114,7 +114,7 @@ def _normalize_interpolate_how(how):
 
 
 def interpolate(agg, low=None, high=None, cmap=None, how='eq_hist', alpha=255, span=None):
-    """Convert a 2D DataArray to an image.
+    """(Deprecated) Convert a 2D DataArray to an image.
 
     Data is converted to an image either by interpolating between a `low` and
     `high` color [default], or by a specified colormap.
@@ -123,7 +123,7 @@ def interpolate(agg, low=None, high=None, cmap=None, how='eq_hist', alpha=255, s
     ----------
     agg : DataArray
     low, high : color name or tuple, optional
-        Deprecated. The color for the low and high ends of the scale. Can be
+        The color for the low and high ends of the scale. Can be
         specified either by name, hexcode, or as a tuple of ``(red, green,
         blue)`` values.
     cmap : list of colors or matplotlib.colors.Colormap, optional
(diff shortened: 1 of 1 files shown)


jbednar:
The corresponding notebooks also need to be updated to use shade(), but I haven't committed that yet to make it simpler to test backwards compatibility.

jbednar:
To get a good default categorical color key, it looks like we can use Bokeh's Set1_9, Set2_8, and Set3_12 concatenated together, with some colors removed from sets 2 and 3.  These color sets originally came from Cynthia Brewer, and are designed to be distinguishable within each set, but the sets are also mostly distinguishable from each other due to saturation differences.  The result is 22 different colors that are reasonably well distinguishable on both white and black backgrounds, and tolerate differences in staturation (for white backgrounds) or value (for black):

image

image

In each of these images, the first colors used will be in the bottom left, increasing left to right and top to bottom, so that a plot with up to 9 colors will use the highly saturated colors from set1, with the other colors used for increasing numbers of categories.   On their own, these colors can be distinguished when in large enough isolated patches as above, but note that mixing between these colors can create serious ambiguity.  E.g. the red and yellow categories overlapping create a nice orange, as expected, making locations with points from both the yellow and red categories ambiguous:

image

image

In practice, there is no way around such mixing being ambiguous for more than 3 categories, so it's still worth choosing colors specifically to reduce that probability for a given dataset.

The new colors seem to work fine on the NYC census data:

image

image

So I'll go with them for now.

jbednar:
Should be ready to merge now, pending the outcome of the tests.

PR #220, opened 2016-08-18
partial fix for dashboard bokeh 0.12.1 integration

brendancol:


brendancol:
This was a partial step for dashboard.  I've made another issue here: https://github.com/bokeh/datashader/issues/221

PR #227, opened 2016-08-30
Added safety check for non-positive log axis range

jbednar:
As reported in issue #226, datashader segfaults for a y_range starting at 0 or a negative number.  

This PR adds a check for this condition at the Canvas level, which provides a clear error message for cases where the user has explicitly set a y_range or an x_range with such an invalid starting value.  

This PR also adds a low-level error check within the Axis itself, to handle the case where the range has been determined automatically based on the data.  In this case, the name and type of the axis are not known, so the message is much more obscure, but there appears to be no simple way around that.  And even an obscure message is better than a segfault!

### DIFF
--- a/datashader/core.py
+++ b/datashader/core.py
@@ -1,5 +1,7 @@
 from __future__ import absolute_import, division, print_function
 
+import math
+
 import numpy as np
 from datashape.predicates import istabular
 from odo import discover
@@ -71,6 +73,10 @@ class Axis(object):
 
         """
         start, end = map(self.mapper, range)
+
+        if math.isinf(start) or math.isnan(start) or math.isinf(end) or math.isnan(end):
+            raise ValueError("Axis range must be a real number after transformation")
+        
         s = (n - 1)/(end - start)
         t = -start * s
         return s, t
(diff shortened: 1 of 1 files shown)


jbednar:
@jcrist, can you please check the (very short!) diffs and see if you have any objections or suggestions?

datashader/core.py
```diff
@@ -71,6 +73,10 @@ def compute_scale_and_translate(self, range, n):
 
         """
         start, end = map(self.mapper, range)
+
+        if math.isinf(start) or math.isnan(start) or math.isinf(end) or math.isnan(end):
```
jcrist: `not (np.isfinite(start) and np.isfinite(end))`

datashader/core.py
```diff
@@ -123,6 +129,12 @@ def __init__(self, plot_width=600, plot_height=600,
         self.x_axis = _axis_lookup[x_axis_type]
         self.y_axis = _axis_lookup[y_axis_type]
 
+        if x_axis_type=='log' and x_range and x_range[0]<=0:
+            raise ValueError("x_range must be >0 for a log x axis")
+        if y_axis_type=='log' and y_range and y_range[0]<=0:
+            raise ValueError("y_range must be >0 for a log y axis")
```
jcrist: It'd be more ideal to attach this validation to the log-axis object. How I'd do this with minimal changes would be:

- Make `LogAxis` and `LinearAxis` subclasses of `Axis`, with the `mapper` and `inverse_mapper` arguments moved to method definitions (with a `staticmethod` decorator).
- Add a `validate` method to both `Axis` subclasses with appropriate validation
- Call the validate methods for both axis in `bypixel`

The benefit of this is that it's more extensible, and the checks get run if the ranges/axis are mutated after initializing `Canvas`.

datashader/core.py
jbednar: Perfect!

datashader/core.py
jbednar: Well, that doesn't sound like minimal changes :-), but it does sound more extensible.  And the staticmethod decorator will work ok with ngjit?

jcrist: Yes, that's how all the reductions are done as well. See: https://github.com/bokeh/datashader/blob/master/datashader/reductions.py#L133-L136.

jbednar:
Superseded by #228.

PR #228, opened 2016-08-31
Added safety check for non-positive log axis range (refactored)

jbednar:
Supersedes #227.

As reported in issue #226, datashader segfaults for a y_range starting at 0 or a negative number.

This PR adds a check for this condition at the Axis level, which required refactoring Axis to use subclasses with methods rather than attributes with function objects to determine the behavior.


jbednar:
@jcrist, look ok now?

Issue #230, opened 2016-09-07
Unexpected behavior with line


johanfforsberg:
I'm seeing a strange behavior with the "line" aggregator. Running this sample code:
```python
import datashader
import numpy as np
import pandas as pd

y = [1, 0, 1, np.NaN, np.NaN, 0, 1, 0]
x = range(len(y))

df = pd.DataFrame(dict(x=x, y=y))
cvs = datashader.Canvas(x_range=[min(x), max(x)], y_range=[-1, 2],
                        plot_width=100, plot_height=100)
agg = cvs.line(df, "x", "y")
datashader.transfer_functions.shade(agg, cmap=["red"])
```
produces an image that does not draw a line from (5,0) to (6,1). I get exactly the same image if I replace x[5] with np.NaN. Somehow, having two NaN y-values in a row seems to cause an issue with drawing the following point.

jbednar:
Thanks for reporting that with such a clear example!  That made it very easy to spot the problem.  Should be fixed by the above PR.

PR #232, opened 2016-09-07
Fixed handling of NaN between line segments

jbednar:
As reported in #230, the behavior for line segments separated by NaN does not appear to be consistent:

```python
import datashader
import numpy as np
import pandas as pd

y = [1, 0, 1, np.NaN, np.NaN, 0, 1, 0]
x = range(len(y))

df = pd.DataFrame(dict(x=x, y=y))
cvs = datashader.Canvas(x_range=[min(x), max(x)], y_range=[-1, 2], plot_width=100, plot_height=100)
agg = cvs.line(df, "x", "y")
tf.shade(agg)
```

image

Here there should be two V shapes, but the first part of the second one is cut off.  Turns out that the first sample after np.NaN was erroneously being ignored.  After fixing that (applying this PR) it works fine:

image

### DIFF
--- a/datashader/glyphs.py
+++ b/datashader/glyphs.py
@@ -190,7 +190,7 @@ def _build_extend_line(draw_line):
             # the entire segment.
             if np.isnan(x0) or np.isnan(y0) or np.isnan(x1) or np.isnan(y1):
                 plot_start = True
-                i += 2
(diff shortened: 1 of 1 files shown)


jcrist:
+1.

PR #235, opened 2016-09-08
Landsat updates

jbednar:
The landsat notebook had a variety of problems:

- At 33MB, it was too large for the updated anaconda.org to display, now that there is a 25MB limit
- Images had a black border around them, obscuring the map in the immediate area when first displaying, and covering the entire map on zooming
- There wasn't much explanation of what was going on

I've decreased the image size slightly, making the rendered notebook be 18MB, and uploaded that to anaconda.org.  @brendancol and I debugged the black-border issues, finding that these images failed to declare that they were using "1" as a value indicating pixels with no data, which we've now hardcoded into the notebook.  I also changed the blue-only channel to use eq_hist, which makes a lot more detail show up than the previous linear colormapping.  The remaining images all used a custom nonlinear normalization, so I left those as is.  Finally, I added text to explain how everything works and how to use it.  Should be ready to merge.

### DIFF
--- a/datashader/core.py
+++ b/datashader/core.py
@@ -203,10 +203,11 @@ class Canvas(object):
                source,
                band=1,
                resample_method='bilinear',
-               use_overviews=True,
-               missing=None):
+               use_overviews=True):
         """Sample a raster dataset by canvas size and bounds. Note: requires
-        `rasterio` and `scikit-image`.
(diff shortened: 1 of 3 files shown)


jbednar:
Also removed the "missing" argument for creating rasters, which was meant for control over missing values, but which @brendancol decided was probably more confusing than it is worth.

PR #238, opened 2016-09-14
Remove .create on Select creation and convert options to list.

hsparra:
The .create function has been removed from Select. Need to pass a list of string or a list of tuples to Select. Will not accept a generator, which is what is returned by zip() in python 3.

### DIFF
--- a/examples/dashboard/dashboard.py
+++ b/examples/dashboard/dashboard.py
@@ -420,8 +420,6 @@ class AppView(object):
 
     def update_image(self):
         self.model.shader_url_vars['cachebust'] = str(uuid.uuid4())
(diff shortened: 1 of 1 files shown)


jbednar:
Thanks!  It looks like the options can now be a bare list of items, no longer needing to be a list of pairs, but it can be either as long as it's a true list (not an iterator or generator)?  For some of the changes you've made it's a regular list, and others it's turning the generator of pairs into a list; can we just pick one of these two approaches?  Presumably the plain list version is simpler, and then the various _options variables can be deleted?  If so I can make a new PR doing that...

jbednar:
Looks good now; thanks!

PR #239, opened 2016-09-14
Replace {get,set}() with getters and setters

mattpap:
This prepares datashader for bokehjs 0.12.3, where `get()` and `set()` will be deprecated. If you used `obj.get('prop_name')` or `obj.set('prop_name', value)` before, those become `obj.prop_name` and `obj.prop_name = value` respectively.

### DIFF
--- a/examples/raster.py
+++ b/examples/raster.py
@@ -52,14 +52,15 @@ dims = ColumnDataSource(data=dict(width=[], height=[], xmin=[], xmax=[], ymin=[]
 dims.on_change('data', on_dims_change)
 dims_jscode = """
 var update_dims = function () {
-    var new_data = {};
-    new_data['height'] = [plot.get('frame').get('height')];
-    new_data['width'] = [plot.get('frame').get('width')];
-    new_data['xmin'] = [plot.get('x_range').get('start')];
-    new_data['ymin'] = [plot.get('y_range').get('start')];
-    new_data['xmax'] = [plot.get('x_range').get('end')];
-    new_data['ymax'] = [plot.get('y_range').get('end')];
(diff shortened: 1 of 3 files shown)


examples/raster.py
```diff
@@ -52,14 +52,15 @@ def update_image():
...
+        width: [plot.frame.width],
+        xmin: [plot.x_range.start],
+        ymin: [plot.y_range.start],
+        xmax: [plot.x_range.end],
+        ymax: [plot.y_range.end]
+    };
+    dims.data = new_data;
 };
 
 if (typeof throttle != 'undefined' && throttle != null) {
```
mattpap: On an unrelated note, this code is redundant, as `throttle != null` covers both cases, because `undefined` and `null` the same under implicit comparison. This `if` statement could simply read `if (throttle != null)`. This is btw. how `?` operator in coffeescript is implemented. Otherwise, one should use explicit comparison operators. In general one should use the later when writing pure javascript to avoid weird implicit conversions.

jbednar:
Thanks!  Will the revised code be compatible with previous Bokeh versions, at least 0.12.1 and up?

mattpap:
@jbednar, this will work from 0.12.0 upwards. Property getters and setters where there for awhile, but we didn't enforce using them, just suggested that new or revised code in bokehjs use them, but now we are switching. Just to note, `get()` and `set()` still will be there, just under `getv()` and `setv()`. This is important specifically for `setv()`, because it implements additional features, like simultaneous and silent assigment, which currently (if at all) can't be achieved with setters.

philippjfr:
Looks good to me, thanks @mattpap for explaining the changes.

Issue #248, opened 2016-09-28
DOC: lack of legend for `ds.count_cat`


mortonjt:
In the last figure in the NYC taxi cab notebook, the order of the colors were given

```
Here the order of colors is roughly red (midnight), yellow (4am), green (8am), cyan (noon), blue (4pm), purple (8pm), and back to red (since hours and colors are both cyclic).
```

But it isn't clear exactly where this order of obtained.  Was it obtained from the ordering of `df['hour']`?  It would be really nice if we could have some sort of legend to explain the colors for these sorts of categories.

jbednar:
The colors were specified directly, per hour:

```
colors = ["#FF0000","#FF3F00","#FF7F00","#FFBF00","#FFFF00","#BFFF00","#7FFF00","#3FFF00",
          "#00FF00","#00FF3F","#00FF7F","#00FFBF","#00FFFF","#00BFFF","#007FFF","#003FFF",
          "#0000FF","#3F00FF","#7F00FF","#BF00FF","#FF00FF","#FF00BF","#FF007F","#FF003F",]
```

I.e., there are 24 colors in this list, starting with red, and each one is used in order for the 24 hours of the day starting at midnight (binning time by hour).   The [legends notebook](https://anaconda.org/jbednar/legends/notebook) shows how to add a legend.  We're still working on it being less awkward to do all this.

mortonjt:
oh awesome!  I didn't see that issue.  Will there be any interest in just passing a `pd.Series`?  When doing a first pass at the data, a legend painted on the plot is nice, but isn't explicitly required.  But it is awkward trying to figure out which colors map to which categories.  Because what I'm noticing is that ordering of the categories doesn't actually affect the ordering the colors.

For example, when I run the following code

```python
import pandas as pd
import datashader as ds
import datashader.transfer_functions as tf
from bokeh.plotting import output_notebook

df = pd.DataFrame(
    {'x': [3, 3, 3, 6, 6, 6, 8, 8, 8],
     'y': [2, 5, 8, 2, 5, 8, 2, 5, 8],
     'group' : ['a', 'a', 'a', 'b', 'b', 'b', 'c', 'c', 'c']
    }
)
df['group'] = df['group'].astype('category')
x_range, y_range = ((0, 10), (0, 10))
colors = ['#FF0000', "#00FF00", "#0000FF"]

plot_width  = int(750)
plot_height = int(plot_width//1.2)

[...]

PR #249, opened 2016-09-29
Added color_mapping function to return color key consistently

jbednar:
To help clarify the mapping from categories to colors as suggested in issue #248, added a function `colormapping(agg,color_key)` returning a dictionary listing the color for each category.


jbednar:
Not sure if this is still needed, but if so it can be brought back.

Issue #284, opened 2017-02-17
Index DataFrame column using `['column_name']` syntax will allow for numeric column names.


StevenCHowell:
As demonstrated in the example from [this stack overflow response](http://stackoverflow.com/a/42164577/3585557), when converting a NumPy array into a Dataframe, the column labels are initially numeric.  When trying to index by these column labels, I realized the backend DataFrame indexing is using the attribute method, `df.column_label`, and not the `df[column_label]` method.  The current method supports less variation in column label in that it only supports valid python identifiers, e.g., column labels cannot be purely numeric.  [This Pandas' doc page](http://pandas.pydata.org/pandas-docs/stable/indexing.html) has more details.  If possible, changing to the `[ ]` method (or `__getitem__ method`) would allow for a wider range of inputs, thereby simplifying the user experience.

jbednar:
We should be able to fix this; doesn't seem too difficult.  Thanks for the suggestion!

jbednar:
[A PR would hurry this process along, by the way! :-)]

StevenCHowell:
I'll see what I can do.  I did not find any "contributing guide".  Is there more than just cloning the repo, branching, then doing a pull request?  Is the `setup.py` designed to accept the `develop` option?

jbednar:
Right; that's all there is to it.  If you have already done ``conda install datashader``, it's probably easiest to do ``conda remove datashader`` (to remove datashader while leaving all the dependencies satisfied), and then do ``python setup.py develop``.  That's what I do, anyway!

StevenCHowell:
Yeah, I got a minute to create a new environment and get it setup; a quick test head positive results. I should have time to tinker on this tomorrow.

StevenCHowell:
I tested the code as it is (without the change made by pr #285), and it did take column names that would not work with the `df.col_name` syntax, e.g., `df['x values']`.  It seems the real issue was it did not take numeric column names.

PR #285, opened 2017-02-18
recast glyph.x and glyph.y as strings when checking if real

StevenCHowell:
This change allowed for using numeric column names in a Pandas DataFrame, the goal of #284.  It was not the change I thought would be required, changing `df.col_label` to `df['col_label']`.  I am not sure if and where the DataFrame indexing change would benefit the code but changing the two lines I did made Datashader one step easier to use for NumPy arrays.

This is essentially what I used now to plot data from an NumPy array where the first row is the x-values and all other rows are set of y-values:
```
n_data = len(data)
df = pd.DataFrame(data=data.T)
data_cols = range(1, n_data)
x_range = data[0, 0], data[0, -1]
y_range = data[1:].min(), data[1:].max()
canvas = datashader.Canvas(x_range=x_range, y_range=y_range, plot_height=300, plot_width=300)
aggs = collections.OrderedDict((c, canvas.line(df, 0, c)) for c in data_cols)
merged = xarray.concat(aggs.values(), dim=pd.Index(data_cols, name='cols'))
img = datashader.transfer_functions.shade(merged.sum(dim='cols'), how='eq_hist')
```

### DIFF
--- a/datashader/glyphs.py
+++ b/datashader/glyphs.py
@@ -23,9 +23,9 @@ class _PointLike(Glyph):
         return (self.x, self.y)
 
     def validate(self, in_dshape):
-        if not isreal(in_dshape.measure[self.x]):
+        if not isreal(in_dshape.measure[str(self.x)]):
             raise ValueError('x must be real')
-        elif not isreal(in_dshape.measure[self.y]):
+        elif not isreal(in_dshape.measure[str(self.y)]):
             raise ValueError('y must be real')
 
     def _compute_x_bounds(self, df):


StevenCHowell:
Do [these failed tests](https://travis-ci.org/bokeh/datashader/jobs/202995916) usually pass? Did my change break them?
```
datashader/tests/test_raster.py::test_raster_aggregate_default FAILED
datashader/tests/test_raster.py::test_raster_aggregate_nearest FAILED
datashader/tests/test_raster.py::test_raster_aggregate_with_overviews FAILED
datashader/tests/test_raster.py::test_raster_aggregate_without_overviews FAILED
datashader/tests/test_raster.py::test_out_of_bounds_return_correct_size PASSED
datashader/tests/test_raster.py::test_partial_extent_returns_correct_size FAILED
```

jbednar:
The tests passed the last time master was run, but maybe some other library has updated since then?  I'll commit something to the repo to see if it forces a re-run of the tests, in case these failures are unrelated.

jbednar:
Yep; it was an unrelated change; some other library has become more stringent, and we'll need to fix that.

Meanwhile, I didn't write the two lines in question, and I'm not entirely sure what they achieve.  I'm a bit worried that casting them to strings might cause problems if someone has unicode column names under Python3, which is not something I've tested.

jbednar:
The tests should now be fixed (PR #287), but I'm still wondering how safe it is to cast all column names to strings.

jbednar:
As I suspected, at least under Python 2 it is *not* safe to cast names to strings, e.g. if any of them are unicode:

```
>>> import pandas as pd, numpy as np
>>> df = pd.DataFrame({"x": np.zeros(3), u'\ua000': np.ones(3)})
>>> df["x"]
0    0.0
1    0.0
2    0.0
Name: x, dtype: float64
>>> df[str("x")]
0    0.0
1    0.0
2    0.0
Name: x, dtype: float64
>>> df[u'\ua000']
0    1.0
1    1.0
2    1.0
Name: ꀀ, dtype: float64
>>> df[str(u'\ua000')]
  File "<stdin>", line 1, in <module>
UnicodeEncodeError: 'ascii' codec can't encode character u'\ua000' in position 0: ordinal not in range(128)
```

So, I think we need a different approach.  You could put a test for the name being number.Number, and only make a string out of that, but it seems like there must be some other problem elsewhere that's really the thing to fix?  Pandas doesn't mind having numbers as column names, so is there a problem elsewhere in datashader that should be fixed instead?

jbednar:
Closing as it's not clear what to do that would make things simpler for Datashader users and is also safe for Pandas in general.

Issue #293, opened 2017-03-08
Bokeh class definition for `_ModelInDocument` has either moved or been deprecated, import needs to be updated


StevenCHowell:
In Bokeh 12.5dev14 release, the class `_ModelInDocument` is no longer defined in bokeh.model.  This is required for `datashader.bokeh_ext.InteractiveImage`.  Here is the traceback:

```
---------------------------------------------------------------------------
ImportError                               Traceback (most recent call last)
<ipython-input-7-9d4688ff35b4> in <module>()
      1 import datashader
----> 2 from datashader.bokeh_ext import InteractiveImage

/Users/schowell/data/code/datashader/datashader/bokeh_ext.py in <module>()
      8 from bokeh.io import notebook_div
      9 from bokeh.models import CustomJS, ColumnDataSource, Square, HoverTool, GlyphRenderer
---> 10 from bokeh.model import _ModelInDocument as add_to_document
     11 from bokeh.io import _CommsHandle
     12 from bokeh.util.notebook import get_comms

ImportError: cannot import name '_ModelInDocument'
```

There is a function definition for `_ModelInDocument` in [`bokeh/embed.py`](https://github.com/bokeh/bokeh/blob/65643839fcb8065e1ffc926c2295e380e4637e4a/bokeh/embed.py#L55) but it is different code.  I am not sure where this got moved to but it breaks the bokeh extension for interactive datashader images.  @bokeh/core was this class definition moved or deprecated?

bryevdv:
> was this class definition moved or deprecated? 

The old version was removed, and because it was a private function, it did not have any deprecation. There is a new version you can see in this PR:

https://github.com/bokeh/bokeh/pull/5932/files

It's not identical to the old version but the old version did things that are no longer relevant. I can't state that this is a permanent location, either. It's a very small internal helper built on stable lower level pieces. I'd suggest you create your own version for data shader.

StevenCHowell:
I did find that one but it seems quite a bit different.  As a reference, here is the code from the 0.12.4 release:
```
class _ModelInDocument(object):
    # 'models' can be a single Model, a single Document, or a list of either
    def __init__(self, models):
        from .document import Document

        self._to_remove_after = []
        if not isinstance(models, list):
            models = [models]

        self._doc = _find_some_document(models)
        if self._doc is None:
            # oh well - just make up a doc
            self._doc = Document()

        for model in models:
            if isinstance(model, Model):
                if model.document is None:
[...]

PR #294, opened 2017-03-08
Add code removed from bokeh 0.12.5 into InteractiveImage

StevenCHowell:
FWIW, this brings in the code removed from bokeh 0.12.5 which `InteractiveImage` relied on.  This serves to patch issue #293, though the discussion there 

> all of that recursive traversal to find a document somewhere was an ugly hack to work around a situation that no longer really occurs (at least not with any kind of normal usage). Dropping it in might "work" but it would not be my recommendation

motivates a closer review of the code that was brought over.


philippjfr:
Thanks for fixing this @StevenCHowell. (Ab)using the context manager like this is something I originally did to avoid copying the code from bokeh. I think it can probably be factored out more cleanly now that we can't rely on bokeh for this anymore. I'll have a stab at it shortly as this is something I'll have to do for [holoviews](https://github.com/ioam/holoviews) anyway.

philippjfr:
Thanks again for patching this so quickly. I think the solution used in https://github.com/bokeh/datashader/pull/296 will suffice because InteractiveImage is quite restricted in its usage anyway.

philippjfr:
@StevenCHowell It seems this is required afterall, sorry for closing it earlier. I'll reopen this now, do you have time to rebase it? Otherwise I'd be happy to do so.

StevenCHowell:
@philippjfr I just rebased my changes.  Hopefully the checks all pass and this is good to go.

StevenCHowell:
I did not get a chance to check that the changes still fix the problem after doing the rebase.  If needed, I could test this tonight.

philippjfr:
Did something go wrong with the rebase? The PR now includes various other, unrelated changes.

StevenCHowell:
Maybe, first rebase.  I followed [this Git for Teams](http://gitforteams.com/resources/rebasing.html).  Basically, I updated the master branch of my fork from the datashader/master branch, then checked out this branch, and rebased in the changes from the master branch.

Perhaps I should simply close this PR, make a new branch, then copy over the changes there?

jbednar:
That might be cleaner, thanks; this version looks like it will cause some issues...

philippjfr:
That probably seems easiest, thanks again for doing this. Once you've done that I'm happy to test it still works.

PR #296, opened 2017-03-09
Handle Document creation and removal directly

philippjfr:
Fix for https://github.com/bokeh/datashader/issues/293, rather than rely on a utility, I've decided to just create a ``Document`` directly and removing any old documents from the references. Without handling that cleanup I've found that ``WMTSTileSources`` which we frequently reuse raise errors because they are already owned by a different ``Document``. @bryevdv I'd appreciate it if you could have a quick look at this to see if the approach is generally okay?

### DIFF
--- a/datashader/bokeh_ext.py
+++ b/datashader/bokeh_ext.py
@@ -6,8 +6,9 @@ import json
 import numpy as np
 
 from bokeh.io import notebook_div
-from bokeh.models import CustomJS, ColumnDataSource, Square, HoverTool, GlyphRenderer
-from bokeh.model import _ModelInDocument as add_to_document
+from bokeh.document import Document
+from bokeh.models import (Model, CustomJS, ColumnDataSource, Square,
+                          HoverTool, GlyphRenderer)
 from bokeh.io import _CommsHandle
 from bokeh.util.notebook import get_comms
 from bokeh.models import Plot, Text, Circle, Range1d
@@ -134,11 +135,6 @@ class InteractiveImage(object):
         self.p.x_range.callback = callback
         self.p.y_range.callback = callback
 
-        # Initialize document
-        doc_handler = add_to_document(self.p)
-        with doc_handler:
-            self.doc = doc_handler._doc
-            self.div = notebook_div(self.p, self.ref)
 
     def _init_callback(self):
         """
@@ -230,7 +226,12 @@ class InteractiveImage(object):
         self.ds.data.update(new_data)
 
     def _repr_html_(self):
-        return self.div
+        self.doc = Document()
+        for m in self.p.references():
+            m._document = None
+        self.doc.add_root(self.p)
+        return notebook_div(self.p, self.ref)
+
 
 class HoverLayer(object):
     """


jbednar:
It looks good to me, and I'm happy to merge once @bryevdv approves.

bryevdv:
It's hard to say. i.e. "WMTSTileSources which we frequently reuse" raises flags for me. The error message is explanatory: Bokeh models can't be shared between documents. So then the follow on question is: if you already have a document, why not just use that? That's what `_ModelInDocument` did, it tried to find a suitable existing document, and then only created a new one if none was found. Except it also delved into all the children to look for possible documents, and this is no longer  advised. 

The reason it searched children was because of this situation
```
p = figure()   # this used to "automagically" add p to a document, but no longer
foo = VPlot(p) # no auto document here!
show(foo)      # foo not in doc, but p is! gotta dig around to find doc that p used
```

This doesn't really happen anymore, at least not without going out of your way to *make* it happen. I would now say the situation where you are trying to add a root to a doc, but some child way down is already owned by another document, should be considered an error condition.  And now in Bokeh, in effect it is: `add_root` will raise that "already in document" exception when it happens. 

Regarding just deleting an old document to get around that exception, that might work "fine" in some limited use cases, and make that's ok for you. But I would not universally recommend it or expect it to work well in all situations. E.g. if you try this mucking around in the presence of `push_notebook`, offhand I'd probably expect things to fall over.

bryevdv:
So maybe just use the old `_ModelInDocument` (in your codebase) after all? What it also did was to carefully *restore* any old documents that it temporarily changed (i.e. that's why it was a context manager) That is not useful for the Bokeh codebase but maybe it's what you want to do. 

But I'd say the best guidance is:

* don't share models between documents
* use existing documents when they are there
* only only "remove" things from documents by going though `remove_root`

bryevdv:
> especially because bokeh already provides a number of WMTSTileSource instances like STAMEN, which you say can only be used once? 

In retrospect, this was probably a design error. The solution now would be to do the thing that was done with `bokeh.palettes` (i.e. the "class as module" hack)

> As far as I can tell ModelInDocument only looks for a document at the top-level. 

The *new* one does. The *old* one (that was removed) traversed all children:
```
+    # Now look in children of models
 +    if doc is None:
 +        for model in models:
 +            if isinstance(model, Model):
 +                for r in model.references():
 +                    if r.document is not None:
 +                        doc = r.document
 +                        break
```

tho it's hidden in the `for r in model.references()` so maybe that's not obvious

I will say the issue is probably not as acute for non-server cases. Sharing models between documents is basically equivalent to sharing them between sessions and will definitely lead to unexpected (bad) things. It's easier to just make that blanket guidance than to try and make fine distinctions. The practical upshot in the non-server case is that you probably have two JS instances of a model with identical UUIDs and like I said, maybe that's OK in your particular case case.

philippjfr:
For the time being I'm tempted to merge this, ``InteractiveImage`` is so restricted anyway I don't think it'll ever encounter the issues you described.

>The solution now would be to do the thing that was done with bokeh.palettes (i.e. the "class as module" hack) so that every access to a module attributes gives a new instance.

Should I file an issue about this on bokeh?

jbednar:
I guess I'll go ahead and merge, as it fixes the current problem.  But I won't close the issue that says we need to pin down a stable API on BokehJS.

PR #299, opened 2017-03-15
Lidar example

PeterDSteinberg:
This is an example that uses the Puget Sound LiDAR consortium data (7zip archives of scattered X,Y,Z points in text files).

Modifications are made to `datasets.yml` to download the data and a 7zip helper package and a new notebook is added.

I will modify the notebook with more visualization options.

### DIFF
--- a/examples/LiDAR-datashader.ipynb
+++ b/examples/LiDAR-datashader.ipynb
@@ -1,0 +1,362 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Visualize LiDAR Scattered Point Elevation Data\n",
+    "\n",
+    "This notebook uses datashader to visualize LiDAR elevation data from [the Puget Sound LiDAR consortium](http://pugetsoundlidar.ess.washington.edu/), a source of LiDAR data for the Puget Sound region of Washington, U.S.\n",
+    "\n",
+    "The data are downloaded in 7z archives, then unzipped and visualized below."
+   ]
+  },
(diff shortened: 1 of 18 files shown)


PeterDSteinberg:
This is what it looks like now:

screen shot 2017-03-15 at 4 31 10 pm

jbednar:
Looks beautiful; thanks so much!  However, I'm going to close this and ask for a new PR, because this one includes unrelated changes to 16 other files.  Can you please submit one with just the new notebook and the changes to datasets.yml?

Also, there are a few other changes that would help this example fit in more smoothly with the others:

* The notebook name shouldn't include "datashader"; all the notebooks in this directory are about datashader.
* The notebook code shouldn't normally include shell commands.  We ask that users run the download script before any of these notebooks, and so if there is special setup to do, please put that into the download script or into the README in this directory.  
* Can installing pylzma just be mentioned in the README and the comments at the start of this file, not in the code cells?  I don't want the notebook to be installing code, and definitely not in data/. 
* Can the unzipping be done in a general way in the download_sample_data script instead?  That script already unzips other things, and it would be nice for a user of this notebook to see the datashader-related code, not a bunch of zipping commands.

For user convenience, we should probably update examples/environment.yml to include all of the packages needed for running the examples, at least if there is one single environment where everything will work (which I'm not sure about, for rasterio-based notebooks, but would be a nice goal to have).

PeterDSteinberg:
@jbednar Great feedback - will fix each of your bulleted items in separate PR (forgot this was a branch off non-master branch).

PR #301, opened 2017-03-16
LiDAR dask jupyter notebook with modifications to datasets

PeterDSteinberg:
Re-open #299 (299 had extraneous files included).  The following TODO items were identified by @jbednar - 
 * [x] The notebook name shouldn't include "datashader"; all the notebooks in this directory are about datashader.
 * [x] The notebook code shouldn't normally include shell commands. We ask that users run the download script before any of these notebooks, and so if there is special setup to do, please put that into the download script or into the README in this directory.
 * [x] Can installing pylzma just be mentioned in the README and the comments at the start of this file, not in the code cells? I don't want the notebook to be installing code, and definitely not in data/. 
 * [x] Can the unzipping be done in a general way in the download_sample_data script instead? That script already unzips other things, and it would be nice for a user of this notebook to see the datashader-related code, not a bunch of zipping commands.


jbednar:
Thanks. Can you please rebase to make Greg's changes not be listed redundantly here?

README.md
```diff
@@ -50,6 +50,22 @@ Datashader is not currently provided on pip/PyPI, to avoid broken or
 low-performance installations that come from not keeping track of
 C/C++binary dependencies such as LLVM (required by Numba).
 
+One way to easily install `datashader` and related GIS and visualization tools is to install `earth-env` ([source code](http://github.com/ContinuumIO/earth-env)) with one a command like the following for Python 3.6 (change the `-36` part for Python versions 2.7, 3.4, 3.5, or 3.6):
+
```
jbednar: earth-env is a private repo, so it can't be referenced here.  I'm not sure why earth-env has different environment specifications for different Python versions, but I don't see any differences between the files currently, and we would certainly strive to keep them the same.  With both those points in mind, I've put one environment.yml (envt name "ds") into datashader/examples/, and I suggest that you just use and discuss that one here.

Issue #129, opened 2016-03-29
Recommended file format for large files


jbednar:
Datashader is agnostic about file formats, working with anything that can be loaded into a dataframe-like object (currently supporting Pandas and Dask dataframes).  But because datashader focuses on having good performance for large datasets, the performance of the file format is a major factor in the usability of the library.  Thus we should use examples that serve to guide users towards good solutions for their own problems, recommending and demonstrating approaches that we find to work well.

Right now, our examples use CSV and castra or HDF5 formats.  It is of course important to show a CSV example, since nearly every dataset can be obtained in CSV for import into the library.  However, CSV is highly inefficient in both file size and reading speed, and it also truncates floating-point precision  in ways that are problematic when zooming in closely to a dataset.

[Castra](https://pypi.python.org/pypi/castra) is a relatively high-performance binary format that works well for the large datasets in the examples, but it is not yet a mature project, and is not available on the main conda channel.  Should we invest in making castra be more fully supported?  If not, I think we should choose another binary format (HDF5?) to use for our examples.

apaytuvi:
Why not the npz format for a numpy matrix?

jbednar:
Our input data isn't usually a numpy matrix, and can't easily be converted to one, because in most cases the different columns have different data types.  Of course, we can store that as a set of different numpy arrays, but that's a bit clunky, so npz is just one of several possible alternatives with different tradeoffs.

miktoki:
Is [`bcolz`](https://github.com/Blosc/bcolz) an option? It seems fast, compact and somewhat pydata-friendly (numpy + import/export to pandas,HDF5), plus it intends to support GPU. It is based on `Blosc` like Castra.

jbednar:
Sure, bcolz is a good option worth considering.   We'll need to try out the various options and see if we can find a good balance of speed, file size, and cross-platform support.  [Feather](https://blog.rstudio.org/2016/03/29/feather/) is another possibility.

brendancol:
For now, we are going to continue to use the castra py2.7 file in the census notebook, but fallback to the `census.h5` version which has better.  Moving forward, it looks like `feather-format` would be a great solution to performance and interop issues

jbednar:
A set of utilities for measuring read and write performance for various columnar files have been added to the examples directory.  To use them, follow these steps:

1. Pull the latest datashader github master and cd to the examples/ directory
[...]

PR #303, opened 2017-03-23
Update filetimes.yml to use conda instead of pip

gbrener:
There are a lot of pip packages listed in examples/filetimes.yml. Since its creation, relevant conda packages have been created and updated. This PR updates the filetimes.yml file to reflect mostly conda packges, with two that are still only available through `pip`.


examples/filetimes.sh
```diff
@@ -19,8 +14,8 @@ ${timer} python filetimes.py ${1}.parq        dask    census meterswest metersno
 ${timer} python filetimes.py ${1}.snappy.parq dask    census meterswest metersnorth race
 ${timer} python filetimes.py ${1}.parq        pandas  census meterswest metersnorth race
 ${timer} python filetimes.py ${1}.snappy.parq pandas  census meterswest metersnorth race
-${timer} python filetimes.py ${1}.castra      dask    census meterswest metersnorth race
-#${timer} python filetimes.py ${1}.castra      pandas  census meterswest metersnorth race
+#${timer} python filetimes.py ${1}.castra      dask    census meterswest metersnorth race
+${timer} python filetimes.py ${1}.castra      pandas  census meterswest metersnorth race
```
jbednar: Why no dask version of the test for castra anymore?

jbednar:
Looks good! Happy to merge after the one small comment is addressed. After merging, can you please update #129 to show the new performance for a similar set of tests as in the last comments in that issue, now that various libraries involved have been updated?  I can then run that same test on my own machine to see if I get similar results, to make sure the hardware is comparable.

gbrener:
Sure, no problem @jbednar . With the commit from a couple minutes ago, I also added some helpful debugging output when an extra argument is provided to `filetimes.sh`, or the `--debug` flag is provided directly to `filetimes.py`. I also moved the timing functionality closer to the actual function calls, to make it easier to rule-out unrelated code affecting the results.

examples/filetimes.sh
gbrener: I get `AttributeError: module 'dask.dataframe' has no attribute 'from_castra'` from filetimes.py. I'm pushing up a commit shortly that simply comments it out - perhaps that's a better approach.

gbrener:
One issue that's coming up, when executing the following command (note the `dftype='dask'` argument):
```
python -c "import filetimes as ft ; ft.p.base='census' ; ft.p.x='meterswest' ; ft.p.y='metersnorth' ; ft.p.categories=['race']; ft.timed_write('data/tinycensus.csv',dftype='dask')"
```
I get the following exception:
```
  File "<string>", line 1, in <module>
  File "/Users/gbrener/datashader/examples/filetimes.py", line 164, in timed_write
    duration, res = code(df,fname,p)
  File "/Users/gbrener/datashader/examples/filetimes.py", line 123, in <lambda>
    write["parq"]         ["dask"]   = lambda df,filepath,p:  benchmark(dd.io.parquet.to_parquet, (filepath, df)) ## **p.parq_opts
  File "/Users/gbrener/datashader/examples/filetimes.py", line 95, in benchmark
    res = fn(*posargs, **kwargs)
  File "/Users/gbrener/miniconda/envs/filetimes/lib/python3.5/site-packages/dask/dataframe/io/parquet.py", line 295, in to_parquet
    out = delayed(writes).compute()
  File "/Users/gbrener/miniconda/envs/filetimes/lib/python3.5/site-packages/dask/base.py", line 95, in compute
    (result,) = compute(self, traverse=False, **kwargs)
  File "/Users/gbrener/miniconda/envs/filetimes/lib/python3.5/site-packages/dask/base.py", line 202, in compute
    results = get(dsk, keys, **kwargs)
  File "/Users/gbrener/miniconda/envs/filetimes/lib/python3.5/site-packages/dask/threaded.py", line 76, in get
    **kwargs)
  File "/Users/gbrener/miniconda/envs/filetimes/lib/python3.5/site-packages/dask/async.py", line 500, in get_async
    raise(remote_exception(res, tb))
dask.async.TypeError: expected list of bytes

Traceback
---------
  File "/Users/gbrener/miniconda/envs/filetimes/lib/python3.5/site-packages/dask/async.py", line 266, in execute_task
    result = _execute_task(task, data)
  File "/Users/gbrener/miniconda/envs/filetimes/lib/python3.5/site-packages/dask/async.py", line 247, in _execute_task
    return func(*args2)
  File "/Users/gbrener/miniconda/envs/filetimes/lib/python3.5/site-packages/dask/compatibility.py", line 47, in apply
    return func(*args, **kwargs)
  File "/Users/gbrener/miniconda/envs/filetimes/lib/python3.5/site-packages/fastparquet/writer.py", line 580, in make_part_file
    rg = make_row_group(f, data, schema, compression=compression)
  File "/Users/gbrener/miniconda/envs/filetimes/lib/python3.5/site-packages/fastparquet/writer.py", line 568, in make_row_group
    compression=comp)
  File "/Users/gbrener/miniconda/envs/filetimes/lib/python3.5/site-packages/fastparquet/writer.py", line 453, in write_column
    bdata = encode['PLAIN'](pd.Series(data.cat.categories), selement)
  File "/Users/gbrener/miniconda/envs/filetimes/lib/python3.5/site-packages/fastparquet/writer.py", line 248, in encode_plain
    return pack_byte_array(list(out))
  File "fastparquet/speedups.pyx", line 110, in fastparquet.speedups.pack_byte_array (fastparquet/speedups.c:2570)
```

I don't think this is related to the changes in this branch - I suspect it's coming from a change in the newer version of Dask. Looking into it now.

gbrener:
Ok, I believe I've fixed the issue. The cause was a combination of two things:
1. The column names in `tinycensus.csv` don't include the index. Pandas and Dask usually detect the index column automatically, so this normally isn't a problem...
2. The `usecols=[...]` option was being passed to `dd.read_csv()` and `df.read_csv()`, which caused the resulting dataframes to ignore the index column and effectively shift each column one position to the left.
Because the columns were misidentified, Dask choked when the rightmost column (a "category" type) was being written out to parq format for values that were actually floating-point numbers.

jbednar:
I'd prefer to just remove the index column entirely from the CSV files, as this will make them smaller, make them match other CSV files more closely, and will let us use `usecols`, which is very important for performance for multi-column datasets.

Issue #305, opened 2017-03-24
Add memory usage benchmarks


gbrener:
Adding benchmarks for memory usage could illuminate details about datashader's runtime performance that were previously not being measured. Among other things we should benchmark memory usage for different dataset sizes, and see how memory usage changes as datasets grow.

gbrener:
# Memory usage takeaways

With memory usage measurements in place, we can start validating the dataframe size and how it corresponds to memory usage in the overall datashader benchmarks. The following statements were added to filetimes.py immediately before the first call to `timed_agg`, to display the dataframe's per-column memory usage and data types:
```
mem_usage = df.memory_usage(deep=True)
if p.dftype == 'dask':
    mem_usage = mem_usage.compute()
print('DEBUG:', mem_usage)
mem_usage_total = mem_usage.sum()
print('DEBUG: DataFrame size:\t\t\t{} MB'.format(mem_usage_total / 1e6))
for colname in df.columns:
    print('DEBUG: column "{}" dtype: {}'.format(colname, df[colname].dtype))
```

The results show that the memory usage of the persisted dask dataframe align with what we'd expect:
```
$ python filetimes.py times/census.parq dask census meterswest metersnorth race --debug --cache=persist
DEBUG: Cache "persist" mode enabled
DEBUG: Memory usage (before read):      135.733248 MB
DEBUG: read_parquet(times/census.parq, index=False, columns=['meterswest', 'metersnorth', 'race'])
DEBUG: Force-loading Dask dataframe
DEBUG: Memory usage (after read):       5997.34272 MB
DEBUG: Index                 560
metersnorth    2453400032
meterswest     2453400032
race            306677314
dtype: int64
DEBUG: DataFrame size:                  5213.477938 MB
DEBUG: column "meterswest" dtype: float64
DEBUG: column "metersnorth" dtype: float64
DEBUG: column "race" dtype: category
DEBUG: Memory usage (after agg1):       7549.800448 MB
DEBUG: Memory usage (after agg2):       8219.004928 MB
times/census.parq            dask    Aggregate1:017.39 (009.76+007.63)  Aggregate2:003.94  In:05213483268  Out:00000625044  Total:026.49
```

Let's verify these calculations:
- Number of rows of data in dataset: 306675004
- `metersnorth` and `meterswest` are each represented as 8-byte double-precision floats (`np.float64`); our expected size for each column is 8 x 306675004 = 2453400032 (correct)
[...]

PR #309, opened 2017-04-12
Add memory profiling, datashader-data S3 bucket

gbrener:
- The census.zip archive in the `bokeh_data` S3 bucket now contains a census.parq file that was written out uncompressed with a categorical dtype for the "race" column. Will shortly create a new `datashader-data` bucket and move this file into it (see TODOs below).

- filetimes.py now has some rudimentary memory profiling features which measure:
    a) the memory of the overall process (max resident set size) at various points in the program, and
    b) the memory of the dataframe after it's read into memory

- A CSV file-detection issue is fixed (it was happening when dask wrote out the CSV files)

This PR is still a work-in-progress:
- [x] Create a new S3 bucket called `datashader-data`. Move the files that only relate to datashader from `bokeh_data` to the new S3 bucket. Underscores are no longer allowed in S3 bucket names, so the underscore is now a dash
- [x] Investigate runtime performance issue for Agg2
- [x] Investigate memory issue that happens after the first `timed_agg()`
- [x] Update `datashader-data` bucket with a SNAPPY-compressed version of census.parq, provided that it ends up being faster to load than the uncompressed version

### DIFF
--- a/examples/filetimes.py
+++ b/examples/filetimes.py
@@ -49,7 +49,7 @@ p=Parameters()
 filetypes_storing_categories = {'parq','castra'}
 
 
-class Kwargs(dict):
+class Kwargs(odict):
     """Used to distinguish between dictionary argument values, and
     keyword-arguments.
     """
--- a/examples/filetimes.py
+++ b/examples/filetimes.py
@@ -261,10 +261,10 @@ def main(argv):
 
     if DEBUG:
         print('DEBUG: Memory usage (after read):\t{} bytes'.format(resource.getrusage(resource.RUSAGE_SELF).ru_maxrss, flush=True))
-        mem_usage = df.memory_usage(deep=True)
-        print('DEBUG: DataFrame size:\t\t\t{} bytes'.format(mem_usage.sum(), flush=True))
     img,aggtime1 = timed_agg(df,filepath,5,5)
     if DEBUG:
+        mem_usage = df.memory_usage(deep=True)
+        print('DEBUG: DataFrame size:\t\t\t{} bytes'.format(mem_usage.sum(), flush=True))
         print('DEBUG: Memory usage (after agg1):\t{} bytes'.format(resource.getrusage(resource.RUSAGE_SELF).ru_maxrss, flush=True))
 
     img,aggtime2 = timed_agg(df,filepath)
--- a/examples/filetimes.py
+++ b/examples/filetimes.py
@@ -206,6 +206,10 @@ def get_size(path):
     return total
 
 
+def get_proc_mem():
+    return resource.getrusage(resource.RUSAGE_SELF).ru_maxrss / 1e6
+
+
 def main(argv):
     global DEBUG, DD_FORCE_LOAD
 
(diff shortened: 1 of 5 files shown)


jbednar:
Presumably the file to download should be compressed, with whichever compression type has the best performance?  I don't think we should be distributing an uncompressed version, unless there is a major performance benefit (in which case we need to compress it using zip to reduce transfer time).

jbednar:
The memory measurement looks good, thanks!  Is it practical to put the memory usage into the same output format rather than debug statements, or would that be too overwhelming in the current table format?  I guess we could collect the results into a pandas dataframe and display that more compactly in a Jupyter notebook, but maybe that's overkill.

gbrener:
@jbednar - yes, the file is compressed and is technically called "census.zip", compressed down to 1.2GB or so. It expands as the "census.parq" directory, which inflates to something like 4.4GB. I haven't given much thought to the output format, but I think at some point it would be nice if we had some Bokeh histograms. Planning on addressing that once I get to the bottom of the perf/memory usage issues.

jbednar:
In addition to/instead of the .zip compression, the .parq file itself should be compressed if that's the fastest performance.  Last I checked, snappy compression was the fastest, much faster than uncompressed, and if that's still true that's what should be in the bucket.

gbrener:
Updated Issue #129 with cachey results (obtaining the sub-2-second agg2 time). Also using this opportunity to link with Issue #305 .

examples/filetimes.py
```diff
@@ -90,7 +90,7 @@ def benchmark(fn, args, filetype=None):
         if p.dftype == 'dask' and DD_FORCE_LOAD:
             if DEBUG:
                 print("DEBUG: Force-loading Dask dataframe", flush=True)
-            df = res.persist()
+            res = res.persist()
```
jbednar: Just using ``res.persist()`` might be clearer, to make it obvious that persist acts by side effects.

gbrener: After looking into things a bit more, I found that `persist()` does not act by side-effect only. It a makes a copy of the graph: https://github.com/dask/dask/blob/master/dask/base.py#L457-L460 , modifies the copy, and returns it.

jbednar: So it does *not* use side effects?  I thought your comments above said it did.  Anyway, if it works as you say here, the code is fine.

gbrener:
My latest commit adds a large speed improvement, with `df.persist()` boosting Agg2 to ~5 seconds for the dask/parquet combination. Using dask's cache feature still outperforms this by about 2x, and more investigation needs to be done into the dask internals to figure out why this is happening. I'm picking up the [diagnostics tools](http://dask.pydata.org/en/latest/diagnostics.html) based on @jcrist suggestion to gather more information.

gbrener:
I believe this PR is ready to be merged.

examples/filetimes.py
gbrener: I was incorrect in my comment above.

examples/filetimes.py
```diff
@@ -178,12 +175,6 @@ def timed_read(filepath,dftype):
     filetype=extension.split(".")[-1]
     code = read[extension].get(dftype,None)
 
-    if filetype=="csv":
-        if dftype=="dask":
-            filepath = filepath.replace(".csv","*.csv")
-        else:
-            filepath = glob.glob(filepath.replace(".csv","*.csv"))[0]
-
```
jbednar: Does this need to be restored to make the CSV "In:" size correct?

jbednar:
Merging so that we can move forward, but please use cachey by default until we can figure out another way to have better performance, because I want this file to show the way to get the best performance, not something middling like currently found with `persist`.

PR #312, opened 2017-04-18
Lidar notebook example - Cleanup and simplify PR #301

PeterDSteinberg:
I did a number of experiments with the LiDAR data source and downloading options, resulting in unclear commit history on PR #301.  This is a simplification that includes only the relevant final changes:

 * Creates`examples/lidar_dask.ipynb`
   * Seattle area Lidar datashader figure with tile background
 * Modifies `examples/download_sample_data.py `
   * Installs `pylzma` so that 7zip files can be unpacked
 * Modifies `examples/datasets.yml`
   * Now a dataset item may include a key `func` with a value that is a string function name from `download_sample_data.py`
 * Modifies `README.md` for [earth-env](https://github.com/ContinuumIO/earth-env), a env for geographic data tools like rasterio


PeterDSteinberg:
@jbednar I made a local mistake during rebase (as suggested on #301), so I just opened new PR with fewer commits.

examples/download_sample_data.py
```diff
@@ -24,23 +24,28 @@
...
     sys.exit(1)
 
-from collections import OrderedDict
-
-from os import path
+try:
+    from py7zlib import Archive7z
+except:
+    subprocess.check_output(['pip', 'install', 'pylzma'], env=os.environ)
+    from py7zlib import Archive7z
```
jbednar: I'm a bit uncomfortable modifying the user's environment behind the scenes in this way.

I would favor adding pylzma to environment.yml instead, and then there testing if it's installed, and just skipping the lidar example data altogether if it's not.

examples/download_sample_data.py
```diff
@@ -217,6 +233,40 @@ def _process_dataset(dataset, output_dir):
...
+    output_paths = [path.join(here, 'data', os.path.basename(fname))
+                    for fname in files]
+    title_fmt = 'Puget Sound Lidar {} of {}'
+    for idx, (url, output_path) in enumerate(zip(urls, output_paths)):
+        if os.path.exists(output_path.replace('7z', 'gnd')):
+            print('Skipping {0}'.format(title_fmt.format(idx + 1, len(urls))))
+            continue
+        _url_to_binary_write(url, output_path, title_fmt.format(idx + 1, len(urls)))
+        _unzip_7z(output_path, delete_7z=True)
+
```
jbednar: It's a shame that we need such dataset-specific code in here.  I haven't studied it to see how it works and precisely what it is accomplishing, but maybe it could be converted into a general function that accepts an argument, with the argument supplied based on data in the datasets.yml file?  I.e., perhaps change "title" in the .yml from "Puget Sound LiDAR Consortium" to "Puget Sound LiDAR", and then rename this function (and add a docstring) to explain concretely what it's doing with the files, rather than just saying it's about lidar and leaving it mysterious.

examples/download_sample_data.py
PeterDSteinberg: This `pylzma` install has been moved to `environment.yml` in the recent commits.

examples/lidar_dask.ipynb
```diff
@@ -0,0 +1,270 @@
+{
```
PeterDSteinberg: Fixed in latest commits

README.md
```diff
@@ -50,6 +50,22 @@ Datashader is not currently provided on pip/PyPI, to avoid broken or
...
+The command above will download the `earth-env-36` environment from [anaconda.org](http://anaconda.org).  `earth-env` aims to solve version conflicts among gdal and rasterio and other file readers for GeoTiff, NetCDF, HDF4, HDF5 formats.  The downloaded and installed conda environment can optionally be named with the `--name` argument to `conda env create`.
+
+The `earth-env` environment `yaml` specification files can be downloaded with `anaconda` as follows (change the `-36` part for other Python versions):
+
+```
+anaconda download psteinberg/earth-env-36
+```
+
+Issue tracking and source for earth-env are [in the earth-env github repo](https://github.com/ContinuumIO/earth-env/issues).
+
```
PeterDSteinberg: Changed the notes at the top of the notebook to use the `ds` environment created by that yml.

examples/download_sample_data.py
PeterDSteinberg: I went through a documented all the funcs in `download_sample_data.py` and tried to simplify more.  There are now two ways a dataset may be specified in datasets.yml:

* Same `files, url, title` keys pattern we already had in there.  This means download an archive from `url` if not all `files` are present already, then unpack the contents from `url` to create the `files`.
* `files, url, title, tag` keys pattern used by this lidar data example.  `tag` is used to separate special cases.  In the `tag="lidar"` case, a util is called to replace the url pattern to form 25 urls of 7z files, then download each and un7zip each.  The alternative to this special logic would have been to make 25 entries in `datasets.yml` and leave `download_sample_data.py` as it was, but I thought that would make `datasets.yml` less clear.

examples/datasets.yml
```diff
@@ -34,8 +34,10 @@ data:
       - MERCATOR_LC80210392016114LGN00_BQA.TIF
 
   - url: http://pugetsoundlidar.org/lidardata/newdata/2016/PSLC_KingCounty_Delivery1/BE_ASCII.html
-    title: "Puget Sound LiDAR Consortium"
-    func: download_puget_sound_lidar
+    title: "Puget Sound Lidar"
```
jbednar: Gulp!  Apparently the issue of how to capitalize Lidar is complicated!

image

I guess I favor Lidar, so this change is fine.

README.md
jbednar: Perfect, thanks!

PeterDSteinberg: I just now pushed a commit to change the README a bit (previously I responded to this comment only changing the notebook's mention of earth-env)

jbednar:
Thanks for the improvements!  I think this is much better.  However, I still don't see why there needs to be lidar-specific code.  Would the following alternative approach work?

1. Automatically unpack any .7z files downloaded (seems like that would already work now)
2. Specify the URL as `http://pugetsoundlidar.org/lidardata/newdata/2016/PSLC_KingCounty_Delivery1/BE_ASCII/`
3. Have a special check that if the URL has a trailing slash, and if so, fetch each of the listed files from that URL
4. Handle the file extension in some general way -- somehow it's currently deciding to use ".gnd", but (a) shouldn't that be done by 7z (not us forcing it?), and (b) shouldn't it just be ".csv" if we're forcing it (since that appears to be all it is)?

I *think* that would work as well as the current approach, while removing anything lidar-specific, and I believe this approach would also work for the taxi dataset, which was originally also a bunch of compressed files (though we've repackaged it for simplicity).

That said, I think there's still an issue -- the files listed are .7z in datasets.yml, but these filenames are used to check to see if data needs to be downloaded, so shouldn't they be listed as .gnd or .csv?  In which case the above plan would need to be modified to include a .yml field for the expected extension, which the script would need to replace.

Sorry to drag this out, but I strongly believe we will add many more datasets over time, and I want to make sure the machinery for this doesn't blow up out of our ability to maintain it.  I also want people to see clearly what steps we are doing so that they can see how to prepare their own datasets for use with datashader.

PeterDSteinberg:
Sounds good, @jbednar , shouldn't take long to make those changes.

jbednar:
Great, thanks.  BTW, it might just be simpler to make the "URL:" field check to see if it's a list or a single item, and if it's a list, to download them all.  I don't mind datasets.yml having a long list; lists are better than code!

PeterDSteinberg:
I think the comments are addressed at this point, @jbednar

jbednar:
Looks great, thanks!  I don't think any of these changes affect the tests, so I'll merge now.  Thanks so much for sticking with me on this!

PR #318, opened 2017-04-24
Fix bogus column and row at edge

### LINKED ISSUE #259: Bogus column and row at edge?
There seems to be an empty column and row at the edge of aggregated arrays:

```python
import pandas as pd
import numpy as np
import datashader as ds

df = pd.DataFrame(dict(x=np.random.normal(0,10,10000),y=np.random.normal(0,10,10000)))

canvas = ds.Canvas(plot_width=8, plot_height=8, x_range=(-4,4), y_range=(-4,4))
agg = canvas.points(df, 'x', 'y')
agg.values
```

```
array([[21, 13, 21, 26, 17, 18, 20,  0],
       [22, 16, 28, 23, 17, 16, 23,  0],
       [17, 20, 22, 20, 20, 19, 26,  0],
       [16, 14, 28, 14, 12, 16, 22,  0],
       [16, 17, 16, 25, 19, 18, 18,  0],
       [ 9, 24, 24, 24, 17, 14, 17,  0],
       [19, 24, 17, 19, 13, 12, 21,  0],
       [ 0,  0,  0,  0,  0,  0,  0,  0]], dtype=int32)
```

The extra row and column ends up being transparent, at least for counts, so it has little effect, but it still seems like we should be returning a fully filled-out array.

jbcrail:
I applied @jbednar's changes to make the upper bound exclusive.

I also updated the test data to reflect the new upper bound semantics. In particular, the log axis tests required a seemingly magic value, `x_max_index`. This value results from scaling, translating, and indexing the original x range from `(1, 11)` to `(1, sqrt(11))`. To account for precision errors, I calculated the final value using the same formula used by Datashader.

Fix #259

### DIFF
--- a/datashader/tests/test_dask.py
+++ b/datashader/tests/test_dask.py
@@ -1,11 +1,13 @@
+from dask.async import get_sync
+from dask.context import set_options
+import dask.dataframe as dd
 import numpy as np
 import pandas as pd
 import xarray as xr
-import dask.dataframe as dd
-from dask.context import set_options
-from dask.async import get_sync
 
 import datashader as ds
+
+import pytest
 
 set_options(get=get_sync)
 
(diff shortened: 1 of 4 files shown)


datashader/tests/test_dask.py
```diff
@@ -152,14 +154,16 @@ def test_multiple_aggregates():
 
 
 def test_log_axis():
+    # Upper bound for scale/index of x-axis
+    x_max_index = 10 ** (1 / (2 / np.log10(11)))
     sol = np.array([[5, 5], [5, 5]], dtype='i4')
```
jbednar: Thanks for tracking that calculation down!

datashader/tests/test_pandas.py
```diff
@@ -70,7 +72,7 @@ def test_sum():
 
 
 def test_min():
-    out = xr.DataArray(df.i64.reshape((2, 2, 5)).min(axis=2).astype('f8').T,
+    out = xr.DataArray(df.i64.values.reshape((2, 2, 5)).min(axis=2).astype('f8').T,
                        coords=coords, dims=dims)
```
jbednar: What's the implication of this change?

jbcrail: It removes the deprecation warning about `reshape`. It looks like it was deprecated in Pandas 0.19.0.

jbcrail:
This is ready for merging.

philippjfr:
@jbcrail We've been seeing crashes in the notebook since this PR was merged. I did a git bisect and tracked it down to this commit:

[68d51fe6cae0dad7439ba03cda21cfef256b9f5a] Fix off-by-one error

You can reliably reproduce this error by executing this code twice in a notebook:

```python
import pandas as pd
import datashader as ds

data = pd.read_csv('crash.csv')

data['k'] = data['k'].astype('category')

cvs = ds.Canvas(plot_width=400,
                plot_height=400,
                x_range=(data['x'].min(), data['x'].max()),
                y_range=(data['y'].min(), data['y'].max()))

cvs.line(data, 'x', 'y', ds.count_cat('k'))
```

You can download the crash.csv [here](http://philippjfr.com/ipynbs/crash.csv). I suspect it's because ``extend_line`` wasn't updated in this PR.

jbcrail:
@philippjfr I think you're right. Thanks for catching this. I'll open an issue and fix it.

PR #319, opened 2017-04-24
Add tests for Python 3.6

jbcrail:


jbednar:
It would probably be ok to skip any tests that have to do with gdal on py3.6; no reason not to test the rest just because gdal isn't available.  Of course, if it's a matter of specifying a different channel so we can get gdal; then we can just do that...

jbcrail:
I ran the tests locally and noticed that gdal wasn't needed, so I removed it from the Travis requirements. We'll see if it passes the full tests without it.

PR #320, opened 2017-04-24
Add code removed from Bokeh 0.12.5 into InteractiveImage class

StevenCHowell:
This PR supersedes #294, serving to patch #293.  @philippjfr, please verify that this resolves the `InteractiveImage` issue when using Bokeh 0.12.5 (or the 0.12.6 dev version).

### DIFF
--- a/datashader/bokeh_ext.py
+++ b/datashader/bokeh_ext.py
@@ -186,6 +186,11 @@ class InteractiveImage(object):
         self.p.x_range.callback = callback
         self.p.y_range.callback = callback
 
+        # Initialize document
+        doc_handler = _AddToDocument(self.p)
+        with doc_handler:
+            self.doc = doc_handler._doc
+            self.div = notebook_div(self.p, self.ref)
 
     def _init_callback(self):
         """
@@ -247,7 +252,6 @@ class InteractiveImage(object):
         msg = self.get_update_event()
         self.comms_handle.comms.send(msg)
         return 'Complete'
-
 
     def get_update_event(self):
         """
--- a/datashader/bokeh_ext.py
+++ b/datashader/bokeh_ext.py
@@ -18,6 +18,57 @@ from . import transfer_functions as tf
 from .utils import downsample_aggregate, summarize_aggregate_values
 
 
+class _AddToDocument(object):
+    # 'models' can be a single Model, a single Document, or a list of either
+    def __init__(self, models):
+
+        self._to_remove_after = []
+        if not isinstance(models, list):
+            models = [models]
+
+        self._doc = _find_some_document(models)
+        if self._doc is None:
+            # oh well - just make up a doc
+            self._doc = Document()
+
+        for model in models:
+            if isinstance(model, Model):
+                if model.document is None:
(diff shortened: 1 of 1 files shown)


philippjfr:
@jbednar What's the best way to test this? Do all InteractiveImage plots have weird interactions right now or was it just under certain conditions?

jbednar:
Hmm; I don't know.  I just tested datashader master (before this PR) using bokeh 0.12.5 and could find no issues with the nyc_taxi.ipynb notebook, so I can't tell what the problem was.

StevenCHowell:
The original problem, described in #293, was that [`datashader/bokeh_ext.py`](https://github.com/bokeh/datashader/blob/6101791e876b4cc8db24a5b3c1a457b21987d098/datashader/bokeh_ext.py#L10) imported `_ModelInDocument` from `bokeh.model` but this was removed in Bokeh 0.12.5.  @philippjfr removed this import on in PR #294 (changes made on March 8 and merged on March 13), [here is the updated code](https://github.com/bokeh/datashader/blob/841406f5162932f230faacd6034a70733ebc6280/datashader/bokeh_ext.py).  

The old version was using [`_ModelInDocument` from `bokeh.model`](https://github.com/bokeh/datashader/blob/6101791e876b4cc8db24a5b3c1a457b21987d098/datashader/bokeh_ext.py#L10) to [defile `self.doc`](https://github.com/bokeh/datashader/blob/6101791e876b4cc8db24a5b3c1a457b21987d098/datashader/bokeh_ext.py#L138-L140).

The new version uses [`Document` from `bokeh.document`](https://github.com/bokeh/datashader/blob/841406f5162932f230faacd6034a70733ebc6280/datashader/bokeh_ext.py#L9) to [define `self.doc`](https://github.com/bokeh/datashader/blob/841406f5162932f230faacd6034a70733ebc6280/datashader/bokeh_ext.py#L229).

I just setup an environment with Bokeh 0.12.5, installed Datashader from source, then created an interactive image.  It worked fine; I did not get any errors and the pixel averaging updates as I zoom in/out.  It seems like @philippjfr's changes on March 8, #296, resolved the issue (though there were comments in the PR about persistent uncertainty about the API stability).  What led to the conclusion that PR #294 "is required afterall"?

jbednar:
> What led to the conclusion that PR #294 "is required afterall"?

I can't tell any reason this PR is required, so I guess don't merge it unless someone demonstrates something not working with bokeh 0.12.5.

philippjfr:
>What led to the conclusion that PR #294 "is required afterall"?

@StevenCHowell Someone reported seeing strange interactions between multiple documents, which Bryan mentioned could occur using the approach in master under certain circumstances. @jbednar Said he could reproduce this at some point but it seems that is no longer the case.

jbednar:
I certainly can't reproduce right now, and I can't remember under what conditions I saw an issue before.  The nyc_taxi.ipynb notebook should have been a good test of this, as it has very many InteractiveImages in different cells.  Let's leave this open until we see a problem in practice or it gets old enough that we don't think there will be one.

Issue #313, opened 2017-04-19
General performance optimizations


gbrener:
# Tasks
For the past couple weeks I've been investigating datashader's performance and how we can improve upon it. I'm now documenting my remaining tasks, in case I get pulled away on a different project. Below is a list of tasks/issues that I'm currently addressing:
- [x] Extend the filetimes.py and filetimes.yml benchmarking environment to find the optimal file format for datashader/dask (Issue #129)
- [x] Benchmark numba compared to handwritten ufuncs in vaex (Issue #310)
- [x] Gather perf information about dask locking behavior (Issue #314)
- [x] Investigate why Cachey leads to better runtime performance for repeat datashader aggregations
- [x] Document memory usage findings (Issue #305)
- [x] Investigate how datashader's performance changes with data types (doubles vs floats, etc) (Issue #305)
- [x] Verify that repeat aggregations no longer depend on file format (Issue #129)
- [x] Investigate distributed scheduler vs threaded scheduler for single-machine use case (#331, #332, #334)
- [x] Identify issues hindering distributed scheduler from performing more effectively - credit goes to @martindurant ( #332, #336, #337 )

# Performance takeaways
Below are some performance-related takeaways that fell out of my experiments and optimizations with datashader and dask:
## General
- Use the latest version of numba (>=0.33). This includes bugfixes providing ~3-5x speedups for many cases (numba/numba#2345, numba/numba#2349, numba/numba#2350)
- When interacting with data on the filesystem, store it in the [Apache Parquet](https://parquet.apache.org/) format when possible. Snappy compression should be used when writing out parq files, and the data should rely on categorical dtypes (when possible) before writing the parq files, as parquet supports categoricals in its binary format (#129)
- Use the categorical dtype for columns with data that takes on a limited, fixed number of possible values. Categorical columns use a more memory-efficient data representation and are optimized for common operations such as sorting and finding uniques. Example of how to convert a column to the `categorical` dtype:

  ```
  df[colname] = df[colname].astype('category')
  ```

- There is promise with enhancing datashader's performance even further by using single-precision floats (`np.float32`) instead of double-precision floats (`np.float64`). In past experiments this cut down the time to load data off of disk (assuming the data was written out in single-precision float) as well as datashader's aggregation times. Care should be taken using this approach, as using single-precision (in any software application, not just datashader) leads to different numerical results than double-precision (#305)

[...]

PR #323, opened 2017-04-25
Add latest numba build + debugging stmts for memory usage

gbrener:
This PR includes the recent numba dev build, which contains some performance fixes (useful for benchmarking), into the benchmarking/filetimes environment.

Also included are some additional debug/print statements related to memory usage, per issue #313 and issue #313 .

A flag is added to `timed_write()` so we can control the floating-point precision when writing out files, and see how it affects both file-size, write performance, and (more importantly) read performance.


examples/filetimes.py
```diff
@@ -230,7 +238,8 @@ def main(argv):
     else:
         if args.cache == 'cachey':
             from dask.cache import Cache
-            Cache(p.cachesize).register()
+            cache = Cache(p.cachesize)
+            cache.register()
```
jbednar: What's the effect of this change?

gbrener: This makes it easier to inspect cachey's results.

PR #329, opened 2017-04-27
Fix issues with csv benchmarks

gbrener:
The "filetimes" benchmarks have had some issues with CSV files in the past, due to the fact that dask writes out CSV data as separate files (due to partitioning), but pandas writes the CSV data as a single large file. This PR corrects the file size calculations to account for all CSV files (when dask writes them out), and the benchmarks will involve loading all CSVs, regardless of whether they were written out by dask or pandas.

This has the implication that when pandas writes out the CSV data, pandas should have an advantage at CSV read-time, and when dask writes out CSV data, dask will have an advantage at CSV read-time (pandas takes an additional "concatenation" step after reading in each file).

### DIFF
--- a/examples/filetimes.py
+++ b/examples/filetimes.py
@@ -98,7 +98,7 @@ def benchmark(fn, args, filetype=None):
     
 
 
-read = odict([(f,odict()) for f in ["parq","bcolz","feather","castra","h5","csv"]])
+read = odict([(f,odict()) for f in ["parq","snappy.parq","gz.parq","bcolz","feather","castra","h5","csv"]])
 
 def read_csv_dask(__filepath, usecols=None):
     # Pandas writes CSV files out as a single file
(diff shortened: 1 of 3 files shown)


gbrener:
It appears that the tests are failing due to a segmentation fault that is unrelated to the changes in this PR.

examples/filetimes.py
```diff
@@ -132,6 +147,7 @@ def write_castra_pandas(__filepath, __df, __cats):
 write["bcolz"]        ["pandas"] = lambda df,filepath,p:  benchmark(bcolz.ctable.fromdataframe, (df, Kwargs(rootdir=filepath)))
 write["feather"]      ["pandas"] = lambda df,filepath,p:  benchmark(feather.write_dataframe, (df, filepath))
 write["parq"]         ["pandas"] = lambda df,filepath,p:  benchmark(fp.write, (filepath, df, Kwargs(**p.parq_opts)))
+write["gz.parq"]  ["pandas"] = lambda df,filepath,p:  benchmark(fp.write, (filepath, df, Kwargs(compression='GZIP', **p.parq_opts)))
```
jbednar: Please compare GZIP to Snappy file times, in case we should be using GZIP instead!

jbednar: (not as part of this PR)

jbednar:
These changes look good, thanks!  Very annoying stuff that I'm glad I didn't have to do.  Thanks for taking the hit!

PR #330, opened 2017-04-27
Adjust line to account for exclusive upper bounds

### LINKED ISSUE #328: Segmentation fault when rendering line
As noted by @philippjfr, the off-by-one PR introduced a segmentation fault.

See the comments in PR #318 for more details.

jbcrail:
When extending a line, I decremented the upper bounds for both x and y.

Fix #328


jbcrail:
I tried two strategies for this fix. I focused on `extend_line` since it was the only method that used the bounds, instead of just passing the value along. My intent with both strategies was to shorten the bounding box for the clipped area.

First, I changed the comparisons in `_compute_outcode` for both maximum bounds from `>` to `>=`. This seemed like the least intrusive approach while still accounting for the exclusive upper bounds. Unfortunately, the while loop in `extend_line` never exited for several tests.

Second, I decremented the maximum bounds prior to clipping the segment. This approach worked better and is the one in this PR. The modified tests accounted for the newly clipped segments at the exclusive upper bounds. For instance, one test had a top-left to bottom-right line segment that was clipped in the bottom-right corner as expected after the fix.

I missed the change to the slope of the line segment. Ideally, I'd prefer the first strategy but I don't know the code or the algorithms well enough to know what changes are needed to make it work.

jbednar:
Let's go ahead and merge this to avoid the segfaults, but I'd still like to compare to an example worked out on paper before release.

PR #331, opened 2017-04-28
Add --distributed option to benchmark + fix benchmarks for feather and bcolz

gbrener:
Benchmarks for bcolz and feather have not worked properly for some time. This PR adds fixes for both so that we can view them for comparison. It also removes the "castra" format, since there's no longer first-class support for it in dask.


gbrener:
There's a new option to the script called `--distributed`, which will allow us to compare performance between the distributed scheduler in dask vs the threaded one we've been using so far. Preliminary results show that, when `df.persist` is not used and all data fits in RAM on one machine, the distributed scheduler is more performant than the threaded one. With `df.persist()`, however, the threaded one outperforms distributed (in some cases) by nearly a factor of 3.

Issue #332, opened 2017-04-28
Reduced performance with distributed scheduler with client.persist()


gbrener:
The following code illustrates an issue with dask's distributed scheduler and the client.persist(df) functionality:

```
from dask import distributed
import multiprocessing
cluster = distributed.LocalCluster(n_workers=multiprocessing.cpu_count(), threads_per_worker=1)
dask_client = distributed.Client(cluster)
dask_df = dd.from_pandas(df, npartitions=multiprocessing.cpu_count())
...
cvs = datashader.Canvas(...)
agg = cvs.points(dask_df, ...)
```

When we insert the following line before `agg = cvs.points(daskdf, ...)`:
```
dask_df = dask_client.persist(dask_df)
```

The runtime performance of `cvs.points()` gets much worse, regardless of dataset size.

martindurant:
To be clear, `persist` should improve performance when there is more than one computation on the same data, to prevent having to load it into memory more than once. In the above example, the data is dead baggage after it has been used, and would have been freed if not for the persist. Furthermore, not sure if this is the case, but if the aggregation touches only some of the data, persist would still have loaded all of it.

gbrener:
I'm seeing that in both cases - when the agg touches some of the data, and when it touches all of the data - the performance is reduced.

mrocklin:
@gbrener note that persist is non-blocking.  If you are profiling it might be worthwhile to call `distributed.wait(dask_df)` after calling persist to wait until dask finishes loading in that data.

gbrener:
Thanks for the tip @mrocklin. I tried your suggestion and the issue remains. Here is more output, comparing two runs with the distributed scheduler, one with `dask_df = client.persist(df); distributed.wait(dask_df)`, and the other with no caching/waiting:

With persisting/waiting:
```
$ python filetimes.py times/census.snappy.parq  dask    census meterswest metersnorth race --debug --cache=persist --distributed
DEBUG: Cache "persist" mode enabled
DEBUG: "distributed" scheduler is enabled
DEBUG: Memory usage (before read):      141.836288 MB
DEBUG: read_parquet(times/census.snappy.parq, columns=['meterswest', 'metersnorth', 'race'], index=False)
DEBUG: Memory usage (after read):       143.29856 MB
DEBUG: Index                 560
metersnorth    2453400032
meterswest     2453400032
race            306677314
dtype: int64
DEBUG: DataFrame size:                  5213.477938 MB
[...]

PR #334, opened 2017-05-01
Improvements to distributed scheduler benchmarks

gbrener:
Add `distibuted.wait(dask_df)` after `client.persist(dask_df)` for distributed scheduler, based on suggestion from @mrocklin in #332 .

### DIFF
--- a/examples/filetimes.py
+++ b/examples/filetimes.py
@@ -95,6 +95,7 @@ def benchmark(fn, args, filetype=None):
                 #   aggregation performance (both --distributed and
                 #   --cache=persist were provided)
                 res = DASK_CLIENT.persist(res)
+                distributed.wait(res)
             else:
                 if DEBUG:
                     print("DEBUG: Force-loading Dask dataframe", flush=True)


jbednar:
Hmm.  Does this change slow things down?  It will help us sort the time into categories more accurately, so we should do it if that's the only effect, but we should be careful in case this ends up blocking computation that in practice would be proceeding asynchronously.

gbrener:
This change actually helps performance of the distributed scheduler when used with `client.persist()` on a single machine. The blocking doesn't happen when client.persist is not used.

jbednar:
Really?  Interesting.  I'll merge this then.

But how much does it improve?  If it's just a little, that's fine.  If it's a lot, it would need to go into the recommendations listed in that issue.

gbrener:
Thanks. It improves by nearly 3x, but I think this is because there's an issue with dask's distributed scheduler interacts with the client.persist feature. My main concern is that dask's distributed scheduler has the best performance when there is no persistence/caching whatsoever, regardless of data size. So that's what I put as the recommendation in #313.

PR #344, opened 2017-05-05
Optimize range calculation operations

gbrener:
Optimize the operations that datashader uses to calculate the x_range and y_range variables when the ranges are not provided by the user. Addresses issue https://github.com/bokeh/datashader/issues/336 and some comments that came out of https://github.com/bokeh/datashader/issues/332 . Agg2 now takes the same amount of time with df.persist() as it had previously with cachey.

~~Add a caching feature to Canvas which allows it to avoid recalculating the x_range and y_range when the Canvas object is reused for multiple aggregations. This caching feature can be turned off when instantiating the Canvas object `Canvas(..., cache_ranges=False)`, or be circumvented by calling `cvs.points(..., recalc_ranges=True)` or `cvs.line(..., recalc_ranges=True)`.~~

Update `filetimes.py` to do its own caching of the range variables. Also add a `--recalc-ranges` flag to see timing difference when the range variables are not cached.

### DIFF
--- a/datashader/core.py
+++ b/datashader/core.py
@@ -138,35 +138,23 @@ class Canvas(object):
     plot_width, plot_height : int, optional
         Width and height of the output aggregate in pixels.
     x_range, y_range : tuple, optional
-        A tuple representing the bounds inclusive space ``[min, max]`` along the
-        axis. These will be calculated and cached by datashader during the first
-        aggregation, if not provided. The cached values may be
-        invalidated/recalculated by providing the ``recalc_ranges`` keyword
-        argument to the ``Canvas.points()`` and/or ``Canvas.line()`` methods.
+        A tuple representing the bounds inclusive space ``[min, max]`` along
+        the axis.
     x_axis_type, y_axis_type : str, optional
         The type of the axis. Valid options are ``'linear'`` [default], and
         ``'log'``.
-    cache_ranges : bool, optional
-        Whether to cache x_range and y_range variables when they get
-        updated. Default is ``True``. This provides a speedup by relying on the
-        x and y data remaining unchanged between data aggregations. If the x or
-        y data may have changed, one may either set this option to ``False``, or
-        provide the ``recalc_ranges=True`` when calling ``Canvas.points()``
-        and/or ``Canvas.line()``.
     """
     def __init__(self, plot_width=600, plot_height=600,
                  x_range=None, y_range=None,
-                 x_axis_type='linear', y_axis_type='linear',
-                 cache_ranges=True):
+                 x_axis_type='linear', y_axis_type='linear'):
         self.plot_width = plot_width
         self.plot_height = plot_height
-        self.x_range = tuple(x_range) if x_range is not None else None
-        self.y_range = tuple(y_range) if y_range is not None else None
+        self.x_range = None if x_range is None else tuple(x_range)
+        self.y_range = None if y_range is None else tuple(y_range)
         self.x_axis = _axis_lookup[x_axis_type]
         self.y_axis = _axis_lookup[y_axis_type]
-        self.cache_ranges = cache_ranges
 
-    def points(self, source, x, y, agg=None, recalc_ranges=False):
+    def points(self, source, x, y, agg=None):
         """Compute a reduction by pixel, mapping data to pixels as points.
 
         Parameters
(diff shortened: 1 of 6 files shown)


datashader/core.py
```diff
@@ -138,23 +138,35 @@ class Canvas(object):
...
                  x_range=None, y_range=None,
-                 x_axis_type='linear', y_axis_type='linear'):
+                 x_axis_type='linear', y_axis_type='linear',
+                 cache_ranges=True):
         self.plot_width = plot_width
         self.plot_height = plot_height
-        self.x_range = tuple(x_range) if x_range is not None else x_range
-        self.y_range = tuple(y_range) if y_range is not None else y_range
+        self.x_range = tuple(x_range) if x_range is not None else None
+        self.y_range = tuple(y_range) if y_range is not None else None
```
jbednar: I guess you found this way of writing it clearer (as it would seem to have no effect)?  I personally would maybe prefer:

```
self.x_range = None if x_range is None else tuple(x_range)
self.y_range = None if y_range is None else tuple(y_range)
```

datashader/glyphs.py
```diff
@@ -29,10 +29,12 @@ def validate(self, in_dshape):
             raise ValueError('y must be real')
 
     def _compute_x_bounds(self, df):
-        return df[self.x].min(), df[self.x].max()
+        xs = df[self.x].values
+        return xs.min(), xs.max()
```
jbednar: Should this and `_compute_y_bounds` be @jit so that Numba can combine the max and min into a single pass?

gbrener: Good point, I'll investigate.

datashader/core.py
gbrener: Ok, not a problem - happy to make the change if it's easier to read.

datashader/glyphs.py
gbrener: @jbednar I had some difficulties using numba with dask. For this reason there are now two versions of the bounds computations: one is for pandas dataframes, relying on numba and only doing one pass through the data. The other is a memoized version of the existing code (except with `np.nanmin` and `np.nanmax`). The memoization gets us the same performance as cachey did - which is great news - but still does two passes through the data. The memoization does not work with pandas dataframes because they're mutable, whereas dask dataframes are not. However there is still a solid speedup to the pandas side of datashader.

gbrener:
@jbednar Updated results posted here: https://github.com/bokeh/datashader/issues/129

datashader/glyphs.py
```diff
@@ -28,11 +29,45 @@ def validate(self, in_dshape):
         elif not isreal(in_dshape.measure[self.y]):
             raise ValueError('y must be real')
 
-    def _compute_x_bounds(self, df):
-        return df[self.x].min(), df[self.x].max()
+    @staticmethod
+    @ngjit
+    def _compute_x_bounds(xs):
+        minval = maxval = xs[0]
```
martindurant: This is good in many cases, but probably falls over if the first item happens to be nan.
Another little loop to skip over any initial nan values would do here, and I suppose there should be a way to raise an error if it happens all values are nan.

gbrener: Good catch @martindurant. Those cases are covered now. Also added the check for the empty array condition.

gbrener: Also, I'm on vacation this week - feel free to make additional changes to this PR as you see fit.

datashader/glyphs.py
```diff
@@ -32,25 +31,49 @@ def validate(self, in_dshape):
     @staticmethod
     @ngjit
     def _compute_x_bounds(xs):
+        if len(xs) == 0:
+            raise ValueError('x coordinate array is empty.')
```
jbednar: Computing len() for a Dask dataframe is non-trivial, since it is loaded lazily, but I think this entire branch can be avoided by simply setting minval=maxval=np.NaN.

martindurant: In that case x < minval and x < maxval will both always be False, so you would only ever get NaN out.

gbrener:
@jbednar How's this?

datashader/glyphs.py
```diff
@@ -31,47 +31,27 @@ def validate(self, in_dshape):
...
-                minval = maxval = x
-            idx += 1
-        while idx < len(xs):
-            x = xs[idx]
-            if not np.isnan(x):
-                if x < minval:
+                if np.isnan(minval) or x < minval:
                     minval = x
-                elif x > maxval:
+                if np.isnan(maxval) or x > maxval:
```
martindurant: Yes, I think this does the right thing at the slight cost of checking nan every loop.

gbrener: Indeed. But at least it can be reused for dask at some point in the future.

datashader/glyphs.py
jbednar: True; that would need an extra isnan(minval), etc. inside the loop. @jcrist points out that at this point the chunk of data has already been loaded, and so the len() should be cheap, so I'll leave it as-is.

datashader/glyphs.py
jbednar: Ok, I'll go ahead and merge this, and we can separately consider further optimizations in the future.  Thanks, all!

Issue #325, opened 2017-04-26
Punch list for 0.5.0


jbednar:
Various tasks we hope to do before releasing 0.5.0:

* [x] Verify line example after off-by-one changes [JC]
* [x] Fix dashboard to use census.parq [JC]
* [x] Make a new census.parq: [JC]
   - [x] Use snappy compression
   - [x] Use single, not double, precision
   - [x] Change to use northing,easting, not metersnorth,meterswest
   - [x] Upload to S3 bucket
* [x] Fixes/improvements for `census.*` notebooks: [JC]
   - [x] Update to use new census.parq, with recommended options as in #313 (fixed in #348)
* [ ] Update all examples to use holoviews where appropriate (JB; can be done after release)
* [x] Test and fix all notebook examples (legend, hover likely have issues) (JC,JB)
* [x] Add conda package for newer pip-only python-snappy version (JC) (conda-forge pkgs already exist)
* [x] Update conda packages to get higher performance: dask>0.14.1, fastparquet>0.0.5 (waiting on next release), numba (after new PR) (JC to verify that these get installed; may need to specify numba channel) (fixed in #349)
* [x] Update docs to include performance recommendations from #313 and #129.

PR #350, opened 2017-05-10
Miscellaneous fixes for notebook examples

jbcrail:
Related to #325

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -50,8 +50,9 @@ jupyter notebook --NotebookApp.iopub_data_rate_limit=100000000
 ```
 (The data_rate setting here is required with Jupyter 5.0, but can be omitted for earlier or later versions).
 
-While running the notebooks, if you see `Loading BokehJS ...` but never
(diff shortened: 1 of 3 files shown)


examples/README.md
```diff
@@ -46,7 +46,7 @@ Cloud](https://anaconda.org/jbednar/notebooks). To run these notebooks on your
 own system, start up a Jupyter notebook server:
 
 ```
-jupyter notebook --NotebookApp.iopub_data_rate_limit=100000000
+BOKEH_RESOURCES=inline jupyter notebook --NotebookApp.iopub_data_rate_limit=100000000
```
jbednar: Why do we want to suggest ignoring the CDN?

jbcrail: I ran into this issue: https://github.com/bokeh/bokeh/issues/5351. It seems helpful to default to this to prevent future issues.

Perhaps we should just add an extra note mentioning how to resolve it if it occurs, while continuing to use the CDN by default.

PR #351, opened 2017-05-11
Add performance documentation

jbcrail:
This is based on @gbrener's writeup on #129 and #313.

Related to #325


jbednar:
That's helpful, thanks!

PR #352, opened 2017-05-11
Fix hover rendering for categoricals

jbcrail:
The hover tooltips failed to render in hover_layer.ipynb. It appears that the values column is required to be in hover_data in order to render the tooltip for categoricals.

Related to #325


jbednar:
Looks good, thanks!

PR #360, opened 2017-05-19
Update conda pkgs in filetimes.yml

gbrener:
Update the package versions in examples/filetimes.yml and remove the pip workarounds from examples/filetimes.sh (since they're no longer needed).


jbednar:
Can you roll back "add the 1-billion pt OSM dataset to examples/datasets.yml"?  I've left that off deliberately so that only people who explicitly request this dataset get it, to reduce our download bandwidth. I guess I can be argued out of this position...

gbrener:
@jbednar - done.

jbednar:
Looks good, thanks!

PR #367, opened 2017-06-01
Explicitly transform span argument for non-linear normalizations

jbednar:
Right now, the `span` argument for `tf.shade()` is only useful for `linear` transformations, because the specified span values are applied *after* transformation, by which point the data values can be in some arbitrary range (for `eq_hist`) or in some non-arbitrary but awkward to specify range (for `log`, `cbrt`, etc.).  Now transforms the provided `span`, where appropriate, and aborts where such transformation is not yet supported.


jlstevens:
Looks fine to me!

philippjfr:
Looks good to me too, handling the ``log`` case correctly definitely seems worthwhile. Need to make sure this is mentioned in the release notes in case someone was using it (e.g. by manually taking the log). Merging.

PR #370, opened 2017-06-07
Parameterized the edge bundling code

jbednar:
One of the great things about Datashader in general is that there are not normally any "magic" parameters, i.e., arbitrary values that need to be set just so an algorithm can run, with no clear constraints that determine those values.  For better or worse, the new edge bundling code contains multiple such arbitrary values, and it's now crucial to make sure that the values are declared clearly, documented, interpreted consistently, and easily modifiable by the user.  Moreover, in part because of these arbitrary choices, users are likely to need multiple such edge bundling algorithms, each with tradeoffs, and so it's important to be able to bundle up each algorithm into a clearly maintainable chunk.

Accordingly, I have packaged up the two edge bundling algorithms currently provided into classes that can be called just as the old functions were, except that some positional arguments have been made keyword-only arguments for consistency.

Making this change would not normally affect the actual results, but it does in this case, because two of the parameters were *not* previously being respected consistently: MIN_SEG and MAX_SEG.  These parameters were defined as module constants and then passed around divided by 2 in some cases, and defaulting to the module-level value in others.  They should now all be respected consistently, with declared values that can safely be overridden in subclasses if needed.

### DIFF
--- a/examples/edge_bundling.ipynb
+++ b/examples/edge_bundling.ipynb
@@ -30,7 +30,7 @@
    },
    "outputs": [],
    "source": [
-    "def render_points(df, width=4000, height=4000, cmap=[\"lightblue\", \"darkblue\"], bgcolor=None):\n",
+    "def render_points(df, width=4000, height=4000, cmap=[\"lightcoral\", \"darkred\"], bgcolor=None):\n",
     "    cvs = ds.Canvas(plot_width=width, plot_height=height, x_range=(0, 1), y_range=(0, 1))\n",
     "    agg = cvs.points(df, 'x', 'y',  ds.count())\n",
     "    img = tf.shade(agg, cmap=cmap)\n",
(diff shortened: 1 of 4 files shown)


jbednar:
Note that I didn't actually move much of the code into the new classes.  Presumably most of it could be made into class methods, but I'm not sure how all that works with Numba, so I left it alone for now.  Once we have multiple such classes sharing code we'll probably need to sort it out more cleanly.

examples/edge_bundling.ipynb
```diff
@@ -366,12 +369,10 @@
...
    "cell_type": "code",
    "execution_count": null,
-   "metadata": {
-    "collapsed": true
-   },
+   "metadata": {},
    "outputs": [],
    "source": [
-    "render_points(circular_nodes_df, width=4000, height=4000)"
+    "render_points(circular_nodes_df, width=2000, height=2000)"
```
jbcrail: With the reduced width and height, the plotted points are clipped. The original dimensions show the full plot.

jbednar:
The clipped plot is presumably because the range is 0,1 and the circular one is placed around zero; I didn't investigate further as I was already planning to update that for other reasons.

PR #371, opened 2017-06-07
Made alpha scaling respect min_alpha consistently

jbednar:
PR #345 added alpha-based scaling for non-categorical types, based on the support that was already provided for categorical types.  However, it mapped data values into the alpha range 0 to 255, with the result that low data values were completely invisible, which is particularly acute when there are only a few data values total. E.g. if plotting counts where 1000 pixels have count 1 and 2 pixels have count 2, the resulting plot has only two visible pixels, which is highly misleading.  Moreover, such situations are very common when zooming in; at some point just a couple of overlapping points are shown, and the entire dataset is invisible, with it reappearing when zooming in just slightly more to remove that overlap.

The categorical alpha-scaling code already provided a `min_alpha` parameter for this purpose, and this PR now makes the non-categorical alpha-scaling code respect it.


jbednar:
Tests will need to be updated to reflect this change.  @jbcrail, can you please push the required fixes to this branch, review, and merge if appropriate?

PR #373, opened 2017-06-09
Expanded edge_bundling notebook using HoloViews

jbednar:
Working with graph data requires both zooming and being able to overlay node and graph data easily, a combination that is not possible with Datashader's PIL support or with its InteractiveImage class.  These things are simple with HoloViews, though, and this new notebook replaces the previous one to tell a story about how to work with graph data.  The rendered notebook is at: https://anaconda.org/jbednar/edge_bundling 

image

This PR is marked WIP not because it's not complete, but to make it possible for anyone with suggestions to push to it before it is merged.  Improvements that could be made:

- [ ] The "force directed algorithm" for layout should be described more fully, with a link to the specific code used or at least a better description.
- [ ] The code for testing the various bw and decay values could probably be simplified, perhaps using NDLayout?  Not sure how to do that when it's not a full cross product of the parameter values (which seems too expensive).  Perhaps a GridMatrix would be a better approach, with smaller images (and maybe a smaller network) but a better exploration of the full space of parameter values.
- [ ] The hover information should show id,institution, rather than the current x,y,institution.  Not sure how to get access to the id for hovering, but x,y should be suppressable if using a custom HoverTool specification.
- [ ] It would be nice to be able to select a node and then show all nodes that have connections to it (@Philipp?).

### DIFF
--- a/examples/datasets.yml
+++ b/examples/datasets.yml
@@ -101,3 +101,7 @@ data:
       - calvert_uk_research2017_nodes.snappy.parq
       - calvert_uk_research2017_edges.snappy.parq
 
+  - url: http://s3.amazonaws.com/datashader-data/calvert_uk_research2017_nodes.zip
+    title: 'Institutions for Edge Bundling (Calvert UK Research 2017)'
+    files:
+      - calvert_uk_research2017_nodes.csv
--- a/examples/edge_bundling.ipynb
+++ b/examples/edge_bundling.ipynb
@@ -1,25 +1,159 @@
 {
  "cells": [
   {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Plotting graphs in Datashader with edge bundling\n",
+    "\n",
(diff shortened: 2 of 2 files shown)


philippjfr:
>The code for testing the various bw and decay values could probably be simplified, perhaps using NDLayout? Not sure how to do that when it's not a full cross product of the parameter values (which seems too expensive). Perhaps a GridMatrix would be a better approach, with smaller images (and maybe a smaller network) but a better exploration of the full space of parameter values.

I'll try that and see how expensive it is and whether the notebook will get too large.

>The hover information should show id,institution, rather than the current x,y,institution. Not sure how to get access to the id for hovering, but x,y should be suppressable if using a custom HoverTool specification.

This should do it:

```
%%opts Points [tools=[HoverTool(tooltips=[('Institution', '@{Institution}')])]
```

> It would be nice to be able to select a node and then show all nodes that have connections to it (@philipp?).

Would require a Selection stream and seems like a reasonable thing to do. ``decimate`` makes it a whole lot more difficult though.

jbednar:
>  decimate makes it a whole lot more difficult though.

Good point. I guess the way it would have to work is that once one has grabbed a certain node shown in the decimated plot, all the other nodes would disappear, to be replaced with a decimated set of the connected nodes.  I think it still has to be decimated even after selection, in case the network is fully connected and thus the connected node list is too large to display, but it needs to be decimated after selecting the connected nodes, not before.

IanCal:
For previews to try and choose the right kind of parameters, there's usually a choice made on a fairly high level visual feel to it (way too wirey, far too brushy). In which case it's likely that a subset of the edges would be fine to use. You can calculate what is (iirc) called a backbone of the network, largely reduces the edges but keeps the overall structure. Even if we want to render the whole graph finally, this subset could be very useful for previews. Might get around the issue of how to create the initial grid.

I'll have a scout around the literature, I think there's a reasonable way of calculating initial values which should work well across fairly diverse datasets.

PR #375, opened 2017-06-14
Replace rasterio with xarray, skimage with gridtools

gbrener:
Datashader no longer requires _rasterio_ as a hard dependency - in favor of the new [xarray rasterio backend](http://xarray.pydata.org/en/stable/generated/xarray.open_rasterio.html) - however _rasterio_ will be required for running some of the notebooks and executing the unit tests. Also, the `cvs.raster()` method no longer relies on `skimage.transform.resize` for resampling; we are now using [datashader/resampling.py]() which came out of gridtools.

A few functions were also added to [datashader/utils.py](datashader/utils.py), which we should consider offering to _xarray_ at some point:
* *calc_res*: Replaces _rasterio.Dataset.res_ attribute. This is not (yet?) being exposed by the _rasterio_ backend in _xarray_, so it needs to be re-calculated from the data
* *calc_bbox*: Replaces _rasterio.Dataset.bounds()_ method. Simply calculating max and min on xarr.x and xarr.y yields incorrect results - more details can be found [here](https://github.com/mapbox/rasterio/issues/556) - as _rasterio_ relies on _gdal_ (through the _affine_ library) to perform affine transformation on the points based on the source raster's crs. As a substitute for this functionality I implemented an affine transform based on the "Augmented Matrix" approach described [here](https://en.wikipedia.org/wiki/Affine_transformation#Augmented_matrix). It assumes the flat-earth model, and does not currently perform reprojection - although theoretically we could add reprojection capabilities to it in the future.

This PR is still a work in progress. I'd like to wrap up the following items before merging:
- [x] Test `cvs.rasterize()` on a file using the EPSG:32616 crs. It's been tested so far (and verified as working) with EPSG:3857, but there may be subtle issues e.g. with the affine transformation matrix
- [x] Fix a unit test (which broke after a recent change) that is testing when the Canvas object's x_range and/or y_range exceeds the plot_width/plot_height
- [x] Clean up the notebooks so they no longer depend on _rasterio_ directly. They can use _xarray_'s rasterio backend instead, which indirectly depends on _rasterio_
- [x] Replace our usage of _skimage.transform.resize_ with gridtools _resample2d_. ~~Although we initially wanted _gridtools_ to be part of the replacement for _rasterio_, it turns out that it would actually be replacing [skimage.transform.resize](http://scikit-image.org/docs/dev/api/skimage.transform.html#skimage.transform.resize). I made a rudimentary attempt to replace `resize()` with it, but it oesn't seem to like floats, and will require some effort to become conda-installable (it's not available on OSX). @philippjfr may be able to confirm these details. Do we want gridtools as a conda package?~~

### DIFF
--- a/datashader/core.py
+++ b/datashader/core.py
@@ -205,14 +205,14 @@ class Canvas(object):
                resample_method='bilinear',
                use_overviews=True):
         """Sample a raster dataset by canvas size and bounds. Note: requires
-        `rasterio` and `scikit-image`.  Missing values (those having the value
-        indicated by the "nodata" attribute of the raster) are replaced with
-        `NaN` if floats, and 0 if int.
+        `xarray`, `rasterio`, and `scikit-image`.  Missing values (those having
+        the value indicated by the "nodata" attribute of the raster) are
+        replaced with `NaN` if floats, and 0 if int.
 
         Parameters
         ----------
-        source : rasterio.Dataset
-            input datasource most likely obtain from `rasterio.open()`.
+        source : xarray.DataArray
+            input datasource most likely obtain from `xarray.open_rasterio()`.
         band : int
             source band number : optional default=1
         resample_method : str, optional default=bilinear
@@ -227,7 +227,7 @@ class Canvas(object):
 
         Notes
         -------
-        requires `rasterio` and `scikit-image`.
+        requires `xarray`, `rasterio`, and `scikit-image`.
         """
 
         try:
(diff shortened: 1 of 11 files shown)


philippjfr:
> I made a rudimentary attempt to replace resize() with it, but it doesn't seem to like floats

That's odd, I could only get it to work with floats.

> and will require some effort to become conda-installable (it's not available on OSX).

I think as long as licenses permit (I think they do), the idea was we'd just include the ``resample`` module in datashader so we don't have another dependency and have to make a conda package.

gbrener:
Ok, it could have been user error on my part. I'll try again, and include the file on its own into this repo.

jbednar:
Right; we'll probably need to extend things like datatype handling, and we already depend directly on numba, so it seems simplest just to bring that file directly into datashader.

gbrener:
@philippjfr You're right - I had a moment of dyslexia there. The gridtools resample function only works on floats, which required casting the array to a float dtype. We might want to revisit that later on, but for now based on the notebook visualizations it seems to be working.

gbrener:
This PR looks ready to be merged; the build is failing because the xarray conda-package doesn't support Python 3.4. This is strange because the main xarray repo claims to support it. Here's the conda-forge commit that removed 3.4 support: https://github.com/conda-forge/xarray-feedstock/commit/7dfd7a70e5ce9f17b3964a9b2752e321031d9cc6

datashader/core.py
```diff
@@ -227,65 +228,39 @@ def raster(self,
...
-
-        is_int = np.issubdtype(data.dtype, np.integer)
-        data[data == np.array(source.nodata)] = 0 if is_int else np.nan
-
-        # TODO: this resize should go away once rasterio has overview resample
-        data = resize(data,
-                      (h, w),
-                      order=resample_methods[resample_method],
-                      preserve_range=True)
+        data = resample_2d(source.values[0].astype(np.float32), w, h)
```
philippjfr: Should we pass the ``resample_method`` through here to the ``us_method`` argument here? It also supports 'bilinear' and 'nearest'. Secondly gridtools allows controlling the aggregation method, which is 'mean' by default, we could expose the ``ds_method`` and provide options for 'mean', 'first', 'var', 'std' and 'last'. @jbednar What do you think?

jbednar: Sounds like important extensions, to me.

philippjfr:
Apart from my comment above this looks great to me. My only other suggestion is clearing the notebooks and pushing again.

datashader/core.py
gbrener: It's a good idea, I just didn't want to make the API more complicated (removing the `resample_method` kwarg in favor of `us_method` and `ds_method`).

jbednar: Both ds and us methods seem important, and better to expose them now than have to change later.

gbrener:
@jbednar @philippjfr the downsample and upsample methods are now exposed via the `cvs.raster()` method signature, and are covered by the unit tests. I cleared output from the notebook as well, and pinned the xarray version explicitly in the meta.yaml.

gbrener:
Quick note: I created an issue on xarray which requests additional attributes from their rasterio backend (https://github.com/pydata/xarray/issues/1456) and opened a PR to address it (https://github.com/pydata/xarray/pull/1455). If these or something similar gets into the next release, then we'll be able to reuse the transform matrix and resolutions from the geotiff files instead of having to recalculate the resolution and hard-code the matrix which assumes flat-earth.

philippjfr:
Great, thanks for making the suggested changes. All looks good to me now, I'll let @jbednar do a final review and merge it.

datashader/core.py
```diff
@@ -5,7 +5,10 @@
 from odo import discover
 from xarray import DataArray
 
-from .utils import Dispatcher, ngjit
+from .utils import Dispatcher, ngjit, calc_res, calc_bbox
+from .resampling import (resample_2d, US_NEAREST, US_LINEAR, DS_FIRST, DS_LAST,
+                         DS_MEAN, DS_MODE, DS_VAR, DS_STD)
+
```
jbednar: I'd rather not import all those here, with names like that; it's not pythonic.  Instead, can you have `resampling` provide the `upsample_methods` and `downsample_methods` dictionaries?  In fact, we should just go ahead and make these classes Parameterized, as at the end of bundling.py, so that these values can be expressed directly; no need for the special logic that is in this file to handle how to drive some other file.

PR #376, opened 2017-06-15
Add edge bundling tests

jbcrail:


jbednar:
Oops; there are test failures.  Seems like some type coercion needs to be added for some Python versions to be happy?

PR #387, opened 2017-06-22
Minor readme modifications

ceball:
Note: you might not like the second commit in this PR. Happy to remove it if so. Or - because someone might try to install via pip even if datashader is not on pypi - I could reword it to say explicitly that we recommend not installing via pip (...except for editing the source, once you already used conda to install the dependencies).

Sorry, so many words for such an insignificant change!


jbednar:
Thanks!

PR #389, opened 2017-06-22
Various fixes for canvas.raster regridding

philippjfr:
Fixe for issues mentioned in https://github.com/bokeh/datashader/issues/380

Handles several issues:

* Handles ascending/descending x-/y-coordinates correctly (assuming they are always aligned at the center)
* Handles both 2D and 3D arrays assuming that the last two DataArray.dims are the y- and x-axis respectively
* Handles non rasterio input without ``xarr._file_data.nodata`` metadata.
* Adds coordinates to returned DataArray
* Keeps coordinate names

Further issues:

*  ~~Still always assumes bin edges as both input and output~~
* ~~Does not address issue slicing the input image according to the canvas x_range and y_range~~

You can find a notebook with further details here: https://anaconda.org/philippjfr/regridding/notebook

### DIFF
--- a/datashader/core.py
+++ b/datashader/core.py
@@ -209,6 +209,10 @@ class Canvas(object):
                downsample_method='mean'):
         """Sample a raster dataset by canvas size and bounds.
 
+        Handles 2D or 3D xarray DataArrays, assuming that the last two
+        array dimensions are the y- and x-axis that are to be
+        resampled.
+
         Missing values (those having the value indicated by the
         "nodata" attribute of the raster) are replaced with `NaN` if
         floats, and 0 if int.
@@ -247,7 +251,7 @@ class Canvas(object):
             raise ValueError('Invalid downsample method: options include {}'.format(list(downsample_methods.keys())))
 
         res = calc_res(source)
-        ydim, xdim = source.dims if source.ndim == 2 else source.dims[1:]
+        ydim, xdim = source.dims[-2:]
         left, bottom, right, top = calc_bbox(source[xdim].values, source[ydim].values, res)
         array = orient_array(source, res, band)
         dtype = array.dtype
--- a/datashader/core.py
+++ b/datashader/core.py
@@ -5,7 +5,7 @@ from datashape.predicates import istabular
 from odo import discover
 from xarray import DataArray
 
-from .utils import Dispatcher, ngjit, calc_res, calc_bbox, get_indices
+from .utils import Dispatcher, ngjit, calc_res, calc_bbox, orient_array, compute_coords, get_indices
 from .resampling import (resample_2d, US_NEAREST, US_LINEAR, DS_FIRST, DS_LAST,
                          DS_MEAN, DS_MODE, DS_VAR, DS_STD)
 
@@ -247,7 +247,13 @@ class Canvas(object):
             raise ValueError('Invalid downsample method: options include {}'.format(list(downsample_methods.keys())))
 
         res = calc_res(source)
-        left, bottom, right, top = calc_bbox(source.x.values, source.y.values, res)
+        if source.ndim == 2:
+            ydim, xdim = source.dims
+        else:
+            ydim, xdim = source.dims[1:]
+        left, bottom, right, top = calc_bbox(source[xdim].values, source[ydim].values, res)
+        array = orient_array(source, res, band)
+        dtype = array.dtype
 
         # window coordinates
         xmin = max(self.x_range[0], left)
@@ -264,12 +270,18 @@ class Canvas(object):
         w = int(np.ceil(self.plot_width * width_ratio))
         h = int(np.ceil(self.plot_height * height_ratio))
 
-        cmin, rmin = get_indices(xmin, ymin, source.x.values, source.y.values, res)
-        cmax, rmax = get_indices(xmax, ymax, source.x.values, source.y.values, res)
-        source_window = source[:, rmin:rmax, cmin:cmax]
+        cmin, rmin = get_indices(xmin, ymin, source[xdim].values, source[ydim].values, res)
+        cmax, rmax = get_indices(xmax, ymax, source[xdim].values, source[ydim].values, res)
+        if rmin > rmax:
+            rmin, rmax = rmax, rmin
+        if cmin > cmax:
+            cmin, cmax = cmax, cmin
+        if array.ndim == 2:
+            source_window = array[rmin:rmax, cmin:cmax]
+        else:
+            source_window = array[:, rmin:rmax, cmin:cmax]
 
-        data = resample_2d(source_window.values[0].astype(np.float32),
-                           w, h,
+        data = resample_2d(source_window.astype(np.float32), w, h,
                            ds_method=downsample_methods[downsample_method],
                            us_method=upsample_methods[upsample_method])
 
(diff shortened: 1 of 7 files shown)


philippjfr:
I think the way I've handled ascending and descending coordinates may be incorrect now, the current implementation makes the following assumptions:

* If x-coords are ascending it assumes coords describe left edges
* If x-coords are descending it assumes coords describe right edge
* If y-coords are ascending it assumes coords describe bottom edge
* If y-coords are descending it assumes coords describe top edge

This probably doesn't make much sense, but if we decide to switch to bin centers this needs fixing anyway.

jbednar:
Right.  Let's just put this PR on hold until using bin centers is in place, then fix this up to match.

philippjfr:
Just rebased this against #392, other than the inverted row/column indices in that PR this now works correctly and as far as I can tell this PR now has everything I need from regridding. I'm no longer sure whether we should handle center-based indexing right away particularly because most datasets I've run across are **not** bin-center indexed. I'd be happy to clean this PR up a bit, adding unit tests and better docstrings but I'll let @gbrener and @jbednar weigh in first.

philippjfr:
Actually given that we cannot rely on any one fixed convention I'd suggest we add an argument to the ``canvas.raster`` method to indicate whether the data is indexed by the bin edges or centers.

jbednar:
I think we should consistently use bin-center indexing for our own purposes, whenever we use floating-point indexes, but that we should allow the convention to be declared for the source data, and we should have the default for that assumed convention match whatever the most common value is in practice.  That way most people will get the right plots, and most people who don't care about the 1/2-pixel errors will never worry about it, leaving only a few people ever having to adjust the declared convention.

philippjfr:
>I think we should consistently use bin-center indexing for our own purposes, whenever we use floating-point indexes, but that we should allow the convention to be declared for the source data, and we should have the default for that assumed convention match whatever the most common value is in practice. 

Taking bin edges as input and spitting out bin centers seems like a recipe for confusion. I think if we allow declaring the input convention it would be easiest if the output would just follow that.

jbednar:
Philipp and I discussed further and I think we agreed that Datashader needs to define a consistent output format respected across all of its data types (points, lines, rasters), which I think entails using center-based bins.  So I've asked @gbrener to submit a separate PR switching to center-based bins for rasters, which we can merge before updating and merging this PR.

philippjfr:
Okay, I'm happy to update this PR once that's happened.

philippjfr:
@gbrener @jbednar I've now updated the code assuming bin centers as input and adding centered coordinates on the output. Greg mentioned something about the nodata attribute, which I don't recall now, that could presumably be resolved here as well. Hopefully I'll have time to add some unit tests on the flight.

philippjfr:
Spoke too soon, looks like I broke a bunch of tests.

philippjfr:
Suppose we have to make sure to test with latest xarray which includes Greg's fix for rasterio loading.

gbrener:
@philippjfr I think it might be better to wait on the next xarray release before updating/changing datashader to rely on its new functionality, since the xarray team might still change or remove things before the release.

philippjfr:
Good point, you think there's anything we can do to push them to do a release soon? Some of our projects will need it fairly soon.

gbrener:
I'm unfamiliar with their release procedure. Perhaps we can add a feature toggle, which will make it easier to migrate once they do the release?

philippjfr:
@gbrener As far as I can tell your xarray conda package contains an old version (https://github.com/pydata/xarray/commit/22ff955d53e253071f6e4fa849e5291d0005282a) without your rasterio fix. So the tests on this PR will continue failing.

philippjfr:
@jbednar @gbrener Ready to review. I'll restart the tests once Greg's new xarray build is up.

datashader/utils.py
```diff
@@ -68,9 +68,18 @@ def calc_res(raster):
     """Calculate the resolution of xarray.DataArray raster and return it as the
     two-tuple (xres, yres).
     """
-    w, h = raster.shape[2], raster.shape[1]
-    xres = (raster.x.values[w-1] - raster.x.values[0]) / (w - 1)
-    yres = (raster.y.values[0] - raster.y.values[h-1]) / (h - 1)
+    if raster.ndim == 2:
+        h, w = raster.shape
+        ydim, xdim = raster.dims
+    elif raster.ndim == 3:
```
gbrener: I wonder if we can make this more generalizable? What is preventing this from working on >3 dimensions, for example?

datashader/utils.py
```diff
@@ -95,22 +104,22 @@ def calc_bbox(xs, ys, res):
...
-        if y < ymin:
-            ymin = y
-        if y > ymax:
-            ymax = y
-    return xmin, ymin, xmax, ymax
+        if x < xmin: xmin = x
+        if x > xmax: xmax = x
+        if y < ymin: ymin = y
+        if y > ymax: ymax = y
+    xpad, ypad = res[0]/2., res[1]/2.
```
gbrener: Is the half-grid adjustment still necessary, since we are now relying on xarray to provide bin-centered coordinates?

datashader/utils.py
```diff
@@ -154,6 +166,73 @@ def get_indices(x, y, xs, ys, res):
...
+        if res[0] < 0: array = array[:, ::-1]
+        if res[1] > 0: array = array[::-1]
+    else:
+        if band is not None: array = array[band-1]
+        if res[0] < 0: array = array[:, :, ::-1]
+        if res[1] > 0: array = array[:, ::-1]
+    return array
+
+
+def compute_coords(width, height, x_range, y_range, res):
```
gbrener: Is `compute_coords` still necessary, since we are now relying on xarray to provide bin-centered coordinates?

datashader/utils.py
philippjfr: Yes, because this is basically undoing the bin centering returning the left/right and top/bottom edge to return the bounding box.

datashader/utils.py
philippjfr: That's a good question, I'll have another look at the code to see if there's an easy way to generalize this.

datashader/utils.py
philippjfr: Yes, this computes the coordinates of the returned array, which are independent of the input and are computed from the bounds and then oriented to match the input.

datashader/utils.py
philippjfr: I've decided that it's not too urgent for now and have simply documented the current behavior and simplified the implementation a bit.

philippjfr:
@jbednar Ready to merge once tests pass and you've reviewed it.

datashader/core.py
```diff
@@ -282,25 +301,48 @@ def raster(self,
             lpct = lpad / (lpad + rpad) if lpad + rpad > 0 else 0
             left = int(np.ceil(num_width * lpct))
             right = num_width - left
-            left_pad = np.empty(shape=(self.plot_height, left)).astype(source_window.dtype) * np.nan
-            right_pad = np.empty(shape=(self.plot_height, right)).astype(source_window.dtype) * np.nan
+            lshape, rshape = (self.plot_height, left), (self.plot_height, right)
+            if bands > 1:
```
jbednar: I think we agreed to call these "layers" for generality, as "bands" mainly makes sense for satellite data, and channels was too confusing.

jbednar:
Looks good apart from minor naming issue and conflicts to resolve.

philippjfr:
@jbednar Updated the terminology and rebased the PR.

Issue #380, opened 2017-06-17
Various issues with new Canvas.raster


philippjfr:
Currently the newly ``Canvas.raster`` method for regridding makes several assumption that keep it from being general. Here are the issues I encountered working with it:

* It assumes a 3D array shape even though the regridding happens in 2D. Ideally it would check whether the data is a 2D or 3D ``xarray.DataArray`` and work correctly on both. The ``datashader.util.calc_res`` function needs to be updated.
* The ``calc_res`` utility also currently assumes that the coordinate arrays are named ``x`` and ``y``, which is not a safe assumption to make in general.
* The ``attrs`` inherited from the input array should be optional i.e. this line in Canvas.raster should be conditional in some way: ``attrs = dict(res=res[0], nodata=source._file_obj.nodata)``
* If the ycoords are ascending rather than descending the ``yres`` ends up negative, resulting in a wrong bounding box. Need to consider flipped coordinates along both axes.
* When the canvas x_range and y_range are smaller than the canvas range, the data is not sliced correctly, i.e. even though some of the data is situated outside of the canvas all the data is still included.

To visualize the last issue here's what happens when you try to regrid an image with a bounding box of (-0.5, -0.5, 0.5, 0.5) onto a canvas with a bounding box of (-0.2, -0.2, 0.2, 0.2) this is the result:

<img width="616" alt="screen shot 2017-06-17 at 7 36 18 pm" src="https://user-images.githubusercontent.com/1550771/27255336-45ccd2d6-5394-11e7-9a39-58099a3f9c6d.png">

i.e. the whole image has been squeezed into the new canvas rather than appropriately excluding samples that fall outside the new bounding box. @gbrener @jbednar

jbednar:
Looks like merging that PR was premature.  @gbrener, will you be able to address these issues on Monday?  If not, I guess we should back out those changes for now.

gbrener:
@philippjfr @jbednar Did the previous implementation of `cvs.raster()` meet these requirements? Either way it would be very helpful if we had unit tests for these cases (decorated with `@pytest.mark.skip` even though they'd be failing). I'd be happy to add that to my TODO list if @philippjfr has too much on his plate.

philippjfr:
>Did the previous implementation of cvs.raster() meet these requirements? 

[...]

PR #392, opened 2017-06-26
Add slicing based on canvas bounds

gbrener:
Add slicing for the source data array based on user-defined canvas boundaries. This should address the final bullet point in Issue #380 .

### DIFF
--- a/datashader/core.py
+++ b/datashader/core.py
@@ -5,7 +5,7 @@ from datashape.predicates import istabular
 from odo import discover
 from xarray import DataArray
 
-from .utils import Dispatcher, ngjit, calc_res, calc_bbox
+from .utils import Dispatcher, ngjit, calc_res, calc_bbox, get_indices
 from .resampling import (resample_2d, US_NEAREST, US_LINEAR, DS_FIRST, DS_LAST,
                          DS_MEAN, DS_MODE, DS_VAR, DS_STD)
 
@@ -249,7 +249,7 @@ class Canvas(object):
         res = calc_res(source)
         left, bottom, right, top = calc_bbox(source.x.values, source.y.values, res)
 
-        # window coodinates
+        # window coordinates
         xmin = max(self.x_range[0], left)
         ymin = max(self.y_range[0], bottom)
         xmax = min(self.x_range[1], right)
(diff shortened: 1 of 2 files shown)


datashader/core.py
```diff
@@ -264,7 +264,11 @@ def raster(self,
         w = int(np.ceil(self.plot_width * width_ratio))
         h = int(np.ceil(self.plot_height * height_ratio))
 
-        data = resample_2d(source.values[0].astype(np.float32),
+        rmin, cmin = get_indices(xmin, ymin, source.x.values, source.y.values, res)
+        rmax, cmax = get_indices(xmax, ymax, source.x.values, source.y.values, res)
```
philippjfr: Are these inverted by any chance? ``get_indices`` seems to be returning x and y, which should map onto column and row indices respectively. This fixes it for me.

jbednar: Can you clarify what you mean by "this" fixing it?  Do you need to push a change to this PR before it can be merged, or do you mean the PR is already correct?

philippjfr: By "this" I mean inverting rmin/cmin and rmax/cmax. Happy to push the fix to this PR, but would be good if @gbrener could confirm.

jbednar: Please do push the fix; @gbrener is away today, and he can make a new PR tomorrow if this wasn't the correct approach.  Better for us to be able to move on and merge this and your other PR.

gbrener: Hey @philippjfr , I think you're correct - about to push up a fix.

philippjfr:
Perfect thanks. I've tested it and it worked well, looks ready to merge.

datashader/utils.py
```diff
@@ -73,11 +73,10 @@ def calc_res(raster):
     yres = (raster.y.values[0] - raster.y.values[h-1]) / (h - 1)
     return xres, yres
 
-
 def calc_bbox(xs, ys, res):
     """Calculate the bounding-box of a raster, and return it in a four-element
     tuple: (xmin, ymin, xmax, ymax). This assumes a flat-earth crs to do an
-    affine transformation with the "Augmented Matrix" approach (does not handle
+    Affine transformation with the "Augmented Matrix" approach (does not handle
```
jbednar: "Affine" should not be capitalized; I'll edit that and push it.

jbednar:
I updated the docstrings to make it clear that we are not assuming a raster contains geographic data.  I mention the flat-earth assumption as an example, but in general a raster could be anything, and even discussing "flat earth" in that context is not meaningful.  Happy to merge once tests pass.

PR #398, opened 2017-07-05
Add initial graph layout algorithm

jbcrail:


datashader/layout.py
```diff
@@ -0,0 +1,145 @@
...
+        nodes, edges : pandas.DataFrame
+        """
+        try:
+            import networkx as nx
+        except ImportError:
+            raise ImportError('install networkx to use this feature')
+
+        # This comes from the sparse FR layout in NetworkX
+        A = nx.to_scipy_sparse_matrix(graph, dtype='f')
+        nnodes, _ = A.shape
```
jbednar: Is this just a temporary implementation for testing?  We can't depend on networkx in general.

datashader/layout.py
```diff
@@ -0,0 +1,145 @@
...
+        nohubs : bool
+            Whether to grant authorities (nodes with a high indegree) a
+            more central position than hubs (nodes with a high outdegree)
+        k : float
+            Compensates for the repulsion for nodes that are far away
+            from the center. Defaults to the inverse of the number of
+            nodes.
+        dim : int
+            Coordinate dimensions of each node.
+
```
jbednar: If this is a ParameterizedFunction, it should have parameters, i.e. declare ``iterations = param.Integer(10,doc="Number of iterations")`` at the class level, with ``__call__`` set up as other ParameterizedFunctions are (no explicit named arguments like that) and no docstring like this.

datashader/layout.py
```diff
@@ -0,0 +1,145 @@
...
+        else:
+            pos = pos.astype(A.dtype)
+        if k is None:
+            k = np.sqrt(1.0 / nnodes)
+
+        # the initial "temperature" is about .1 of domain area (=1x1)
+        # this is the largest step allowed in the dynamics.
+        t = 0.1
+
+        # simple cooling scheme.
```
jbednar: Should the meat of this function be in a separate Numba-ized function?

datashader/layout.py
jbcrail: This was meant as temporary until I could switch over to using dataframes for input. The original implementation relied on NetworkX.

datashader/layout.py
jbcrail: Thanks. I updated the class.

jbcrail:
@jbednar Thanks for the feedback. The initial algorithm is now fully parameterized, uses only dataframes for both input and output, and uses Numba as appropriate.

I'm going to merge this so that I can work on a simple demo and then create a new dev release.

jbednar:
Looks good, thanks!

PR #401, opened 2017-07-07
Generalized string testing to work with unicode on python2

jbednar:
As suggested by @philippjfr to allow unicode to be passed for colors in Python2.


philippjfr:
Looks good, thanks for the fix.

PR #408, opened 2017-07-14
Add random and circular graph layout

jbcrail:


datashader/layout.py
```diff
@@ -125,3 +125,67 @@ def __call__(self, nodes, edges, **params):
...
+
+
+@nb.jit(nogil=True)
+def _create_random_xy_values(df, seed):
+    np.random.seed(seed)
+
+    points = np.asarray(np.random.random((len(df), 2)))
+    df['x'] = points[:, 0]
+    df['y'] = points[:, 1]
+    return df
```
jbednar: Why is this function using Numba?  It doesn't seem like there is anything for numba to do here.  If that's true, I'd move this code into ``random_layout.__call__`` directly.

datashader/layout.py
```diff
@@ -125,3 +125,67 @@ def __call__(self, nodes, edges, **params):
...
+
+    if uniform:
+        thetas = np.arange(circumference, step=circumference/len(df))
+    else:
+        thetas = np.asarray(np.random.random((len(df),))) * circumference
+
+    df['x'] = x0 + r * np.cos(thetas)
+    df['y'] = y0 + r * np.sin(thetas)
+
+    return df
```
jbednar: Same here -- if there's no loop, what would Numba have to optimize? Seems like it could be moved into ``circular_layout.__call__``.

datashader/layout.py
jbcrail: You're right that Numba served no function here. I merged the code into the constructor.

datashader/layout.py
jbcrail: Same as above.

jbednar:
Looks great, thanks!

PR #418, opened 2017-07-26
Fix conversion from graph to sparse matrix

### LINKED ISSUE #417: High memory usage with force-directed layout and large integer node IDs
When swapping `circular_layout` with `forceatlas2_layout` in `examples/packet_capture_graph.ipynb`, the Jupyter kernel dies unexpectedly due to high memory usage. The notebook consistently dies in `_convert_edges_to_sparse_matrix`.

jbcrail:
When converting to a sparse matrix, if the nodes dataframe has IDs with high integer values, the generated sparse matrix would take on a shape with a range of (0, max), where max is the highest node ID. This would cause high memory usage even if only a relatively few nodes existed.

Fixes #417

### DIFF
--- a/datashader/layout.py
+++ b/datashader/layout.py
@@ -98,14 +98,20 @@ def _extract_points_from_nodes(nodes):
     return points
 
 
-def _convert_edges_to_sparse_matrix(edges):
-    if 'weight' in edges:
-        weights = edges['weight']
-    else:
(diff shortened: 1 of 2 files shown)


jbednar:
Can you check out what's making the tests fail?

jbednar:
Ready to merge?

jbcrail:
Yes

PR #422, opened 2017-07-31
Ensure canvas.points and canvas.line return bin center coordinates

philippjfr:
Previously the xarray ``DataArray``s returned by the canvas.points and canvas.line have specified the coordinates of the data as bin edges. However xarray in general assumes bin centers and we have all agreed in various discussions that bin centers are more consistent. While this introduces yet another change in the coordinate semantics in datashader it is the correct choice going forward and will reduce headaches for everyone involved. The change itself is trivial but all the tests had to be updated accordingly.

### DIFF
--- a/datashader/core.py
+++ b/datashader/core.py
@@ -82,7 +82,7 @@ class Axis(object):
         -------
         index : ndarray
         """
-        px = np.arange(n)
+        px = np.arange(n)+0.5
         s, t = st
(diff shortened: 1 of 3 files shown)


jbednar:
Looks good, thanks.

philippjfr:
@jbednar Ready to merge.

Issue #424, opened 2017-08-01
Remove datashader's dependency on odo


gbrener:
Datashader currently faces a couple of issues introduced by `odo`:
- Deprecation warnings identified by issue https://github.com/blaze/odo/issues/543
- Performance degradation identified by issue #396 

I attempted to fix the performance issue in `odo`, after tracing it to a line which converts all the categorical codes into a list to be passed to `datashape`'s `Categorical` constructor, however I'm not aware of a straightforward fix that achieves similar performance to the non-categorical case. Furthermore, a new release of `odo` will be required to realize any of the improvements made to it, and there is no indication of one arriving anytime soon. Given the situation it seems easier to remove `odo` from datashader altogether, which should achieve a performance boost for the #396 use-case, remove the warnings from https://github.com/blaze/odo/issues/543, and simplify the codebase. So I assign myself to this issue, and hope to open a PR within the next day or so.

gbrener:
This is done, as of PR #428.

Issue #396, opened 2017-07-01
Pandas .astype('category') dramatically decreases performance


alexanderkuk:
I have a DataFrame with 2 000 000 rows. It has columns client, driver, pickup_latitude, pickup_longitude. I do
```
table['client'] = table.client.astype('category')
table['driver'] = table.driver.astype('category')
```
to save some memory space. Aggregation
```
aggregate = canvas.points(table, 'pickup_longitude', 'pickup_latitude', ds.count())
```
runs for 4 seconds. 

If I do not cast client and driver to category, aggregation runs in 20ms. 

It is not obvious for me why .astype('category') decreases performance so much. It was hard to debug. Maybe add some notice or warning?

jbednar:
Is this a Pandas or a Dask dataframe?  Normally using categorical datatypes does not reduce performance, so this is definitely surprising, all the more so because the aggregation here is not using either of those columns. I'd need a reproducible example to try to figure it out.

alexanderkuk:
[data.csv.zip](https://github.com/bokeh/datashader/files/1117409/data.csv.zip)
```
table = pd.read_csv('data.csv.zip')
canvas = ds.Canvas(plot_width=700, plot_height=662, x_range=(4146921.8723705406, 4224887.706012843), y_range=(7470522.750106069, 7544365.4667577855))

%%timeit
aggregate = canvas.points(table, 'longitude', 'latitude', ds.count())
3.25 ms ± 350 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

table = pd.read_csv('data.csv.zip')
table['client'] = table.client.astype('category')
table['driver'] = table.driver.astype('category')

canvas = ds.Canvas(plot_width=700, plot_height=662, x_range=(4146921.8723705406, 4224887.706012843), y_range=(7470522.750106069, 7544365.4667577855))

%%timeit
aggregate = canvas.points(table, 'longitude', 'latitude', ds.count())
572 ms ± 24.7 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

jbednar:
Thanks!  I've asked Greg to look into that when he gets the time.

gbrener:
I've taken a quick look at this, and it appears that in the categoricals case most of the time is spent doing schema validation with `odo`. Without categoricals, most of the time is spent in `Canvas.points()` (as expected). I suspect that there is a performance issue with the `odo` library in handling categorical dtypes.

jbednar:
[...]

PR #428, opened 2017-08-02
Remove dependency on odo. Optimize categoricals

gbrener:
This PR removes the dependency on `odo`, addressing #424. Some code was borrowed from `odo` and placed into [utils.py](https://github.com/bokeh/datashader/blob/optimize_categoricals/datashader/utils.py#L329-L356), but optimized to address #396 and be much more performant when handling categorical dtypes.


jbednar:
Looks great.  Ready to merge?

gbrener:
Yes, I believe so @jbednar. Forgot to mention - this PR also gets rid of the `dask` import warning about `get_sync` being moved.

PR #431, opened 2017-08-07
Sidestep datashape.Categorical bottleneck

gbrener:
This PR adds an additional performance enhancement which emerged when running `canvas.points()` over one thousand iterations with the dataset from #396. When categorical columns are detected, the `datashape.Categorical` constructor casts the categorical column to a tuple, resulting in a slowdown. Datashader now sidesteps this bottleneck by checking if a categorical dtype is in the dataframe; if so, the dataframe's columns are subset to only the "x", "y", and column(s) used for aggregation. Benchmarks suggest this leads to a ~40% performance increase on dataframes that contain categorical dtypes, and ~10% performance decrease on dataframes that don't contain categoricals.

### DIFF
--- a/datashader/core.py
+++ b/datashader/core.py
@@ -4,6 +4,7 @@ import numpy as np
 import pandas as pd
 import dask.dataframe as dd
 from xarray import DataArray
+from collections import OrderedDict
 
 from .utils import Dispatcher, ngjit, calc_res, calc_bbox, orient_array, compute_coords, get_indices, dshape_from_pandas, dshape_from_dask, categorical_in_dtypes
 from .resampling import (resample_2d, US_NEAREST, US_LINEAR, DS_FIRST, DS_LAST,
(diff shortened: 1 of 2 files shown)


datashader/core.py
```diff
@@ -368,10 +368,25 @@ def bypixel(source, canvas, glyph, agg):
...
     agg : Reduction
     """
-    if isinstance(source, pd.DataFrame):
-        dshape = dshape_from_pandas(source)
-    elif isinstance(source, dd.DataFrame):
-        dshape = dshape_from_dask(source)
+    # Avoid datashape.Categorical instantiation bottleneck
+    # by only retaining the necessary columns:
+    # https://github.com/bokeh/datashader/issues/396
+    if 'category' in str(source.dtypes.values):
```
jbednar: Is converting source.dtypes.values to a str the source of the 10% slowdown in the non-categorical case?  If so, is it faster not to convert it to a string?

jbednar:
Seems fine, thanks. It would be good if we could figure out a way not to incur any cost in the usual case, and there are also some test failures to address.

gbrener:
Ok, thanks for the feedback. The PR is still a work-in-progress - I'll fix the test failures and try to minimize the cost in the usual case.

datashader/core.py
gbrener: I speculate that this is the source of the slowdown, but I haven't confirmed yet. Before committing I tested a few different approaches for identifying whether a dataframe contains a categorical dtype, and this was the fastest method of the ones I tried. I may be able to get a speedup with a vectorized ufunc or a numba function; I'll try those shortly and push up the one that ends up being faster.

gbrener:
@jbednar Yes this will reorder the non-xy columns, but it keeps the xy columns at the beginning. We might be able to simplify this with a `set` if that property doesn't matter, but my benchmarks didn't show the `set` to be faster than the current approach. I'm not sure what we'll gain from copying the list of cols, since we don't know which cols we'll need until we look at all of the aggregations. Or maybe I'm not fully understanding what you mean?

jbednar:
I just mean that this looks like a complicated way that only guarantees to keep two of the column names the same order; wouldn't it be simpler just to keep *all* of the column names the same order, and simply delete any column name from that list if it's not needed?

gbrener:
@jbednar I just made the changes so that column order is preserved. This only impacts categorical performance very slightly, since list/dict comprehensions are very fast.

jbednar:
Looks good, thanks!  What's the overall impact on performance now, for the case with and without categoricals?

jbednar:
Ok; I see the performance numbers on the other PR.  Seems like this one is ready to merge, then?

gbrener:
Yep - ready to merge!

PR #432, opened 2017-08-08
Add categorical data to PCAP example

jbcrail:
Instead of filtering non-TCP traffic, we now add a category for the network protocol in order to distinguish traffic.

### DIFF
--- a/examples/pcap_to_parquet.py
+++ b/examples/pcap_to_parquet.py
@@ -20,32 +20,51 @@ def ip_to_integer(s):
     return struct.unpack("!I", socket.inet_aton(s))[0]
 
 
+def get_ip_protocol(s):
+    if "tcp" in s:
+        return "tcp"
+    if "UDP" in s:
+        return "udp"
+    if "EIGRP" in s:
+        return "eigrp"
+    if "ICMP" in s:
+        return "icmp"
+    return None
+
+
 def to_parquet(filename, prefix="maccdc2012"):
     with open(filename) as f:
         traffic = {}
         nodes = set()
 
         for line in f.readlines():
+            if "unreachable" in line:
+                continue
             fields = line.split()
             if not fields:
+                continue
+            if fields[1] != "IP":
+                continue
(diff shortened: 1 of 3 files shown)


examples/pcap_to_parquet.py
```diff
@@ -20,32 +20,51 @@ def ip_to_integer(s):
...
+    if "tcp" in s:
+        return "tcp"
+    if "UDP" in s:
+        return "udp"
+    if "EIGRP" in s:
+        return "eigrp"
+    if "ICMP" in s:
+        return "icmp"
+    return None
+
```
jbednar: I assume that this is a complex if statement because s contains other characters some of the time?  If not a direct lookup table (dictionary) would be both faster and clearer.

examples/pcap_to_parquet.py
```diff
@@ -20,32 +20,51 @@ def ip_to_integer(s):
...
+                continue
+            protocol = get_ip_protocol(line)
+            if protocol not in ("tcp", "udp", "eigrp", "icmp"):
+                continue
             try:
                 addresses = []
 
                 # Extract source IP address and convert to integer
-                m = re.match(r'\d+\.\d+\.\d+\.\d+', fields[2])
+                m = re.match(r'(?P<address>\d+\.\d+\.\d+\.\d+)', fields[2])
```
jbednar: Not sure if compiling the re beforehand would speed this up.

examples/pcap_to_parquet.py
jbcrail: My quick-and-dirty parsing approach was to split naively on spaces. Sometimes there are extra characters or punctuation in the original line, so a generic full-string search was simpler and more robust.

examples/pcap_to_parquet.py
jbcrail: Good catch. Given the 8.5M rows for one hour worth of data, I assume compiling beforehand would definitely help.

PR #445, opened 2017-08-23
Ignore IPython checkpoints for conda builds

jbcrail:
Even though we ignore checkpoint files when building pip packages, those files are still copied into conda packages.


jbednar:
Looks good, thanks.

PR #449, opened 2017-08-30
Fix notebook errors

### LINKED ISSUE #448: Errors in notebook examples
I ran through the examples on the master branch and encountered a few errors. I have an upcoming PR to fix some.

Need to fix:
- [x] `census-hv-dask.ipynb`
- [x] `nyc_taxi.ipynb`
- [x] `osm.ipynb`
- [x] `osm-1billion.ipynb`
- [x] `packet_capture_graph.ipynb`
- [x] `race_elevation.ipynb`
- [x] `tseries.ipynb`

Untested due to large dataset size:
- `opensky.ipynb`

Skipped for now due to multiple errors:
- `landsat.ipynb`

jbcrail:
Fix #448


jbednar:
These changes look good, but please track down which change to datashader or its dependencies made them necessary.

jbcrail:
I narrowed the reason for the change down to xarray, but I couldn't find an exact cause. I tried several xarray versions (0.7.2+) starting from August 2016 using the following example. There was significant churn in `xarray.concat` this past year, so it's hard to tell what behavior to expect.

```
import numpy as np
import pandas as pd
import xarray as xr

n = 10
cols = list("abc")
data = {col: list(range(0, n)) for col in cols}
data['time'] = list(range(0, 5*n, 5))

def make_array(name, data, x, y, p=0.9):
    dims = [x, y]
    coords = OrderedDict([(x, data[x]), (y, data[y])])
    data = np.random.choice(a=[False, True], size=(len(data[x]), len(data[y])), p=[p, 1-p])
    return xr.DataArray(data, dims=dims, coords=coords, name=name)

aggs = {col: make_array(col, data, col, 'time') for col in cols}
try:
    # Original code used in several notebook examples
    # KeyError exception occurs when dimensions have different names
    merged = xr.concat(aggs.values(), dim=pd.Index(cols, name='cols'))
    print("concatenated DataArrays with different dimension names")
except Exception as e:
    renamed = [aggs[key].rename({key: 'value'}) for key in aggs]
    merged = xr.concat(renamed, dim='cols')
    print("concatenated DataArrays with same dimension names")
print(merged.any(dim='cols'))
```

jbednar:
Ok, thanks.  I'm not sure it's specific to ``concat``; I think I've seen similar issues for other operators that combine or compare xarrays.  They are probably just checking things more strictly now, which makes certain operations less convenient, but should increase safety overall.

jbednar:
Although this change suggests that datashader itself has also changed?

-    "xs, ys = np.meshgrid(people.x_axis, people.y_axis)\n",		
+    "xs, ys = np.meshgrid(people.easting, people.northing)\n",

I.e. we used to generate dimensions named `x_axis`, and now we generate ones named `easting`?

jbednar:
In any case, I'm going to go ahead and merge this because I need to use these notebooks.

PR #450, opened 2017-09-01
Update example to match latest holoviews

jbednar:
Defaults for colors changed in holoviews for the `datashade()` operation in https://github.com/ioam/holoviews/pull/1697, and these changes ensure that the output for holoviews_datashader.ipynb is the same as before.


philippjfr:
Would you mind making the same fix in the Large_Data User Guide in HoloViews?

PR #452, opened 2017-09-05
Partial fix to landsat example; selected band even when only 1

jbednar:
I can't tell what changed to require this PR, but the landsat example is broken in current master. From the error messages, it looks like xarray used to ignore the ``band`` dimension when it has only a single value in a TIFF file, but now we have to explicitly select the value for that dimension if we want to get down to an x,y array.  

Still WIP, because the orientation of the image is now completely wrong.  It should be like:

image

But instead it's:

image

@jbcrail and @philippjfr , any idea why these changes were needed?  Any idea why the rotation is mixed up now (could it be a side effect of my "fixes"?)? Something to do with the new rasterio backend for xarray? The new regridding code? Any idea how to get it working?


jbednar:
Superseded by #453.

PR #453, opened 2017-09-05
Ensure shade transfer function respects array orientation

philippjfr:
Recent changes to ``canvas.raster`` now allow regridding of xarray DataArrays with ascending and descending coordinates. Currently the ``shade`` transfer function and the ``_interpolate`` and ``_colorize`` methods don't respect these orientations, resulting in flipped images in some cases. This PR uses the existing utilities to reorient the array to the expected orientation to fix this issue.

### DIFF
--- a/examples/landsat.ipynb
+++ b/examples/landsat.ipynb
@@ -194,13 +194,14 @@
    "outputs": [],
    "source": [
     "def combine_bands(r, g, b):\n",
+    "    r, g, b = [ds.utils.orient_array(img) for img in (r, g, b)]\n",
     "    a = (np.where(np.logical_or(np.isnan(r),r<=nodata),0,255)).astype(np.uint8)    \n",
     "    r = (normalize_data(r)).astype(np.uint8)\n",
     "    g = (normalize_data(g)).astype(np.uint8)\n",
     "    b = (normalize_data(b)).astype(np.uint8)\n",
     "    col, rows = r.shape\n",
     "    img = np.dstack([r, g, b, a]).view(np.uint32).reshape(r.shape)\n",
-    "    return tf.Image(data=img[::-1])"
+    "    return tf.Image(data=img)"
    ]
   },
   {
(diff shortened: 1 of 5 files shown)


examples/landsat.ipynb
```diff
@@ -203,7 +200,7 @@
     "    b = (normalize_data(b)).astype(np.uint8)\n",
     "    col, rows = r.shape\n",
     "    img = np.dstack([r, g, b, a]).view(np.uint32).reshape(r.shape)\n",
-    "    return tf.Image(data=img)"
+    "    return tf.Image(data=img[::-1])"
```
philippjfr: It's somewhat unfortunate this is needed, in general it is not safe to simply grab the ``.data`` off an xarray ``DataArray`` as is done in the function below because the coordinates could be oriented in any number of ways.

jbednar: Is ::-1 needed for this particular file, but may need to be different for a different file with some other orientation?  If so, can we not compute the "-1" (or "1") from the declared orientation?

philippjfr: Yes, I could use the ``orient_array`` utility here instead.

jbednar: Ok, sounds good. Mergeable after that?

philippjfr: (A better name for that utility is would be welcome btw).

philippjfr: I'll add some unit tests first.

jbednar: reorient_array? convert_array_to_canonical_orientation?  canonicalize_array?  They all seem as bad as each other.

philippjfr: Since we like none of them I'll just stick with what we've got.

philippjfr:
@jbednar Ready to review/merge.

jbednar:
Looks great, thanks!

PR #454, opened 2017-09-06
Ensure bundling retains order of edges

philippjfr:
Turned out to be a trivial fix, simply sorting by the index after performing the merge of the node and edge dataframes. Retaining this ordering is important if we want to use the edges for anything other than datashading.


jbednar:
Looks good to me...

PR #458, opened 2017-09-11
Add edge ID to bundled dataframe

jbcrail:
The edge ID is added to the bundled dataframe so that a user can retrieve the additional edge metadata without copying it into the dataframe.

The implementation was refactored for two reasons:

1. The edge segment constants and methods are collected into a specialized class. This replaces using the dimensions of an edge to differentiate.

2. We can now switch to structured NumPy arrays or XArray for easier reference to the individual points (instead of opaque numeric indices).

### DIFF
--- a/datashader/bundling.py
+++ b/datashader/bundling.py
@@ -207,10 +207,10 @@ def _convert_graph_to_edge_segments(nodes, edges, ignore_weights=False):
     the accumulator function for drawing to an image.
     """
 
-    df = pd.merge(nodes, edges, left_index=True, right_on=['source'])
+    df = pd.merge(edges, nodes, left_on=['source'], right_index=True)
     df = df.rename(columns={'x': 'src_x', 'y': 'src_y'})
 
-    df = pd.merge(nodes, df, left_index=True, right_on=['target'])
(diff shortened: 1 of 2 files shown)


philippjfr:
Looks good to me, one question I have is whether there is an option to control whether to include the index or not to avoid the additional memory overhead if you don't need it. Doesn't seem like a major issue though, just want to be clear.

jbcrail:
Thanks! The index and edge ids are mandatory at this time, but both can be made optional. It should be easier now to toggle edge ids as needed.

jbednar:
Ok, please make them optional and we can then merge this.

jbcrail:
@philippjfr I've made the edge IDs optional by default in the bundled dataframe.

Regarding the index, I'm unclear as to which index to make optional. Are you referring to the index on the provided edge dataframe?

philippjfr:
I really only meant the edge ids, sorry for the confusion in terminology above.

jbcrail:
No problem. Thanks for clearing this up.

jbednar:
Ok, looks like it's ready to merge.

PR #459, opened 2017-09-12
Fix upper-bound bin error for auto-ranged data

### LINKED ISSUE #457: Points dropped?  2D Histogram vs. DataShader
I'm wondering if there is a parameter required to capture the upper right corner point (1.2, 1.2) in datashader.

INPUT DATA
```
time,lon,lat
1,1.0,1.0
2,1.1,1.1
3,1.2,1.2
```

**2DHistogram:** 
```
        df=df[['_x','_y']].compute()
        minx=df._x.min()
        maxx=df._x.max()
        miny=df._y.min()
        maxy=df._y.max()
        xbins=outputsizedict['xpixels']
        ybins=outputsizedict['ypixels']

        xedges, yedges = np.linspace(minx,maxx,xbins), np.linspace(miny,maxy,ybins)
        hist, xedges, yedges = np.histogram2d(df._x, df._y, (xedges, yedges))
        xidx = np.clip(np.digitize(df._x, xedges, right=True), 0, hist.shape[0]-1)
        yidx = np.clip(np.digitize(df._y, yedges, right=True), 0, hist.shape[1]-1)
        z=np.zeros((yedges.shape[0],xedges.shape[0]),dtype=np.uint32)
        z[yidx,xidx]=hist[xidx,yidx]
        z=np.flipud(z)
```

**Datashader:** 
```
agg=cvs.points(df, '_x', '_y', ds.count('_t'))
img=np.flipud(agg)
```

RESULT from 2DHIST method
image

```
[[0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0]
 [0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 1 0]
 [0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0]
 [0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0]
 [0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0]
 [0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0]
 [0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0]
 [0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0]
 [0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0]
 [0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0]
 [0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0]
 [0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0]
 [0 0 0 0 0 0 0 0 0 0 0 0 1 0 0 0 0 0 0 0 0 0 0 0 0]
 [0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0]
 [0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0]
 [0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0]
 [0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0]
 [0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0]
 [0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0]
 [0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0]
 [0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0]
 [0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0]
 [0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0]
 [0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0]
 [1 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0]]
```

RESULT from datashader
image

```
[[0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0]
 [0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0]
 [0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0]
 [0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0]
 [0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0]
 [0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0]
 [0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0]
 [0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0]
 [0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0]
 [0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0]
 [0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0]
 [0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0]
 [0 0 0 0 0 0 0 0 0 0 0 0 1 0 0 0 0 0 0 0 0 0 0 0 0]
 [0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0]
 [0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0]
 [0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0]
 [0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0]
 [0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0]
 [0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0]
 [0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0]
 [0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0]
 [0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0]
 [0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0]
 [0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0]
 [1 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0]]
```

jbcrail:
Fix #457


jbednar:
Thanks, @jbcrail. I'm trying to decide whether this is truly the appropriate fix. 

To start with, I don't think either of the behaviors shown in #457 is correct -- previously, datashader resulted in the point being just over the edge to the top right and thus not visible, but the numpy.histogram2d example had an extra row and column on one side, which doesn't seem correct either. With this PR, datashader now does the right thing:  

image

Given our upper-exclusive bounds policy, it achieves this result by enlarging the floating-point auto range just barely enough to ensure that all points will map into the resulting set of pixels, without changing any user-supplied range.  This clearly works, but one implication is that if the user does compute an auto range based on the max and min, the max points won't be included in the plot.

One could argue that such behavior is just doing what the user asks, but we should consider whether it would be less surprising to users if we just changed the definition of x_range and y_range to be upper inclusive in general, e.g. by adding np.spacing to compute_scale_and_translate instead.  That might surprise users less and is less likely to hide data points unbeknownst to the users.

Any opinions, @jbcrail, @philippjfr, and @hsparra?

apiszcz:
Naive question, why doesn't auto-scale should include all values from the
input. If the user supplies min,max values i'm guessing that is the same.

If both methods are in question is there a more fundamental problem, I was
'hoping 2D histogram and datashader would have the same result.

BTW: I'm comparing the results with datashader (before this change) and 2d
histogram on larger datasets, they are always different, furthermore on
very large sparse input datashader exhausts memory on problems where 2D
histogram does not.

jbednar:
Internally, datashader has to map every single datapoint to one, and only one, grid cell.  That's true everywhere in the array, from the bottom left to the top right, and the only way that can be achieved for all locations, including those precisely on the borders, is if two of the borders on each grid cell are inclusive, and two are exclusive.  If the borders were all inclusive, a datapoint on the border between two grid cells would be counted in both of them, which is clearly incorrect statistically.  Unfortunately, if that policy is applied consistently to all cells, which for good performance is the only reasonable choice, then at the border of the overall array, some of the borders will be exclusive.  Having those borders be exclusive is fine as a definition, but with autoranging it's clearly not the right overall answer, and is clearly confusing and surprising.  This PR works around the problem by bumping up the two exclusive edges just very slightly so that they include the value on the edge, effectively making the outer borders all be inclusive and thus avoiding the problem in the autoranging case but not if someone naively supplies bounds that they expect to be inclusive.

The behavior of histogram2d is not correct in this case, at least not based on the example you supplied, and so we are certainly not going to match that (adding an extra dummy row and column).  But I would like to address the specific problem of not including the edge datapoints, as this PR addresses.

Memory and performance issues should be addressed in other issues, which I think have been raised already, but those will take some investigation.

apiszcz:
I recently observed this difference between DS and NP 2DH, in places other
than the edges 'obviously' ? due to the approach you describe.

CONCUR with counting on the seams.

Should there be a reference aggregation data set for benchmark and known
results? I have not reviewed all of your test cases perhaps there is one.
Something tiny simple, medium and 'large'.

Too much to do, however correct or repeatable results would may be helpful.
Comparing the methods of implementation through inspection of outputs for:
2DH, DS, (there are some other techniques out there, which 'very' fast with
hash codes), etc.
I also 'wonder' what the GRID (weather) community has for their
aggregations.

jbcrail:
@jbednar I agree that extending the upper-exclusive bounds policy for the entire array could be confusing and surprising to users. Anecdotally, I pre-calculated the min and max when generating my first Datashader examples, so I also ran into this issue. I think upper-inclusive bounds are the least surprising and probably why they are used in other frameworks. Here's why upper-exclusive bounds might cause confusion:

First, the `x_range` and `y_range` parameters are similar in name to Python's `range`, thus, implying an upper-exclusive range (i.e. `[a, b)`). If the parameters were `x_limit` and `y_limit` instead (like other plotting frameworks) or used lists for values instead of tuples, then an upper-inclusive range (i.e. `[a, b]`) would be natural.

Second, even though Datashader only handles numeric values, if we were plotting dates or strings, then an upper-inclusive range would also be least surprising, e.g. plotting `A` to `Z` or `2017-Jan` to `2017-Dec`.

As you mentioned, I think `compute_scale_and_bounds` is the right place to add np.spacing. When I added it to `_compute_x_bounds` and `_compute_y_bounds`, I had to adjust the tests to account for the new auto-range value. This didn't feel right, since the implementation was bleeding into the tests.

jbednar:
OK, can you please update the PR to add the fix in the new spot?

jbednar:
Let me know when it's ready to review, once the tests pass...

jbcrail:
I had two strategies to address the uniform distribution of points, which we did not previously test.

The first strategy was to increase the upper bound on the x-axis and y-axis by the next nearest floating point value by using `np.spacing`. This changed the scale and translate value when mapping to the axis, thus, making some bins slightly larger. The details of this change were messy and I found it hard to get all tests to pass.

The second strategy (currently used) was to adjust the point and line glyphs to decrement data that landed on either upper bound using `np.spacing`. This made both ranges fully inclusive without changing the bin sizes. One positive side effect was that I refactored the glyph code to be clearer. In particular, the line glyph uses only the non-mapped floating points for range checks before mapping to integer values and then drawing the line. This simplified the code and may address #464 though I haven't tested it yet.

datashader/glyphs.py
```diff
@@ -92,14 +92,21 @@ def _build_extend(self, x_mapper, y_mapper, info, append):
...
             xmin, xmax, ymin, ymax = bounds
+
+            def map_onto_pixel(x, y):
+                xx, yy = x_mapper(x) * sx + tx, y_mapper(y) * sy + ty
+                if x == xmax:
+                    xx -= np.spacing(xx)
+                if y == ymax:
+                    yy -= np.spacing(yy)
+                return int(xx), int(yy)
+
```
jbednar: I have *not* tested this, but would it be better to fix it up in the integer domain, instead of relying on np.spacing to be sufficient to bump it back by an integer?

```
yy = int(y_mapper(y) * sy + ty)

return (xx-1 if x == xmax else xx,
        yy-1 if y == ymax else yy)
```

datashader/glyphs.py
```diff
@@ -169,13 +170,16 @@ def _compute_outcode(x, y, xmin, xmax, ymin, ymax):
...
-                int(x_mapper(x1) * sx + tx),
-                int(y_mapper(y0) * sy + ty),
-                int(y_mapper(y1) * sy + ty))
+        _, xmax, _, ymax = bounds
+        xx, yy = x_mapper(x) * sx + tx, y_mapper(y) * sy + ty
+        if x == xmax:
+            xx -= np.spacing(xx)
+        if y == ymax:
+            yy -= np.spacing(yy)
+        return int(xx), int(yy)
```
jbednar: Same comment as above regarding np.spacing vs. -1.

datashader/glyphs.py
jbcrail: Adjusting in the integer domain instead fails compared to the original. For instance,
```
x_mapper, y_mapper = lambda n: n, lambda n: n
xmax, ymax = 1.0, 1.0
sx, tx, sy, ty = 0.1, 0.1, 0.1, 0.1
```
given the point (1.0, 1.0), the original algorithm returns (0, 0) while the above returns (-1, -1).

datashader/glyphs.py
jbcrail: Same as above.

datashader/glyphs.py
jbcrail: I used this script to compare implementations:
```python
import numpy as np


def map_onto_pixel(x, bound, s, t):
    xx = x * s + t
    if x == bound:
        xx -= np.spacing(xx)
    return int(xx)

def map_onto_pixel2(x, bound, s, t):
    xx = int(x * s + t)
    return xx-1 if x == bound else xx

def floats(origin, variance):
    fs = [origin]
    n = origin
    for _ in range(variance):
        n = n - np.spacing(n)
        fs.append(n)
    n = origin
    for _ in range(variance):
        n = n + np.spacing(n)
        fs.append(n)
    fs.sort()
    return fs

bound = 1.0
values = [0.1 * n for n in range(10)]
for s in values:
    for t in values:
        for f in floats(bound, 10):
            a = map_onto_pixel(f, bound, s, t)
            b = map_onto_pixel2(f, bound, s, t)
            if a != b:
                print("{:.20f} {} {} {} {}".format(f, a, b, s, t))
```

jbednar: Thanks.  That script's output is very confusing for me, because it always prints "f" as "1.00000000000000000000", even though f is clearly varying around 1.0, even if I change the precision to 40 or 60.  Is that the same thing you see?  I can get the actual value by printing ``str([f])`` instead, but there must be a better way!

jbcrail: The two implementations only differ when f is 1.0, so it's the only value shown. If you remove the `if` statement, you should see the other values of f.

jbednar: Right; I delete my comment once I saw that. :-)  Now I'll delete both!

jbednar: For that script, all the cases where the two methods differ are when f==bound, which makes sense given that both versions will otherwise always return ``int(x * s + t)``.  But what I can't tell is which answer (if either) is correct when the two do differ.  When they differ, is that meant to be a valid actual input, given that the points are being cropped against the specified bounds already?

jbcrail: The value of the first implementation, `a`, is assumed to be the correct value since its the current implementation used by all of the passing tests. The script was meant to verify the second implementation would be a drop-in replacement.

I just verified that the newer implementation passes the tests, which makes me believe more investigation may be needed to improve the test cases.

jbednar: I'm not sure that's true; the test cases could well be fine.  It's true that the above script finds differences, but I can't tell whether it's finding them in cases that would ever be able to be encountered in practice, given that s and t are not arbitrary, but are presumably derived from the bounds (and the aggregate array size), and given that the points are already being cropped against the bounds.  What would be helpful would be to take a case where the values are observed to differ, and then figure out whether that case could ever happen, given how s and t are actually calculated.  My guess is that it couldn't ever happen, but I have not worked it through myself.  And given that both approaches pass the tests, I am not at all confident that the existing approach in the PR is safer or more general.

jbcrail: I agree with your assessment. I took my modified script and created additional uniformity test cases. Both implementations produced the same results with the test suite, so I replaced the mapping with the simpler implementation.

Issue #407, opened 2017-07-13
Run (some) example notebooks on travis?


ceball:
Would you take a PR that runs the example notebooks on travis, or are the examples too big to run on travis? What about a subset? (Or perhaps eventually the notebooks could be modified to take parameters, allowing scaled-down data to be used...)

I mean something like ``jupyter nbconvert example/notebook.ipynb --execute --ExecutePreprocessor.timeout=90 --ExecutePreprocessor.kernel_name=python3 --to html`` for each notebook.

Optionally, the html files could be put on s3 so it's possible to see them, but this is probably not necessary (nbconvert shows which cell gives an error, and the last line of the traceback).

jbcrail:
I think this would be an excellent idea. Since most notebooks use large remote datasets, we'll have to use a smaller subset as you suggested to minimize test run time. I also agree that storing the html on s3 shouldn't be necessary.

The following notebooks are self-contained and auto-generate their data:

* pipeline.ipynb
* plotting_pitfalls.ipynb
* trajectory.ipynb
* tseries.ipynb

Let's try these notebooks first to see how Travis handles them.

jbednar:
I think most of the notebooks should work unmodified with stripped-down test data, though a few might have hardcoded bits that depend on the data size.  It would be nice to store some collection of the stripped down versions for use by Travis and then validate against that continuously.  In the longer run we'll have most of the notebooks built as part of the website, but having this stripped-down set of data will still be useful for regression testing even then.

ceball:
Minor update: we're up to one notebook running on travis+appveyor so far, but getting closer to running more of them with "small" datasets/equivalent.

jbednar:
Smoke tests are now supported, and we'll be moving the topics out to pyviz.org and so won't need separate tiny-data tests.

Issue #404, opened 2017-07-13
Travis does not follow any one set of installation instructions


ceball:
Travis has its own list of dependencies (I'm not counting extra testing dependencies), and those dependencies don't appear to be the exactly same as one of examples/environment.yml, conda.recipe/meta.yaml, or requirements.txt.

I think travis should be testing one of the existing lists, but which one? Maybe examples/environment.yml, since that's the one suggested to developers, and the travis build is really for developers? I'm not sure.

ceball:
Sorry, not just the dependencies, but also e.g. ``python setup.py develop --no-deps`` vs instruction to ``pip install -e .``.

jbednar:
I must admit that I have no idea why there are so many lists of dependencies.  Can't we have just one list for datashader itself, and generate any other list needed for that?  As we aren't currently testing the examples, I'd assume that this one main (fairly lean) set of dependencies should be the one used by Travis, not the full environment.yml.

Using pip install -e makes sense to me.

ceball:
I could create a poll and everyone could vote for their favorite dependencies/versions! Or maybe we need a new dependencies/requirements specification format to rule them all? ;)

I'm not yet in a position to know why the lists are different - maybe there are valid reasons, and the lists have been arrived at through hard-won practical experience.

Can someone make 'required', 'optional',  'examples' lists out of these? (Or something like that.)

# travis

```
conda install
   1. colorcet
   2. dask
   3. datashape  
   4. numba
   5. numpy
   6. odo  
   7. pandas
   8. param
   9. pillow
  10. pytest
  11. rasterio
  12. scikit-image
  13. toolz
  14. xarray

testing
  conda install
    a. matplotlib
    b.
    c. 
```

# appveyor

```
conda install
   1. colorcet
   2. dask
   3. datashape  
   4. numba
   5. numpy
   6. odo
   7. pandas  
   8. param
   9. pillow
  10. pytest
  11. rasterio -c conda-forge
  12. scikit-image
  13. toolz
  14. xarray

testing:
  conda install 
    a. matplotlib
    b.
    c. 
```

# conda/meta.yaml

```
channels:
  - numba

run:
   1. colorcet >=0.9.0
   2. dask >=0.7.6
   3. datashape >=0.5.1
   4. numba >=0.24.0
   5. numpy >=1.7
   6. odo >=0.4.1
   7. pandas >=0.15.0
   8. param >=1.5.0,<2.0
   9. pillow >=3.1.1
  10.
[...]

PR #463, opened 2017-09-18
CI: run example notebooks.

ceball:
See #407.

The "test_notebooks" script will, I hope, be replaced by something better in the future. Meanwhile, seems important to at least check that the notebooks run.

I have not attempted to address creation of scaled-down data. If this PR is merged, I will open a separate issue about that.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,16 +28,16 @@ install:
 
   # Optional dependencies, for testing only
   - conda install matplotlib
-  # Dependencies for lint checking only
+  # Dependencies for notebook & lint checking only
   - conda install jupyter flake8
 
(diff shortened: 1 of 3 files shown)


ceball:
The failure relates to holoviews not being installed on travis (required by one of the notebooks suggested in https://github.com/bokeh/datashader/issues/407#issuecomment-323607932), which I think relates to #404.

ceball:
On reflection, I doubt that the datashader developers want to become responsible for maintaining this script (which is really nothing to do with datashader). I guess if such a thing were desired, it would be better to exist as a standalone package that could become a testing dependency.

PR #466, opened 2017-09-18
Enable appveyor

ceball:
Includes allowing python 3.4 to fail (copying travis config).


ceball:
I've added datashader to bokeh-integrations's appveyor account; builds are public and PRs get built.

PR #467, opened 2017-09-18
Reorganize/update CI testing

ceball:
1. Add osx (using osx_image to get a recent osx version).  

2. Switch to language=generic from language=python. Travis's python was not previously being used (conda is being used), but travis's python was still being installed. Apart from being unused, travis's auto-installation of python on osx fails.  

3. Drop python 3.4. Python 3.4 has already effectively been dropped because it's been failing every time (allowed failure).  

4. Some additional minor reorganization.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,40 +1,54 @@
-language: python
-sudo: false
-python:
-    - "2.7"
-    - "3.4"
-    - "3.5"
-    - "3.6"
+# Since Python doesn't work on osx travis images (see
+# https://github.com/travis-ci/travis-ci/issues/4729), and because
+# this project installs miniconda and uses conda rather than using
+# travis's python, we're deliberately not using travis's
+# language=python option.
 
-matrix:
-  fast_finish: True
-  # Allow py3.4 to fail; conda-forge no longer supports it
-  allow_failures:
-    - python: "3.4"
+language: generic
+
+os:
+  - linux
+  - osx
+
+# get a more recent version of osx than the default image
+osx_image: xcode8.3
+
+sudo: false
+
+env:
+  - PYTHON_VERSION="2.7"
+  - PYTHON_VERSION="3.5"
+  - PYTHON_VERSION="3.6"
 
 install:
   # Install conda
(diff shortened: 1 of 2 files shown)


ceball:
@jbednar, I requested a review thinking travis was about to finish, but it turns out travis is currently experiencing problems with its osx service. The push build has finished but the pr build is stuck in a queue. Sorry about that - you'll have to come back later...

.travis.yml
```diff
@@ -1,40 +1,54 @@
...
+  - PYTHON_VERSION="3.6"
 
 install:
   # Install conda
-  - wget http://repo.continuum.io/miniconda/Miniconda-latest-Linux-x86_64.sh -O miniconda.sh
+  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then
+      wget https://repo.continuum.io/miniconda/Miniconda3-latest-MacOSX-x86_64.sh -O miniconda.sh;
+    else
+      wget https://repo.continuum.io/miniconda/Miniconda3-latest-Linux-x86_64.sh -O miniconda.sh;
+    fi
```
jbednar: I'd rather see a lookup table than an if statement, but (a) that would require more bash-fu than I know, and (b) isn't urgent given we have only two platforms like this in the foreseeable future.

ceball: I think you're right. But to explain...

I think everyone is copying (and will continue to copy) https://conda.io/docs/user-guide/tasks/use-conda-with-travis-ci.html. I could discuss those instructions with conda, but I've also been wondering if conda could somehow provide an integration for travis (https://github.com/travis-ci/travis-cookbooks ?) and then various boilerplate copied into many projects could be removed, along with probably a load of documentation.

Meanwhile, what we have here is bash inside yaml. Although I've become more familiar with both of them recently, mix them together like this and I can't get one line right without trying about 5 times. I expect we could simplify by using more yaml and less bash, though.

jbednar: Ok by me.

jbednar: Ready to merge?

ceball: Yes

PR #475, opened 2017-09-22
Fix inverted array indexing and off-by-one errors in canvas.raster

philippjfr:
While testing the regridding implementation @jbednar identified a bug in the code, which were caused by array indices being inverted in certain cases. I tried for a long time to get the previous approach using affine transforms working but kept running into various off-by-one issues, which seemed to be caused by rounding issues in the affine transform code. I've now replaced this code with an approach that simply converts continuous coordinates to array indices by normalizing and then rescaling them. Not only do I understand this approach better but it now works consistently.

If you have a minute it would be good if you could review this @gbrener as I'm still trying to figure out what the affine transform approach gives us that's not handled by this simpler normalize and rescale approach.

### DIFF
--- a/datashader/tests/test_raster.py
+++ b/datashader/tests/test_raster.py
@@ -138,12 +138,12 @@ def test_raster_both_ascending_partial_range():
     ys = np.arange(5)
     arr = xs*ys[np.newaxis].T
     xarr = xr.DataArray(arr, coords={'X': xs, 'Y': ys}, dims=['Y', 'X'])
-    cvs = ds.Canvas(9, 4, x_range=(-.5, 8.5), y_range=(-.5, 3.5))
+    cvs = ds.Canvas(7, 3, x_range=(.5, 7.5), y_range=(.5, 3.5))
     agg = cvs.raster(xarr)
 
-    assert np.allclose(agg.data, arr[1:, :-1])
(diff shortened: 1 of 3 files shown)


datashader/tests/test_raster.py
```diff
@@ -138,12 +138,12 @@ def test_raster_both_ascending_partial_range():
     ys = np.arange(5)
     arr = xs*ys[np.newaxis].T
     xarr = xr.DataArray(arr, coords={'X': xs, 'Y': ys}, dims=['Y', 'X'])
-    cvs = ds.Canvas(9, 4, x_range=(-.5, 8.5), y_range=(-.5, 3.5))
+    cvs = ds.Canvas(7, 3, x_range=(.5, 7.5), y_range=(.5, 3.5))
```
jbednar: Can you motivate this change to the tests?  Is it now testing something more than it was before?

philippjfr: Yes, it's now slicing different numbers of elements along each edge. Previously I was only slicing off the upper bound, which let some bugs slip past.

jbednar: Sounds good.

philippjfr:
Thanks for the background @gbrener. We can revisit if we need to handle any other transformations supplied on the attributes.

PR #476, opened 2017-09-22
Example notebook improvements

philippjfr:
Various improvements for datashader example notebooks. So far just makes some small fixes to census examples but still working on revising network examples.


examples/census-hv-dask.ipynb
```diff
@@ -170,6 +184,7 @@
    },
    "outputs": [],
    "source": [
+    "shaded = datashade(census_points, **shade_defaults)\n",
```
jbednar: Duplicated here to avoid inappropriate sharing of streams? Ok.

examples/census-hv-dask.ipynb
```diff
@@ -133,7 +145,9 @@
    "source": [
     "shape_path = 'data/cb_2015_us_cd114_5m/cb_2015_us_cd114_5m.shp'\n",
     "districts = gv.Shape.from_shapefile(shape_path, crs=crs.PlateCarree())\n",
-    "districts = gv.operation.project_shape(districts)"
+    "districts = gv.operation.project_shape(districts)\n",
+    "districts = hv.Polygons([np.column_stack([xs, ys]) for dist in districts\n",
+    "                     for xs, ys in zip(*gv.util.geom_to_array(dist.data))])"
```
jbednar: That's a bit mysterious.  Why hv.Polygons to replace gv.Shape?

examples/census-hv-dask.ipynb
philippjfr: Yes.

examples/census-hv-dask.ipynb
philippjfr: True, mysterious but a lot more efficient. That said I can actually make an optimization in geoviews, ensuring that we only take the hit on initialization. Previously it was spending about 1 second on each zoom event to unpack the shape into an array of arrays.

jbednar: Doing that in geoviews does sound like useful; this will *definitely* mystify people and they won't know why they are doing it or why performance is bad if they don't.

jbednar:
I agree that census-hv-dask should be renamed, but that name's a bit long.  How about just census-congressional?

philippjfr:
@jbednar Are there any other notebooks you want me to go over? Otherwise this is ready to be merged, do note however that it the graph notebook depends on:

https://github.com/ioam/holoviews/pull/1951

and good performance on the congressional districts notebook depends on:

https://github.com/ioam/holoviews/pull/1952
https://github.com/ioam/geoviews/pull/90

jbednar:
Well, if you're asking, pretty much all of the notebooks need a going-over, but the main one that needs attention is landsat.ipynb.   Instead of InteractiveImage, it needs to use HoloViews with regrid (or, ideally, datashade, after making datashade do regridding as we discussed elsewhere).  That's important on its own, because it illustrates how to use raster data with HoloViews+Datashader, but also because it would make it possible to show the hyperspectral image data that's in those files.  I had a discussion with someone (I've now forgotten whom!) who was imagining having an image plot that when you  hover or click on it, will update an adjacent Curve plot with the full spectrum of radiant energy being measured at that location. Basically, we have 11 wavelength bands available, with a lookup chart at the top of that notebook listing what they are, and we can take any one of the image plots already in there and then for a given coordinate look up one number from that location in each of the 11 files to make a spectrum curve.  Seems like a cool example of HV, to me!

philippjfr:
This could now be merged but it would be nice to switch the pcap example to use the inbuilt datashader forceatlas layout rather than relying on NetworkX. We'd need #485 to be fixed first though.

examples/landsat.ipynb
```diff
@@ -225,14 +225,13 @@
     "band_map = hv.HoloMap({i: hv.Image(band) for i, band in enumerate(bands)})\n",
     "\n",
     "def spectrum(x, y):\n",
+    "    point = gv.Points([(x, y)], crs=ccrs.GOOGLE_MERCATOR)\n",
+    "    point = gv.operation.project_points(point, projection=ccrs.PlateCarree())\n",
+    "    label = 'Lon: %.3f, Lat: %.3f' % tuple(point.array()[0])\n",
     "    try: \n",
```
jbednar: I should probably have explained why I reorganized that function the way I did; it was to avoid starting up with bogus labels like:  

image

It's ok to leave the lat,lon coordinates up there when the curve disappears, as you do now, but I don't want it to start with nan,nan shown, because that's how the page will look on a website when it's exported statically.

philippjfr: Without an initial label the title initializes inside the plot for me, which I think is worse than showing the nans.

jbednar: Seems like a bug report to file somewhere? In that case, can you just use a non-empty string for the initial label, e.g. "Lon:      , Lat:      "?  Or even "_"?

jbednar:
I'm ready to merge as soon as the above issue is addressed.  I think we should deal with forceatlas2 separately, preserving this version as a reference.  Thanks for all your excellent contributions here!

examples/landsat.ipynb
jbednar: nan just looks like a bug, even though it is accurate in this case (as there is no default).

philippjfr: Don't think bokeh can do anything about that. The layout gets initialized with whatever is on the screen at the time, if there's no title to begin with it won't allocate the appropriate space for it and when one is added later it won't show up. We maybe be able to address it by triggering resize events.

jbednar: I wouldn't want a resize event to happen; that would be distracting.  But there's no need for a resize event, because Bokeh lines up by plot edge anyway, with or without a title, if one of the plots has a title:

image

So I do think having the label later show up in the middle of the plot is a bug, not a natural consequence.  Anyway, your fix is fine for this PR, thanks!

PR #480, opened 2017-09-25
Fix segmentation fault due to improper indices

### LINKED ISSUE #477: hammer_bundle causes segmentation fault
Since the recent fix for ``hammer_bundle`` in #473 running it crashes the Python process with an segmentation fault:

```python
import dask.dataframe as dd
from datashader.layout import circular_layout
from datashader.bundling import hammer_bundle

edges_df = dd.read_parquet('./data/maccdc2012_edges.parq').compute()
nodes_df = dd.read_parquet('./data/maccdc2012_nodes.parq').compute()

positions = circular_layout(nodes_df, edges_df)
bundled = hammer_bundle(positions, edges_df)
```

Segmentation fault: 11

jbcrail:
Since edge IDs are optional, the exact coordinate indices for a segment can change. I added pre-defined x-y coordinate indices for a given segment to each subclass.

This will be deprecated if we switch over to structured NumPy arrays or xarray.

Fix #477 and #481

### DIFF
--- a/datashader/bundling.py
+++ b/datashader/bundling.py
@@ -114,18 +114,18 @@ def resample_edges(edge_segments, min_segment_length, max_segment_length, segmen
 
 
 @nb.jit
-def smooth_segment(segments, tension):
+def smooth_segment(segments, tension, idx, idy):
     seg_length = len(segments) - 2
     for i in range(1, seg_length):
         previous, current, next_point = segments[i - 1], segments[i], segments[i + 1]
-        current[1] = ((1 - tension) * current[1]) + (tension * (previous[1] + next_point[1]) / 2)
-        current[2] = ((1 - tension) * current[2]) + (tension * (previous[2] + next_point[2]) / 2)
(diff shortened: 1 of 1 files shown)


philippjfr:
Will try this out now.

philippjfr:
Okay, I can confirm this works now, but I also noticed that bundling appears significantly slower now. I'll try to produce some concrete numbers on that.

jbednar:
I'm happy to merge this functionality, but first it would be good to know if there is indeed a performance impact, in which case somehow maybe it can be declared to numba that those values are constant per loop?  Or would we need to simply (but painfully) specialize the code instead to handle each of the two cases less generally but more quickly?

jbcrail:
I didn't intend to conflate two issues into one, but the segfault, jagged paths, and performance decrease were intertwined. This PR should fix all three.

philippjfr:
Thanks @jbcrail, I tested it and it's working great!

PR #482, opened 2017-09-26
Dynamically change the legend size to fit the number of items.

kgullikson88:
The current categorical legend seems designed for exactly 5 items. This allows to legend to expand to fit however many items there are. I also added a `font_size` parameter so the user can have smaller/bigger fonts if they want. The resulting legend will be smaller/bigger as well.


jbednar:
I'm amazed that the legend code is working at all, but if it is, great!  We've got to get some time to make the legend support really work well and be usable in general, but in the meantime these changes look very reasonable. Thanks!

PR #486, opened 2017-10-03
Optimize resampling operations

philippjfr:
The resampling operations we copied from gridtools are all in a massive monolithic numba-ized function, which has several disadvantages, which I'll try to outline:

1) Huge functions with deep nesting are unreadable
2) The entire function needs to be compiled in one go which slows down the initial compile step (takes about ~5 seconds)
3) The JIT compiler has to validate types and since np.sqrt is apparently only defined for float types all the data **must** be of float type even for downsampling methods which do not use ``np.sqrt``.

While splitting the huge function into individual functions should only be seen as an initial step towards refactoring this code, it already has two major advantages:

1) No super long compile step to begin with.
2) No requirement to cast the input arrays to floats unless using downsampling with standard deviation and variance methods

When running resampling on the landsat data (using mean downsampling) the timings before and after this PR are:

PRE:

  * Compilation: 6.88 s
  * Downsampling: 2.58 s

POST:

  * Compilation: 1.56 s
  * Downsampling: 1.26 s

### DIFF
--- a/datashader/core.py
+++ b/datashader/core.py
@@ -265,8 +265,6 @@ class Canvas(object):
         left, bottom, right, top = calc_bbox(xvals, yvals, res)
         array = orient_array(source, res, layer)
         dtype = array.dtype
-        if dtype.kind != 'f':
-            array = array.astype(np.float64)
 
         if nan_value is not None:
             mask = array==nan_value
@@ -298,10 +296,14 @@ class Canvas(object):
             source_window = array[rmin:rmax+1, cmin:cmax+1]
             if isinstance(source_window, Array):
                 source_window = source_window.compute()
+            if downsample_method in ['var', 'std']:
+                source_window = source_window.astype('f')
             data = resample_2d(source_window, **kwargs)
             layers = 1
         else:
             source_window = array[:, rmin:rmax+1, cmin:cmax+1]
+            if downsample_method in ['var', 'std']:
+                source_window = source_window.astype('f')
             arrays = []
             for arr in source_window:
                 if isinstance(arr, Array):
@@ -348,11 +350,12 @@ class Canvas(object):
             data = data.filled()
 
         # Restore original dtype
-        data = data.astype(dtype)
+        if dtype != data.dtype:
+            data = data.astype(dtype)
 
         # Compute DataArray metadata
         xs, ys = compute_coords(self.plot_width, self.plot_height, self.x_range, self.y_range, res)
-        coords = {xdim: xs.astype(xvals.dtype), ydim: ys.astype(yvals.dtype)}
+        coords = {xdim: xs, ydim: ys}
         dims = [ydim, xdim]
         attrs = dict(res=res[0])
         if source._file_obj is not None:
--- a/datashader/resampling.py
+++ b/datashader/resampling.py
@@ -37,7 +37,7 @@ try:
     def _sentinel_function(): return 0
     _sentinel_function()
 except:
-    ngjit_parallel, prange = ngjit, range
+    ngjit_parallel, prange = ngjit, range # NOQA
 
 
 #: Interpolation method for upsampling: Take nearest source grid cell, even if it is invalid.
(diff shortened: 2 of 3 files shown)


datashader/core.py
```diff
@@ -348,11 +350,12 @@ def raster(self,
...
 
         # Restore original dtype
-        data = data.astype(dtype)
+        if dtype != data.dtype:
+            data = data.astype(dtype)
 
         # Compute DataArray metadata
         xs, ys = compute_coords(self.plot_width, self.plot_height, self.x_range, self.y_range, res)
-        coords = {xdim: xs.astype(xvals.dtype), ydim: ys.astype(yvals.dtype)}
+        coords = {xdim: xs, ydim: ys}
```
philippjfr: I added this before but realized it will result in poor behavior when downsampling or upsampling with integer coordinates. The output coordinates will just always be floats.

philippjfr:
Okay, I didn't stop there and enabled numba parallelization. On my machine the results are now:

* PRE: 2.58 s
* POST (not parallel): 1.26 s
* POST(parallel): 0.515 s

datashader/resampling.py
```diff
@@ -27,15 +27,14 @@
...
-from .utils import ngjit, ngjit_parallel
+from .utils import ngjit
 
 try:
-    # Determine if system supports numba 'parallel' target
-    @ngjit_parallel
-    def _sentinel_function(): return 0
-    _sentinel_function()
+    # Try to create numba JIT with 'parallel' target
+    ngjit_parallel = nb.jit(nopython=True, nogil=True, parallel=True)
```
ceball: Not sure if I am being helpful here (rather than just annoying!) because I am not really following this and I got involved by mistake :) But shouldn't the figuring out of stuff happen in util, where it only needs to happen once?

I.e. in util, you'd do the

```
try:
    ngjit_parallel = nb.jit(parallel=True,...)
    from numba import prange
except:
    ngjit_parallel = ngjit
    prange = range
```

Then all other files wouldn't need to worry about it (they'd just import `ngjit_parallel` and `prange` from util).

I also think you wouldn't need the NOQA this way.

However I might well be missing something. Or it might be bad in that maybe things are less explicit in the files where stuff's actually done. 

Also, finally, maybe this whole approach could result in eventually everyone silently missing out on the parallel stuff when prange is moved within numba, or nb.jit changes in some way, or whatever - anything where stuff changes in the future so that these lines ends up failing for everyone silently and nobody gets parallelized stuff (and worse, nobody knows about it). So maybe at least a warning? And/or maybe catching only whatever error it is you're expecting (`RuntimeError`?). I guess the idea being that for datashader you really don't want to be running without all the optimizations possible, rather than a common opposite requirement of wanting things to at least work even if optimizations are unavailable.

philippjfr: True, that seems somewhat cleaner. I think leaving it in resampling is fine for now since ``parallel`` is only needed there. Even without it being parallel it's still 2-3x faster than it was before, so I'm not sure the warning is necessarily needed. I'll leave it up to @jbednar if he wants a warning, if not I'll probably not bother changing it.

jbednar: Those are all good comments.  I'm happy either way; the platforms where we expect parallel to fail are all platforms no one in their right mind should be using, as far as I can see, so I don't mind polluting them with warnings (which would ordinarily be a concern).

ceball: > the platforms where we expect parallel to fail are all platforms no one in their right mind should be using, as far as I can see

Python 2.7 on 64-bit Windows doesn't seem completely crazy to me.

ceball: > polluting them with warnings (which would ordinarily be a concern)

I'd like to see some kind of system that warns about all the loud, pink warnings that are going to appear in my notebooks if I choose to use certain packages ;)

jbednar:
Thanks!

datashader/resampling.py
jbednar: That's true; Python 2.7 on 64-bit Windows isn't ideal, but isn't crazy; I'm not sure from the ambiguous error message whether that is in the category of platforms not currently supported.

Issue #485, opened 2017-10-01
forcealtas2 layout does not seem to be working.


hsparra:
The results of the `forcealtas2_layout` do not match expectations based upon network force directed layout. Also, the `circle_layout` produces a very similar image.

Using PCAP example data
```
df = forceatlas2_layout(nodes_orig, edges_df)
fda2_img = tf.shade(cvs.line(df[1:], 'x', 'y'), cmap=fire)
tf.set_background(fda2_img, color='black’)
```
forceatlas2 results
![datashader_layouts 2017-10-01 17-03-26](https://user-images.githubusercontent.com/7515105/31059669-c2083946-a6cb-11e7-9877-6aae62193ad6.png)

circle_layout results
![datashader_layouts 2017-10-01 17-08-28](https://user-images.githubusercontent.com/7515105/31059665-b2ac762e-a6cb-11e7-97b0-394407e9ebb6.png)

hsparra:
I have it working for circles:

```
positioned_df = circular_layout(nodes_orig, edges_df)
positioned_df = positioned_df.set_index('id')
direct_circle = directly_connect_edges(positioned_df, edges_df)
cir_direct = hv.Curve(direct_circle)
datashade(cir_direct)
```
![datashader_layouts 2017-10-01 20-03-22](https://user-images.githubusercontent.com/7515105/31061160-9bd1606e-a6e3-11e7-9dc9-458d8356b218.png)

hsparra:
ForceAtlas2 still looks off:
```
fa_positioned_df = forceatlas2_layout(nodes_orig, edges_df)
fa_positioned_df = fa_positioned_df.set_index('id')
fa_edges = directly_connect_edges(fa_positioned_df, edges_df)
fa_direct = hv.Curve(fa_edges)
datashade(fa_direct)
```
![datashader_layouts 2017-10-01 20-05-52](https://user-images.githubusercontent.com/7515105/31061200-0b370350-a6e4-11e7-8126-09722d8c42cb.png)

hsparra:
For reference here is the networkX for directed version:
![datashader_layouts 2017-10-01 14-30-01](https://user-images.githubusercontent.com/7515105/31061209-33b3160c-a6e4-11e7-8a4d-185332b4897e.png)

philippjfr:
As far as I can tell the algorithm still doesn't work well. Here's the new result after 10 iterations:

[...]

PR #488, opened 2017-10-05
Improvements to forceatlas2 algorithm

jbcrail:
The datatype of the coordinates is now consistent. We also check for edges pointing to themselves.

Related to #485

### DIFF
--- a/datashader/layout.py
+++ b/datashader/layout.py
@@ -90,11 +90,11 @@ class circular_layout(LayoutAlgorithm):
         return df
 
 
-def _extract_points_from_nodes(nodes):
+def _extract_points_from_nodes(nodes, params, dtype=None):
     if 'x' in nodes.columns and 'y' in nodes.columns:
         points = np.asarray(nodes[['x', 'y']])
     else:
-        points = np.asarray(np.random.random((len(nodes), 2)))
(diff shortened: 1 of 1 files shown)


datashader/layout.py
```diff
@@ -90,22 +90,22 @@ def __call__(self, nodes, edges, **params):
...
     nlen = len(nodes)
     if 'id' in nodes:
         index = dict(zip(nodes['id'].values, range(nlen)))
     else:
         index = dict(zip(nodes.index.values, range(nlen)))
 
-    if 'weight' not in edges:
+    if params.ignore_weights or 'weight' not in edges:
         edges = edges.copy()
         edges['weight'] = np.ones(len(edges))
```
jbednar: This seems like a very expensive step, and doesn't seem feasible as an approach in general, but I guess we can merge and come back to that later.

jbcrail:
After profiling the memory usage for sparse matrix creation, it surprisingly turned out that removing the temporary tuple list had a far larger impact than the updated steps taken when ignoring weights. The overall time didn't change noticeably in either case.

jbednar:
Ok, sounds great!  I'll merge when tests pass.

PR #492, opened 2017-10-10
Added Images class to format a list of images as an HTML table

jbednar:
Added Images class to format a list of images as an HTML table, to make it simpler to write notebooks and documentation for datashader.  Supports labelling each image with its xarray name, and so also added support for passing name into the various ways to create the image.

### DIFF
--- a/datashader/transfer_functions.py
+++ b/datashader/transfer_functions.py
@@ -35,7 +35,50 @@ class Image(xr.DataArray):
         return fp
 
 
-def stack(*imgs, **kwargs):
+class Images(object):
+    """
+    A list of Images that are expected to be displayed in a table.
+    """
+    
+    def __init__(self, *images):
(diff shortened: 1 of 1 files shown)


jbednar:
With this PR, you can now do things like:

image

Open question: Why is the background a darker gray?  And why does it highlight when the mouse hovers over it?

jbednar:
I think the answer to why the table version is a darker gray and why it highlights is this change to jupyter: https://github.com/jupyter/notebook/pull/1776

This behavior makes sense for tables of text:

image

But not for tables used just for formatting purposes, as here. The relevant CSS appears to be in [this file](https://github.com/jupyter/notebook/blob/master/notebook/static/notebook/less/renderedhtml.less), but I'm not quite sure how to cram it into the inline styles that I can see using here.

PR #494, opened 2017-10-14
Parameterize node/edge columns

### LINKED ISSUE #493: Parameterize graph layout/bundling columns
Currently the graph layout and bundling API hard-codes the node/edge columns.

We need to update this so that:

1. users can use their own columns without a rename step prior to layout/bundling
2. we adhere to the conventions in the core methods (e.g. `points` and `line`)

jbcrail:
This depends on the merger of #488 so that I can simplify how weights are ignored for layout.

Fix #493

### DIFF
--- a/datashader/bundling.py
+++ b/datashader/bundling.py
@@ -257,7 +257,7 @@ class EdgelessWeightedSegment(BaseSegment):
         img[int(point[0] * accuracy), int(point[1] * accuracy)] += point[2]
 
 
-def _convert_graph_to_edge_segments(nodes, edges, include_edge_id, ignore_weights=False):
+def _convert_graph_to_edge_segments(nodes, edges, include_edge_id, use_weights=True):
     """
     Merge graph dataframes into a list of edge segments.
 
@@ -284,7 +284,7 @@ def _convert_graph_to_edge_segments(nodes, edges, include_edge_id, ignore_weight
     if include_edge_id:
         df = df.rename(columns={'id': 'edge_id'})
 
-    include_weight = not ignore_weights and 'weight' in edges
+    include_weight = use_weights and 'weight' in edges
 
     if include_edge_id:
         if include_weight:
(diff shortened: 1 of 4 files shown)


jbednar:
Looks good so far, though I wonder if it would be simpler if you just changed some of the standalone functions to which you are passing ``params`` into methods, so that they could access the params directly.

jbcrail:
I agree about changing how parameters are accessed. It was messy passing the dictionary around.

I fixed it for both bundling and layout by assigning the overridden parameters to the local instance parameters. It required a few lines of boilerplate for each class, so there may be a cleaner way to accomplish it.

datashader/layout.py
```diff
@@ -32,8 +32,8 @@ class LayoutAlgorithm(param.ParameterizedFunction):
     target = param.String(default='target', doc="""
         Column name for each edge's target.""")
 
-    weight = param.String(default='weight', doc="""
-        Column name for each edge weight.""")
+    weight = param.String(default='weight', allow_None=True, doc="""
+        Column name for each edge weight. If None, weights are ignored.""")
```
jbednar: I think None is a better default than 'weight'.

datashader/bundling.py
```diff
@@ -442,7 +444,10 @@ class hammer_bundle(directly_connect_edges):
         Column name for each edge weight. If None, weights are ignored.""")
 
     def __call__(self, nodes, edges, **params):
+        # Merge user-provided parameters with local parameters
         p = param.ParamOverrides(self, params)
+        for k in p:
+            setattr(self, k, p[k])
```
jbednar: Oops; I didn't see these diffs earlier, only the later ones.  This isn't the right approach, as it will be very confusing for users to see arguments they supply to a call being the new persistent values. Definitely would be a big source of confusing bugs for users.  Instead, would need to pass the ``p`` object to the method below that uses it.  Alas, I think that renders my previous suggestion moot and counterproductive -- there is no increase in simplicity by making these functions into methods if they still have to access the parameters via a supplied object.  So I guess this whole commit should be reverted?  Sorry about that; entirely my fault for not looking at the context in which those methods were being used.

jbednar:
Looks great now, thanks!

PR #502, opened 2017-10-18
Updated InteractiveImage for compatibility with bokeh 0.12.10

philippjfr:
Adds support for using the bokeh messaging protocol in bokeh version > 0.12.9, which is required for InteractiveImage to work in bokeh 0.12.10 and above.

### DIFF
--- a/datashader/bokeh_ext.py
+++ b/datashader/bokeh_ext.py
@@ -1,21 +1,122 @@
 from __future__ import absolute_import
 
+from distutils.version import LooseVersion
+
 import uuid
 import json
 
 import numpy as np
+import bokeh
 
-from bokeh.io import notebook_div
+
 from bokeh.document import Document
 from bokeh.models import (CustomJS, ColumnDataSource, Square, HoverTool,
                           GlyphRenderer)
-from bokeh.io import _CommsHandle
-from bokeh.util.notebook import get_comms
 from bokeh.models import Plot, Text, Circle, Range1d
 from bokeh.plotting import Figure
 
 from . import transfer_functions as tf
 from .utils import downsample_aggregate, summarize_aggregate_values
+
+bokeh_version = LooseVersion(bokeh.__version__)
+
+if bokeh_version > '0.12.9':
+    from bokeh.protocol import Protocol
+    from bokeh.embed import notebook_content
+    from bokeh.embed.notebook import encode_utf8
+    from bokeh.io.notebook import CommsHandle, get_comms
+else:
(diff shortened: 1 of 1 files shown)


datashader/bokeh_ext.py
```diff
@@ -188,18 +208,35 @@ def update(self, ranges):
...
+        comm = self.comms_handle.comms
+        if bokeh_version > '0.12.9':
+            comm.send(msg.header_json)
+            comm.send(msg.metadata_json)
+            comm.send(msg.content_json)
+            for header, payload in msg.buffers:
+                comm.send(json.dumps(header))
+                comm.send(buffers=[payload])
+        else:
+            comm.send(msg)
```
jbednar: Can you wrap this up a bit more simply, with the 0.12.9 version being encapsulated into a function to send a message, and then only switch between two versions of that function?  I think that would be easier for us to maintain.

datashader/bokeh_ext.py
```diff
@@ -188,18 +208,35 @@ def update(self, ranges):
...
         """
         Generate an update event json message.
         """
+        if bokeh_version > '0.12.9':
+            events = list(self.doc._held_events)
+            if not events:
+                return None
+            msg = Protocol("1.0").create("PATCH-DOC", events)
+            self.doc._held_events = []
+            return msg
```
jbednar: The 0.12.9 and previous versions are so entirely different that again I think it would be helpful if you wrapped up what each one does as some sort of coherent function, with a reasonable name, and then simply switched between them.  I'd rather such complexities be put into separate functions that can be ignored when understanding the overall logic of this class.

datashader/bokeh_ext.py
```diff
@@ -228,7 +265,14 @@ def _repr_html_(self):
...
         self.doc.add_root(self.p)
-        return notebook_div(self.p, self.ref)
+        if bokeh_version > '0.12.9':
+            js, div, _ = notebook_content(self.p, self.ref)
+            html = NOTEBOOK_DIV.format(plot_script=js, plot_div=div)
+            div = encode_utf8(html)
+            self.doc.hold()
+        else:
+            div = notebook_div(self.p, self.ref)
+        return div
```
jbednar: Sigh!  Bokeh 0.12.10 sure seems to need more verbose code.  Again, making the new bit into a function with a name that explains what the heck it achieves would make this class be so much more readable...

philippjfr:
Made the requested changes. Should be ready to merge once tests pass.

PR #511, opened 2017-10-25
Run Travis-CI osx tests on master only.

ceball:
Mac on travis is slow, and has been for a while. Therefore, we wondered about running Mac tests only after merging a PR to master. Otherwise, tests on PRs take too long.

We hope that generally if the build passes on linux, it'll pass on Mac - but if on the odd occasion it does fail on Mac, we want to know about.

(As requested by Jim - but please still consider whether or not you want this. There are downsides, e.g. potential confusion.)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -39,6 +39,7 @@ jobs:
       script: scripts/ci/test_examples
       os: osx
       osx_image: xcode8.3
+      if: branch = master AND type != pull_request
       env:
       - GROUP=examples
       - PYTHON_VERSION=3.6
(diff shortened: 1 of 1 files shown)


jbednar:
Thanks. Yes, there are downsides, such as the danger of merging things that later break Macs specifically, but I think the upsides outweigh them: faster PR merging (helping keep progress going), and less temptation to merge before the PR tests complete (which leaves confusing records in the logs about whether the PR was ok before merging).  Plus, most of our developers tend to be on OS X when developing, so Mac-specific problems are less likely than for other platforms, making the Travis and Appveyor tests most useful on non-Mac platforms.

Issue #286, opened 2017-02-19
Plotting many lines takes a long time


StevenCHowell:
I am hoping there is a misunderstanding or misuse on my part but plotting lots of lines with Datashader is taking much longer than I would expect, at least compared to plotting points.  To provide an example case, I wrote the following code which demonstrates what I am seeing.  As a summary, this figure shows how many seconds it takes to plot `n` lines, each with 50 data points.
![ds_line_test](https://cloud.githubusercontent.com/assets/9484229/23105572/61290be0-f6af-11e6-83e3-2c4f3afdfec7.png)
While this shows it scales as n log(n), better than n^2, the large amount of time, over 4.5 hours for 8192 lines, makes me wonder if it is actually running using optimized numba code.  Again, I hope this is simply misuse on my part as I want to use this to plot over 100,000 lines.  I have got similar results when running on in linux desktop (12 core, 32 GB ram), and on my mac laptop.  I did run this in a jupyter notebook; would it perform better in a python script?

Here is the code I am running:
```
import pandas as pd
import numpy as np
import datashader
import bokeh.plotting
import collections
import xarray
import time
from bokeh.palettes import Colorblind7 as palette

bokeh.plotting.output_notebook()

# create some data worth plotting
x = np.linspace(0, np.pi * 2)
y = np.sin(x)
n = 100000
data = np.empty([n+1, len(y)])
data[0] = x
prng = np.random.RandomState(123)
offset = prng.normal(0, 0.1, n).reshape(n, -1)
data[1:] = y
data[1:] += offset
df = pd.DataFrame(data.T)
x_range = 0, 2*np.pi
y_range = -1.5, 1.5
y_cols = range(1, n+1)

# iterate over increasing number of lines
run_times = []
imgs = []
test = [2, 4, 8, 16, 32, 64, 128, 256, 512, 1024, 2048, 4096, 8192]
for n in test:
    this_df = df[:n+1]
    y_cols = range(1, n+1)
    tic = time.time()
    canvas = datashader.Canvas(x_range=x_range, y_range=y_range, 
                               plot_height=300, plot_width=300)
    aggs = collections.OrderedDict((c, canvas.line(df, 0, c)) for c in y_cols)
    merged = xarray.concat(aggs.values(), dim=pd.Index(y_cols, name='cols'))
    img = datashader.transfer_functions.shade(merged.sum(dim='cols'), how='eq_hist')
    toc = time.time() - tic
    run_times.append(toc)
    imgs.append(img)

[...]

PR #512, opened 2017-10-26
Plotting large numbers of sequences/time series together: dealing with fixed length numpy arrays

narendramukherjee:
Added a function in ds.utils to convert time series/sequences stored as 2D numpy arrays to a dataframe with NaN separators between individual sequences. Also added an example in tseries.ipynb showing the use of this function. This is in response to issue #286


narendramukherjee:
@jbednar @philippjfr Can you take a look and let me know what you think.

jbednar:
Thanks for the PR!  

In most of our use cases for datashading curves, we want to be able to distinguish between the curves, which is only feasible for up to a few dozen curves if we use count_cat to colorize them.  Here, there doesn't seem to be a way to convey the identity of each curve, but it seems like you have an application in mind where that doesn't matter?  E.g. maybe you could talk about how this approach lets you discover underlying periodicities in pseudorandom number generators?  That's what it *looks* like your example is showing:

image

image

jbednar:
Oops; the periodicities are just due to having many fewer items in your sequence than the number of points in the plot, which again is unusual. The bumps are just for each number involved:

image

image

Still trying to think of a way to motivate what an example like this will be used for...

narendramukherjee:
My specific use case involves looking at 1-2ms long voltage traces from neurons (action potentials) and determining if they are coming from the same neuron. Each neuron produces stereotypical action potentials, and plotting all the action potentials recorded on a electrode on top of each other let's us know if they are coming from one or multiple neurons. So, yes, in my case, the exact identity of each curve doesn't really matter. Check out our Scipy paper again for severely overplotted examples of this kind: http://conference.scipy.org/proceedings/scipy2017/narendra_mukherjee.html

I think that this sort of use case isn't that uncommon - the original question in #286 was trying to achieve exactly this sort of thing. I just used a pseudorandom number generator as an easy way to generate 'dummy' data of the kind I am plotting - I could as well put in my specific use case, with action potentials from a neuron as an example, but that would mean I would have to put in some actual data that I have recorded as well to make those plots work. I didn't know how to do that with a IPython notebook.

Let me know what you think!

jbednar:
I had forgotten that you were the one with the SciPy paper, which I do remember now!  

A use case something like that was what I was imagining, but in that case, won't you want to know the identity of the inappropriately sorted curves, the ones with shapes that suggest that they are not action potentials for this neuron, so that you can exclude them from the group?  I agree that a visualization like this is a good first step, to at least be able to see them, but then if it were my data I'd immediately want to start pulling out the outlier curves and see why they ended up in this bucket inappropriately, which is difficult if I can't identify them.  

Maybe in practice what you do is just adjust some threshold, never dealing with individual curves by name or id?  In that case I guess a good visualization would be to overlay a datashaded plot of the traces included by the threshold in one color, over a datashaded plot of the ones excluded in another color, adjusting the threshold until those two groups were quite visibly distinct.  Doing that shouldn't require anything further from datashader, but it sure seems like it would be helpful to have an example that shows a workflow like this.  

I wonder if there's a good way to do that with synthetic data, synthesizing a bunch of curves from different categories, pooling them all together, and then showing how to use datashader to see visually that there are these categories and then adjusting thresholds until a clustering algorithm correctly sorts out each category.  Hmm; probably too ambitious, so I guess I should just merge this utility as-is and think about that later!

jbednar:
Ok, I tidied up the example notebook a bit to remove extraneous changes and to use an example where each datapoint was countable for clarity, and merged it. Thanks for your contribution!

PR #520, opened 2017-11-06
Improve aggregation of streamed dataframes

jbcrail:
Prior to this, the aggregated images contained an odd purplish background and the data points were unclear.

Now, instead of concatenating and then summing the data, we now just sum the list of xarrays. We also manually set the canvas's range and background, passing that to the helper methods.

Relies on #523 to fix failing tests


jbednar:
Sounds good.  Happy to merge once tests are passing.

PR #524, opened 2017-11-09
Remove interfering brew-installed libraries from travis/osx

ceball:
Based on a discovery in a project I'm working on, seems like all travis osx images have a variety of brew installed packages (which can change over time). Also seems that some conda packages will pick up such 'system' libraries in preference to conda supplied ones. Those are both difficult things for us to address properly, but we can at least remove interfering system libraries from travis as we discover them. In any case, these kinds of issues can affect users too (many have brew installed libraries).

If the build passes, I'll revert the changes that make osx test the PR.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,10 @@ sudo: false
 
 before_install:
   - export PATH="$HOME/miniconda/bin:$PATH"
+  # brew-installed geos interferes with cartopy?
+  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then
+      brew uninstall --ignore-dependencies geos gdal postgis;
+    fi
 
 jobs:
   include:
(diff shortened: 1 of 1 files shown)


ceball:
The osx tests all failed because travis/osx can't reach github right now. Will revisit later.

ceball:
Because of the travis/github integration, it's difficult to see, but this did pass on osx before I restored the don't-test-PRs-on-osx behavior (which canceled the rest of the builds...). 

What fails on master:

screen shot 2017-11-09 at 6 01 23 pm

You can see it passed earlier on this PR before I disabled osx on PRs again:

screen shot 2017-11-09 at 6 01 57 pm

https://travis-ci.org/bokeh/datashader/jobs/299392021

PR #525, opened 2017-11-14
Add Canvas.trimesh(), triangle mesh rasterization

gbrener:
This adds a new method to the Canvas class called `Canvas.trimesh()`. Usage examples:

Basic usage, one value per vertex (`z`):
```
verts = pd.DataFrame({'x': [...], 'y': [...], 'z': [...]}, columns=['x', 'y', 'z'])
tris = pd.DataFrame({'v0': [...], 'v1': [...], 'v2': [...]}, columns=['v0', 'v1', 'v2'])
agg = cvs.trimesh(verts, tris)
# etc..
```

Basic usage, one value per triangle (`value`):
```
verts = pd.DataFrame({'x': [...], 'y': [...]}, columns=['x', 'y'])
tris = pd.DataFrame({'v0': [...], 'v1': [...], 'v2': [...], 'value': [...]}, columns=['v0', 'v1', 'v2', 'value'])
agg = cvs.trimesh(verts, tris)
# etc..
```

Interpolation can be disabled with the `interp` flag, like so:
```
agg = cvs.trimesh(verts, tris, interp=False)
```

During downsampling, the aggregation defaults to `ds.mean`. Other aggregations can be provided with the `agg` option. The following does a "minimum" aggregation using the `z` column:
```
agg = cvs.trimesh(verts, tris, agg=ds.min('z'))
```

The winding direction of the vertices (`v0`, `v1`, `v2` above) is auto-detected for pandas dataframes, and is assumed to be consistent throughout the data. Dask dataframes are not (yet) supported - at the moment they are simply converted to pandas - but I plan to add dask support in a future PR. The approach behind the `agg` feature is explained more thoroughly in #533 . Interpolation is bilinear - additional control over interpolation, and/or alternate interpolation strategies, may be added in future PR(s).


jbednar:
Looks great!

datashader/core.py
```diff
@@ -196,11 +199,113 @@ def line(self, source, x, y, agg=None):
...
+            if (isinstance(vertices, dd.DataFrame) and
+                    isinstance(simplices, dd.DataFrame)):
+                # TODO: Avoid .compute() calls, and add winding auto-detection
+                vals = vertices.values.compute()[simplices.values[:, :3].compute()]
+                vals = vals.reshape(np.prod(vals.shape[:2]), vals.shape[2])
+                CACHED_SOURCE = pd.DataFrame(vals, columns=vertices.columns)
+                if weight_col is not None:
+                    CACHED_SOURCE[weight_col] = simplices.values[:, 3].compute().repeat(3)
+            else:
+                simplex_coords = simplices.values[:, :3]
```
philippjfr: Not sure if this is ready for review but in testing I found that if I add a weight column to the simplices, this array and the ``simplices.values[0, :3]`` array used below become float types causing float indexing errors.

gbrener:
`agg` should now be working, and defaults to `ds.mean`. Also, there is now an `interp` boolean flag that serves as an on/off switch for interpolation.

I updated the PR description to reflect the latest status and API. Planning to defer the following items to future PRs: dask dataframe support, wireframe-only viz, and perf improvements that will speed up processing for the other glyphs. I also plan to add more rigorous testing when I get the chance.

datashader/core.py
gbrener: Thanks for letting me know. This issue should be fixed now.

gbrener:
I think this PR is ready to merge. It's failing for Python 3.6 due to one of the notebooks - I'm not able to reproduce the notebook issue locally, and it seems unrelated to my changes.

datashader/core.py
```diff
@@ -196,11 +196,121 @@ def line(self, source, x, y, agg=None):
...
+
+        Parameters
+        ----------
+        vertices : pandas.DataFrame, dask.DataFrame
+            The input datasource for triangle vertex coordinates. These can be
+            interpreted as the x/y coordinates of the vertices, with optional
+            weights for color interpolation. Columns should be ordered
+            corresponding to 'x', 'y', followed by zero or more (optional)
+            columns containing vertex values. The rows need not be ordered, and
+            the data type must be from the set of real numbers.
```
jbednar: What does it mean for a data type to be from the set of real numbers?  The set of real numbers doesn't include any types, but I assume you meant that the column type must be integer or real?If so, that's fine, as I've reworded it, but if not please correct this.

datashader/core.py
```diff
@@ -196,11 +196,121 @@ def line(self, source, x, y, agg=None):
...
+                    winding = [0, 2, 1]
+
+                vertex_idxs = simplices.values[:, winding].astype(np.int64)
+                vals = vertices.values[vertex_idxs]
+                vals = vals.reshape(np.prod(vals.shape[:2]), vals.shape[2])
+                source = pd.DataFrame(vals, columns=vertices.columns)
+                if not verts_have_weights:
+                    source[weight_col] = simplices.values[:, 3].repeat(3)
+        else:
+            source = mesh
```
jbednar: How is the mesh argument meant to be used?  The docstring doesn't say enough to make it usable, and I would have thought that the code here would all be in some external function that would be called here but which the docstring could suggest that the user call explicitly if they want to avoid having to redo this step.

datashader/glyphs.py
```diff
@@ -73,6 +75,34 @@ def _compute_y_bounds_dask(self, df):
...
 
 
+class _PolygonLike(_PointLike):
+    """_PointLike class, with Triangle-specific methods overriden.
+
+    Key differences from _PointLike:
+        - added self.z as a list, representing vertex weights
+        - constructor accepts additional kwargs:
+            * weight_type (bool): Whether the weights are on vertices (True) or on the triangles (False)
+            * interp (bool): Whether to interpolate (True), or to have one color per triangle (False)
```
jbednar: Is this class implementing polygons, or triangles?  The name indicates polygons, but all the docstrings indicate triangles. I've updated the docs to say that it's about shapes, but if that's not correct please rename and fix the documentation.

datashader/core.py
philippjfr: +1 on providing a function that computes ``mesh`` for you. Doesn't have to be in this PR though.

jbednar:
Sure, just waiting on the tests to complete before I merge.  If you want to look at my recent changes and vouch for them (they *should* be trivial!), then feel free to squash merge before they complete.

jbednar:
Merging since it's passing on Linux.

datashader/core.py
gbrener: Thanks for the feedback. I'll revise the docstring, and provide a utility function to create the mesh data structure, in my next PR.

Issue #540, opened 2017-12-07
TODOs related to Canvas.trimesh


gbrener:
The following issue describes outstanding TODO items related to the `Canvas.trimesh()` feature (from #525):

## API
- [ ] Write utility function (in _ds.utils_) to help generate the data structure accepted by the `mesh` keyword argument
- [ ] Add xarray data structure to datashader's pipeline, via built-in dispatch mechanism, so that the `trimesh` method can accept it as an argument (related to #526, https://github.com/ContinuumIO/earthio/issues/13, https://github.com/ContinuumIO/elm/issues/163, https://github.com/ContinuumIO/xarray_filters/issues/1)
- [ ] With additional interpolation methods (in "Misc Features", below), the `interp` flag should be something other than a boolean

## Docs
- [ ] Notebook for user guide, describing the available options and what they do
- [ ] Notebook for examples directory, plus dataset(s) in _examples/downloads.py_ illustrating sample use-case w/ inference(s)

## Dask
- [ ] Enhance dask dataframe support, to leverage datashader's built-in dispatch mechanism -
 i.e. instead of immediately loading dask dataframes into memory
- [ ] Add vertex winding auto-detection (and correction) for dask dataframes

## Performance
- [ ] Make improvements to to _datashader/compiler.py_, so that the performance optimizations can be applied to other glyphs
- [ ] Other performance-enhancing activities. Current approaches I'd like to explore (time permitting):
    - [ ] Adapting the rasterization algorithm to be a numba gufunc
    - [ ] Leveraging a spatial index to speed up traversal/exclusion of triangles outside viewing area (open question on how to use this with dask)
    - [ ] Partitioning the mesh so it can be processed in parallel (via multithreading, multiprocessing, or distributed/remote workers)
    - [ ] Adapting the rasterization algorithm to run on the GPU

## Misc Features
- [ ] Add the ability to visualize only triangle edges (like a "wireframe")
- [ ] Add other interpolation methods (TBD). This is a prerequisite to an API change mentioned above.

jbednar:
> Notebook for topics directory, plus dataset(s) in examples/datasets.yml

I think we should just use this one example, which I've uploaded to S3 and thus can now be added to datasets.yml:

https://s3.amazonaws.com/datashader-data/Chesapeake_and_Delaware_Bays.3dm

PR #546, opened 2017-12-12
Add ds.utils.mesh

gbrener:
Addresses the first API enhancement in #540 by adding a utility function: `mesh()`, which handles both pandas dataframes and dask dataframes. This utility function supports the `Canvas.trimesh()` API, returning a data structure that can be passed directly via the `mesh` keyword-argument. In addition, the `Canvas.trimesh()` method was simplified and now relies on this utility internally.

### DIFF
--- a/datashader/core.py
+++ b/datashader/core.py
@@ -249,66 +249,31 @@ class Canvas(object):
         """
         from .glyphs import Triangles
         from .reductions import mean as mean_rdn
+        from .utils import pd_mesh, dd_mesh
 
-        cols = vertices.columns.tolist()
-        x, y, weights = cols[0], cols[1], cols[2:]
+        source = mesh
 
-        # Verify the simplex data structure
-        assert simplices.values.shape[1] >= 3, ('At least three vertex columns '
-                                                'are required for the triangle '
-                                                'definition')
-        simplices_all_ints = simplices.dtypes.iloc[:3].map(
-            lambda dt: np.issubdtype(dt, np.integer)
-        ).all()
-        assert simplices_all_ints, ('Simplices must be integral. You may '
-                                    'consider casting simplices to integers '
-                                    'with ".astype(int)"')
-
-        # Choose reduction based on vertex coordinate data
-        weight_col = None
-        verts_have_weights = not not weights
-        if verts_have_weights:
-            weight_col = weights[0]
-            if agg is None:
-                agg = mean_rdn(weight_col)
-            elif agg.column is None:
-                # Rasterization code assumes presence of agg column
-                agg.column = weight_col
-        else:
-            assert simplices.values.shape[1] > 3, 'If no vertex weight column is provided, a triangle weight column is required.'
-            weight_col = simplices.columns[3]
-            if agg is None:
-                agg = mean_rdn(weight_col)
-            elif agg.column is None:
-                # Rasterization code assumes presence of agg column
-                agg.column = weight_col
-
-        if mesh is None:
+        # Validation is done inside the [pd]d_mesh utility functions
+        if source is None:
             if (isinstance(vertices, dd.DataFrame) and
                     isinstance(simplices, dd.DataFrame)):
-                # TODO: For dask: avoid .compute() calls, and add winding auto-detection
-                vertex_idxs = simplices.values[:, :3].astype(np.int64)
-                vals = vertices.values.compute()[vertex_idxs]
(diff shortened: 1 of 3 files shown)


philippjfr:
Looks good to me!

datashader/core.py
```diff
@@ -237,78 +237,42 @@ def trimesh(self, vertices, simplices, mesh=None, agg=None, interp=True):
...
-                    winding = [0, 2, 1]
-
-                vertex_idxs = simplices.values[:, winding].astype(np.int64)
-                vals = vertices.values[vertex_idxs]
-                vals = vals.reshape(np.prod(vals.shape[:2]), vals.shape[2])
-                source = pd.DataFrame(vals, columns=vertices.columns)
-                if not verts_have_weights:
-                    source[weight_col] = simplices.values[:, 3].repeat(3)
+                source = pd_mesh(vertices, simplices)
+
```
jbednar: Can we move this code out as well, to be the user-level API for these functions, i.e.:

```
def mesh(vertices, simplices):
    """..."""
    if (isinstance(vertices, dd.DataFrame) and isinstance(simplices, dd.DataFrame)):
        return _dd_mesh(vertices, simplices)
    else:
        return _pd_mesh(vertices, simplices)
```

It seems like that name will be more obvious to the user, who normally doesn't have to think a lot about whether they are using pandas or dask dataframes.

datashader/utils.py
```diff
@@ -383,3 +383,74 @@ def dataframe_from_multiple_sequences(x_values, y_values):
...
+    assert simplices.values.shape[1] >= 3, ('At least three vertex columns '
+                                            'are required for the triangle '
+                                            'definition')
+    simplices_all_ints = simplices.dtypes.iloc[:3].map(
+        lambda dt: np.issubdtype(dt, np.integer)
+    ).all()
+    assert simplices_all_ints, ('Simplices must be integral. You may '
+                                'consider casting simplices to integers '
+                                'with ".astype(int)"')
+
```
jbednar: This validation code could then move into mesh() to avoid duplication.

datashader/core.py
gbrener: Sure, sounds reasonable. I'll have it done shortly.

datashader/utils.py
gbrener: I thought about putting the validation code in one place to avoid duplication, however I kept them separate due to the subtle differences between the dask and pandas APIs. They ended up being the same in this case, but the xarray version will almost certainly be different. I'll default to your judgement on this one, though.

jbednar: No need to preemptively add complexity; just make a special case for xarray if and when it is needed.  So please do combine them for now.  I find duplicating error messages to be especially confusing in practice, because it's then not possible to track down which bit of code generated the message.  Much better for all messages to have one point of orgin...

gbrener:
Updated the implementation to use one function called `ds.utils.mesh()`. Also updated the PR name and description. cc @jbednar

gbrener:
The Python 3.6 build is resulting in a strange error, which seems unrelated to this PR:

screen shot 2017-12-14 at 12 27 50 pm

Happy to look into it further, if someone else who is more familiar with the example notebooks is not available.

datashader/utils.py
jbednar: Thanks for the update!  The same issue applies at the end of each function, too, though; the entire bit after "# If vertices don't have weights..." is the same in both and contains an error message, so please move that into the same function to make the differences between the two versions clearer, to avoid code duplication, and to give a single generation point for each error message.

jbednar: Damn, I see the ".compute()" now.  Any obvious way around duplicating that, then?

jbednar: It would be nice if pandas made .compute() into a no-op to allow compatibility!

gbrener: Just pushed up the change for de-duplication.

jbednar: Best I can see is to write a trivial helper function like:

```
def compute(obj,dask=False):
    return obj.compute() if dask else obj
```

gbrener: I'm planning to add dask support in another PR, in a way that hopefully won't require us to use `.compute()` at all; I'm only using it here as a stopgap until I get the chance to make the code more dask-friendly.

jbednar: Ok.

jbednar:
The build seems to be erroring because hiveplot.com is down.  I'll go ahead and merge, but we'll need to link to something else if it stays down.

PR #550, opened 2018-01-16
Small optimizations to trimesh pd_mesh utility

philippjfr:
Minor optimizations to ``_pd_mesh`` utility which computes the datastructure passed to the trimesh aggregation. Checking the type before casting has a modest improvement when the data is already of integer type and using ``np.take`` rather than fancy indexing a larger effect.

```python
n_verts = 1000000
vertices = hv.Points(np.random.randint(0, n_verts, (n_verts, 3)), vdims=['z'])
trimesh = hv.TriMesh.from_vertices(vertices)
vdf = vertices.dframe() 
simplices = trimesh.dframe()
print("Vertices: ", len(vdf))
print("Simplices: ", len(simplices))
```

    Vertices: 1000000
    Simplices: 1999965

```python
%%timeit
_pd_mesh(vdf, simplices, optimized=False)
```

    355 ms ± 10.3 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

```
%%timeit
_pd_mesh(vdf, simplices, optimized=True)
```

    237 ms ± 21.3 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

bokeh_plot

### DIFF
--- a/datashader/utils.py
+++ b/datashader/utils.py
@@ -383,7 +383,7 @@ def dataframe_from_multiple_sequences(x_values, y_values):
    # Return a dataframe with this new set of x and y values
    return pd.DataFrame({'x': x, 'y': y.flatten()})
 
-   
+
 def _pd_mesh(vertices, simplices):
     """Helper for ``datashader.utils.mesh()``. Both arguments are assumed to be
     Pandas DataFrame objects.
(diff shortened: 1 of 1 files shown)


jbednar:
Looks and sounds good to me, assuming the tests pass!

PR #556, opened 2018-01-24
New API for Canvas.raster

jbednar:
As mentioned in https://github.com/ioam/holoviews/issues/1909, the ``Canvas.raster()`` API does not match the rest of datashader, due to that code originating in an external project (gridtools). These differences make it difficult for external tools like HoloViews to provide a consistent interface for datashading across object types. Fully rewriting this code would be a lot of work, but this WIP PR represents a first pass at making the Canvas.raster() API more similar to the other Canvas glyph types.  The changes should all be backwards compatible for now, but the previous way of doing it will now be deprecated and removed in a later release.

API Changes:

- Renamed ``Canvas.raster(downsample_method=X)`` to ``Canvas.raster(agg=X)``: What gridtools calls "downsampling" is precisely the same concept as what datashader calls aggregation everywhere else.  "reduction" is perhaps even more accurate, but arguments like that are called ``agg`` elsewhere, so I've adopted that convention here as well.  ``downsample_method`` is still accepted as an alias for now, if ``agg`` is not present.
- Renamed ``Canvas.raster(upsample_method=X)`` to ``Canvas.raster(interpolate=X)``: "interpolate" seems like a better complement to "agg" than "upsample_method".  ``upsample_method`` is still accepted as an alias for now, if ``interpolate`` is not present.
- The ``agg`` argument of other calls accepts any object of type ``.reductions.Reduction``, but ``raster()`` accepted only string arguments.  It now accepts ``Reduction`` objects like ``rd.mean()``, extracting the "column" name, if any, and comparing it to the DataArray's name, if any (signaling an error if a "column" name is specified but doesn't match).  Of course, it's not really a "column", but it's the same idea.  Leaving off the explicit column is fine for a raster, since by definition an xarray DataArray can have only one data dimension, and so Reduction has been changed to allow instantiation without any argument.  String names are still accepted, for backwards compatibility, but would presumably eventually be removed.

Open issues:

- ``Canvas.raster`` supports three aggregations that are not available in ``.reductions``: ``mode``, ``first``, and ``last``.  All three are designed for use with categorical data where numerical averaging is not appropriate and an actual existing value must be returned.  For now, these can be accessed with strings, but that's confusing and inconsistent.  Options:
    1. Delete these three
    2. Keep using strings just for these three
    3. Add dummy reductions for these three raising NotImplementedError on use, just for API consistency
    4. Implement these three for the other Datashader operations.
- I think these three reductions are useful, so I don't want option i, but option ii is annoying.  Maybe we could do option iii with an eye to someday option iv?
- Relaxing ``Reduction`` to allow construction with no argument might make some user errors more obscure by failing to detect them early.  I added an explicit check for an argument when the reduction is used, and it seems to work ok, but we should be on the lookout for confusing errors of this type.

### DIFF
--- a/datashader/core.py
+++ b/datashader/core.py
@@ -10,23 +10,7 @@ from collections import OrderedDict
 from .utils import Dispatcher, ngjit, calc_res, calc_bbox, orient_array, compute_coords, get_indices, dshape_from_pandas, dshape_from_dask, categorical_in_dtypes
 from .resampling import (resample_2d, US_NEAREST, US_LINEAR, DS_FIRST, DS_LAST,
                          DS_MEAN, DS_MODE, DS_VAR, DS_STD, DS_MIN, DS_MAX)
-
-
-class Expr(object):
-    """Base class for expression-like objects.
-
-    Implements hashing and equality checks. Subclasses should implement an
-    ``inputs`` attribute/property, containing a tuple of everything that fully
-    defines that expression.
-    """
-    def __hash__(self):
-        return hash((type(self), self.inputs))
-
-    def __eq__(self, other):
-        return type(self) is type(other) and self.inputs == other.inputs
-
-    def __ne__(self, other):
-        return not self == other
+from .utils import Expr # noqa (API import)
 
 
 class Axis(object):
--- a/datashader/core.py
+++ b/datashader/core.py
@@ -187,7 +187,7 @@ class Canvas(object):
             agg = any_rdn()
         return bypixel(source, self, Line(x, y), agg)
 
-    def trimesh(self, vertices, simplices, mesh=None, agg=None, interp=True):
+    def trimesh(self, vertices, simplices, mesh=None, agg=None, interp=True, interpolate=None):
         """Compute a reduction by pixel, mapping data to pixels as a triangle.
 
         >>> import datashader as ds
@@ -226,10 +226,13 @@ class Canvas(object):
             purposes. This dataframe is expected to have come from
             ``datashader.utils.mesh()``. If this argument is not None, the first
             two arguments are ignored.
-        interp : boolean, optional
-            Specify whether to do bilinear interpolation of the pixels within each
-            triangle. This can be thought of as a "weighted average" of the vertex
-            values. Defaults to True.
+        interpolate : str, optional default=linear
+            Method to use for interpolation between specified values. ``nearest``
+            means to use a single value for the whole triangle, and ``linear``
+            means to do bilinear interpolation of the pixels within each
+            triangle (a weighted average of the vertex values). For 
+            backwards compatibility, also accepts ``interp=True`` for ``linear``
+            and ``interp=False`` for ``nearest``.
         """
         from .glyphs import Triangles
         from .reductions import mean as mean_rdn
@@ -237,6 +240,15 @@ class Canvas(object):
 
         source = mesh
 
+        # 'interp' argument is deprecated as of datashader=0.6.4
+        if interpolate is not None:
+            if interpolate == 'linear':
+                interp = True
+            elif interpolate == 'nearest':
+                interp = False
+            else:
+                raise ValueError('Invalid interpolate method: options include {}'.format(['linear','nearest']))
+            
         # Validation is done inside the [pd]d_mesh utility functions
         if source is None:
             source = create_mesh(vertices, simplices)
@@ -260,11 +272,11 @@ class Canvas(object):
     def raster(self,
                source,
                layer=None,
-               upsample_method='linear',    # Deprecated as of datashader=0.6.5
-               downsample_method=rd.mean(), # Deprecated as of datashader=0.6.5
+               upsample_method='linear',    # Deprecated as of datashader=0.6.4
+               downsample_method=rd.mean(), # Deprecated as of datashader=0.6.4
+               nan_value=None,
                agg=None,
-               interpolate=None,
-               nan_value=None):
+               interpolate=None):
         """Sample a raster dataset by canvas size and bounds.
 
         Handles 2D or 3D xarray DataArrays, assuming that the last two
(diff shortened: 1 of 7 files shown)


philippjfr:
Looks great, this looks ready to merge as is. One possibility we didn't discuss is that we could now allow passing in xarray Datasets and grab the DataArray that corresponds to the supplied "``column``". It's a small, straightforward change and can be convenient in some cases.

>Canvas.raster supports three aggregations that are not available in .reductions: mode, first, and last.

I'd say these can be useful, so I too lean toward option iii. with the aim of implementing them at a later point (i.e. option iv.)

jbednar:
> Looks great, this looks ready to merge as is. One possibility we didn't discuss is that we could now allow passing in xarray Datasets and grab the DataArray that corresponds to the supplied "column". It's a small, straightforward change and can be convenient in some cases.

That sounds great!  Could you add that to this PR?  I'll add the dummy reductions.

datashader/core.py
```diff
@@ -310,22 +298,34 @@ def raster(self,
...
+                              'max':DS_MAX,     rd.max:DS_MAX}
 
-        if upsample_method not in upsample_methods.keys():
+        if interpolate not in upsample_methods.keys():
             raise ValueError('Invalid upsample method: options include {}'.format(list(upsample_methods.keys())))
-        if downsample_method not in downsample_methods.keys():
+
+        if hasattr(agg,"column"):
+            # Accepts a reduction from ./reductions.py, but only uses the class name and the optional column name
+            if hasattr(source,"name") and agg.column != source.name:
```
philippjfr: Did we decide that if agg.column is ``None`` it should match any name on the DataArray? Since it's unambiguous that seems fine to me.

philippjfr:
Okay done, I've improved the validation a fair bit in the process. Feel free to change it as you desire.

datashader/core.py
```diff
@@ -317,16 +318,41 @@ def raster(self,
...
+                               'data variables: %r.' % (column, data_vars))
+            source = source[column]
+
         if agg not in downsample_methods.keys():
             raise ValueError('Invalid downsample method: options include {}'.format(list(downsample_methods.keys())))
         ds_method = downsample_methods[agg]
 
+        if source.ndim not in [2, 3]:
+            raise ValueError('Raster aggregation expects a 2D or 3D '
+                             'DataArray, found %s dimensions' % source.ndim)
```
jbednar: Should this say "expects a 2D DataArray or a 3D Dataset"?  How would we handle a 3D DataArray?

philippjfr: No, this is correct. The operation handles arrays with multiple layers as long as the first dimension defines the layer, e.g. to resample all the layers in the landsat data at once or a time dimension. It's specifically to handle whatever``xr.open_rasterio`` but can be used in other cases too.

jbednar:
Your changes look great, @philippjfr.  I've added stubs for first, last, and mode, but am having second thoughts about first and last.  Now that those operations are parallel, there isn't really any sense in which they provide the first and last item encountered; they just provide *an* item, which will differ depending on the number of cores used.  The ``mode`` is well defined and has similar properties but is more expensive to compute, and I'm pretty sure the only reason to use ``first`` or ``last`` is as a cheaper approximation to ``mode`` (with someone adding ``last`` just to match having added ``first).  So, I propose that we remove ``last`` and rename ``first``, e.g. to ``one`` or ``an`` or ``arb`` (arbitrary) or ``ex`` (example).  Does that (a) seem reasonable and (b) any preference on the name?

philippjfr:
>Now that those operations are parallel, there isn't really any sense in which they provide the first and last item encountered; they just provide an item, which will differ depending on the number of cores used. 

Is that true? The parallelism chunks over the output array not the input so you're still getting the first value for each bin in the output array, so unless I'm confused it's both well defined and correct.

datashader/core.py
jbednar: Ok.  It would be nice if the message mentioned Dataset to make it clear that's an acceptable way to pass the DataArray, but I can't think of any way to word that that makes it less rather than more confusing, so I guess it's fine as is.

jbednar:
@philippjfr, another API issue: All the ``agg`` arguments now consistently accept reductions, but upsampling in ``Canvas.raster`` accepts a string argument ``interpolate={'nearest','linear'}``
while upsampling in ``Canvas.trimesh`` accepts a Boolean argument ``interp={True,False}``, where ``interp==False`` is the same as ``interpolate='nearest'``, and ``interp==True`` is the same as ``interpolate='linear'``.  In both cases we can imagine future implementations that support bicubic or other types of interpolation.  Seems like we should standardize those interfaces as well; should we use strings or Booleans or should we make a new file ``datashader.interpolate.py`` with a couple of stub interpolation operators to allow future implementation like reductions?

philippjfr:
I think standardizing both to strings is fine for now and we could add the proper operators later. Up to you though, booleans definitely don't seem sensible, although perhaps ``True`` could be equivalent to 'linear' on trimesh for now.

jbednar:
@philippjfr, I've standardized the API on strings, but kept compatibility with the old approach for now because HoloViews is using it (even though Trimesh is not released).  Once this is merged and HV master is updated to use the new API then we can remove the "interp" option on Trimesh, though the backwards compatibility bits should stay in for Rasters as those have been released. Can you please review the previous commit to spot any problems?

datashader/core.py
```diff
@@ -226,17 +226,29 @@ def trimesh(self, vertices, simplices, mesh=None, agg=None, interp=True):
...
         source = mesh
 
+        # 'interp' argument is deprecated as of datashader=0.6.4
+        if interpolate is not None:
+            if interpolate == 'linear':
+                interp = True
+            elif interpolate == 'nearest':
+                interp = False
+            else:
+                raise ValueError('Invalid interpolate method: options include {}'.format(['linear','nearest']))
```
philippjfr: Looks fine, if we weren't deprecating this right away I'd say it should error if interp and interpolation are incompatible.

jbednar: Right; hopefully this will quickly disappear.

jbednar:
Ok, I'm ready to merge this once the tests pass.

jbednar:
One last change: Changed the ``layer`` argument of ``Canvas.raster`` to be an xarray coordinate, so that if the DataArray was declared with floating-point coordinate values, it will select the right layer using that value.  Previously, layer was a 1-based integer index, which was confusing in Python as everything else is 0-based.  If the DataArray has been declared with 1-based coordinates the behavior should not change, but now it will handle other coordinates as well.

PR #572, opened 2018-03-13
Allow trimesh()'s doctest to pass, while indicating that the output is untested.

ceball:
Notes:
  * by "untested", I mean the values returned are untested - any exception would show up, though
  * tests here do not cover this issue, so no need to wait on the tests
  * there's probably a doctest way of ignoring output (something like https://docs.python.org/3/library/doctest.html#doctest.ELLIPSIS ?)

### DIFF
--- a/datashader/core.py
+++ b/datashader/core.py
@@ -198,7 +198,7 @@ class Canvas(object):
         ...                       columns=['v0', 'v1', 'v2'])
         >>> cvs = ds.Canvas(x_range=(verts.x.min(), verts.x.max()),
         ...                 y_range=(verts.y.min(), verts.y.max()))
-        >>> cvs.trimesh(verts, tris)
+        >>> untested = cvs.trimesh(verts, tris)
 
         Parameters
         ----------


ceball:
Just in case it might matter in terms of the values being returned, that link was to a pip based build, but it's the same with a conda one (https://travis-ci.org/bokeh/datashader/jobs/352297621#L1065):

screen shot 2018-03-13 at 1 47 31 pm

jbednar:
Oh, it's not that it confuses doctests, it's that it's a confused doctest, i.e. that we didn't include the output in the doctest.  No problem, though in that case the word "untested" is probably not the best variable name; it's just that we want to ignore the output (since pasting that long repr into the docstring would be more confusing than helpful).

ceball:
Yes, I guess `output_ignored` or something would have been better. Or doctest's own way of ignoring output, whatever exactly that is. At some point I'll try out a change on an existing, unrelated PR where the doctests are being run.

ceball:
@jbednar, I couldn't get just `...` on its own to work, but you can do this kind of thing:

```python
        >>> cvs.trimesh(verts, tris)
        <xarray.DataArray (y: 600, x: 600)>
        array([[...]])
        Coordinates:
          * y ...
          * x ...
```

I.e. checking the type/structure of what comes back (kind of, I think).

If you tell me what you prefer, I can make a change if necessary in the PR where this doctest is running.

jbednar:
That seems like it's mainly testing that xarray's repr doesn't change; probably simpler just to ignore the output here as was already merged.

jbednar:
Or if you want to make it a real test, just sum or aggregate the results and show *that* instead; that's something reasonable for a doctest.

ceball:
It's got something about the shape, type, and labels in there. But yes, I'm sure it would be better just to improve the test.

jbednar:
Actually, in this case I'd probably just make it be a something tiny like a 3x3 array in the call to Canvas  (plot_height=3,plot_width=3), then show the output as a numpy array that has a compact representation.

PR #573, opened 2018-03-15
Handle Canvas.raster subpixel condition

philippjfr:
When Canvas.raster is requested to downsample a plot to subpixel range all kinds of errors can occur due to negative dimensions and inverted ranges, additionally the padding code can get confused due to zero sized padding arrays. This ensures that the src_window always has a non-zero shape and that the padding is robust to issues like these.

- [x] Add a number of unit tests

### DIFF
--- a/datashader/core.py
+++ b/datashader/core.py
@@ -395,8 +395,8 @@ class Canvas(object):
         if np.isclose(width_ratio, 0) or np.isclose(height_ratio, 0):
             raise ValueError('Canvas x_range or y_range values do not match closely enough with the data source to be able to accurately rasterize. Please provide ranges that are more accurate.')
 
-        w = int(np.ceil(self.plot_width * width_ratio))
-        h = int(np.ceil(self.plot_height * height_ratio))
+        w = max(int(np.ceil(self.plot_width * width_ratio)), 1)
+        h = max(int(np.ceil(self.plot_height * height_ratio)), 1)
         cmin, cmax = get_indices(xmin, xmax, xvals, res[0])
         rmin, rmax = get_indices(ymin, ymax, yvals, res[1])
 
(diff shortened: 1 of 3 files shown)


jbednar:
I'm a bit confused by why these changes are necessary.  In all cases where I have investigated, the lines being changed here should never be reached, because a negative or zero plot_width or plot_height is being requested.  Are there cases where this code will be activated for some valid plot_width or plot_height?  Perhaps some case where the overall plot widths and heights are fine, but the plot ends up consisting entirely of padding?

In any case I think we should be guarding against plot_width and plot_height ever being set to a value that's less than 1, which is trivial if we make those into Parameters, but meanwhile could be an assertion.

philippjfr:
If the x_range or y_range defines a region that is smaller than 1 pixel three kinds of issue occur:

1. The integer slices into the data array invert due to half pixel offsets
2. The shape allocated to resampled region is 0
3. The shape allocated to the padding region is 0

This is independent of any plot_width or plot_height issues

jbednar:
It sounds to me like you've uncovered issues semantically related to the ones we've been encountering today, but strictly independent of them.  Sometimes it is very helpful to give a vague bug report to someone, so that they discover the problems nearby those that were discovered. :-)

jbednar:
Sounds like a good debugging strategy.  But when the requested plot_width or plot_height is 0 or negative, there is no image to render at all, so no datashader code (other than perhaps an assertion) should ever be executed.  The fixes listed here would address *different* (but equally bad) conditions that arise when the supplied plot_width and plot_height are valid, yet the rendered array would still have been calculated to be smaller than one pixel in x or y because of padding.

philippjfr:
>But when the requested plot_width or plot_height is 0 or negative, there is no image to render at all, so no datashader code (other than perhaps an assertion) should ever be executed.

That's a sensible thing to do in future, but I'm not aware of any code in datashader or holoviews that would have stopped datashader code from executing in this case.

jbednar:
That's right.  HoloViews shouldn't be requesting a zero or negative size, and Datshader should complain when it is asked for one.

philippjfr:
Added some unit tests and simplified the changes. The validation of input parameters can be followed up on as part of https://github.com/bokeh/datashader/issues/574.

philippjfr:
Tests passing, merging.

PR #578, opened 2018-03-20
Clip width and height ratios to avoid numerical precision issues

philippjfr:
The width and height ratios are used to compute the size of the resampled region relative to the requested width and height, therefore they should never exceed 1. In certain cases numerical precision causes them to be ever so slightly above 1, causing issues when the data is not the expected size when constructing the xarray.DataArray. This PR simply clips the ratios at 1.

### DIFF
--- a/datashader/core.py
+++ b/datashader/core.py
@@ -389,8 +389,8 @@ class Canvas(object):
         xmax = min(self.x_range[1], right)
         ymax = min(self.y_range[1], top)
 
-        width_ratio = (xmax - xmin) / (self.x_range[1] - self.x_range[0])
-        height_ratio = (ymax - ymin) / (self.y_range[1] - self.y_range[0])
+        width_ratio = min((xmax - xmin) / (self.x_range[1] - self.x_range[0]), 1)
+        height_ratio = min((ymax - ymin) / (self.y_range[1] - self.y_range[0]), 1)
 
         if np.isclose(width_ratio, 0) or np.isclose(height_ratio, 0):
             raise ValueError('Canvas x_range or y_range values do not match closely enough with the data source to be able to accurately rasterize. Please provide ranges that are more accurate.')


jbednar:
Is there any danger this change will mask erroneous cases where the ratio is greatly higher than 1? If so, is there a way we could catch that and raise an exception?  If not, looks fine.

philippjfr:
>Is there any danger this change will mask erroneous cases where the ratio is greatly higher than 1?

No, if we had infinite precision the ratio would never be more than one since the ranges are also clipped to the maximum extent. The slight numerical error later gets amplified by an np.ceil.

jbednar:
Ok, sounds good.  Will merge when tests pass.

jbednar:
For posterity, can you please paste the error messages that appeared when the ratio was above 1?

philippjfr:
Sure, here it is:

<details>
	  File "/Users/philippjfr/topographica/external/holoviews/holoviews/plotting/util.py", line 258, in get_plot_frame
	    return map_obj[key]
	  File "/Users/philippjfr/topographica/external/holoviews/holoviews/core/spaces.py", line 1119, in __getitem__
	    val = self._execute_callback(*tuple_key)
	  File "/Users/philippjfr/topographica/external/holoviews/holoviews/core/spaces.py", line 905, in _execute_callback
	    retval = self.callback(*args, **kwargs)
	  File "/Users/philippjfr/topographica/external/holoviews/holoviews/core/spaces.py", line 571, in __call__
	    ret = self.callable(*args, **kwargs)
	  File "/Users/philippjfr/topographica/external/holoviews/holoviews/util/__init__.py", line 431, in dynamic_operation
	    return self._process(obj, key)
	  File "/Users/philippjfr/topographica/external/holoviews/holoviews/util/__init__.py", line 417, in _process
	    return self.p.operation.process_element(element, key, **kwargs)
	  File "/Users/philippjfr/topographica/external/holoviews/holoviews/core/operation.py", line 143, in process_element
	    return self._apply(element, key)
	  File "/Users/philippjfr/topographica/external/holoviews/holoviews/core/operation.py", line 121, in _apply
	    ret = self._process(element, key)
	  File "/Users/philippjfr/topographica/external/holoviews/holoviews/operation/datashader.py", line 516, in _process
	    downsample_method=self.p.aggregator)
	  File "/Users/philippjfr/datashader/datashader/core.py", line 480, in raster
	    return DataArray(data, coords=coords, dims=dims, attrs=attrs)
	  File "/Users/philippjfr/development/xarray/xarray/core/dataarray.py", line 234, in __init__
	    coords, dims = _infer_coords_and_dims(data.shape, coords, dims)
	  File "/Users/philippjfr/development/xarray/xarray/core/dataarray.py", line 89, in _infer_coords_and_dims
	    'coordinate %r' % (d, sizes[d], s, k))
</details>
	ValueError: conflicting sizes for dimension 'time': length 2 on the data but length 1 on coordinate 'time'

PR #584, opened 2018-04-10
Converted osm-1billion example to use HoloViews

jbednar:
Should have the same output as before, but without all the code required for InteractiveImage.
See output at: https://anaconda.org/jbednar


jbednar:
Any suggestions on HV style or simplifications appreciated.  Otherwise ready for merge.

PR #587, opened 2018-04-15
Enabling PySpark as a data source

patrick-nicholson:
`datashader` enables meaningful visualization of big data, but these data must be made available through `dask`. Spark is much more widely deployed and, depending on setup and circumstance, it's not always possible to deploy `dask`. 

This PR implements a `pyspark.sql.dataframe` as a data source for `datashader`. It works by bootstrapping the core `pandas` implementations on serialized `pyspark.rdd` partitions. There is a large serialization penalty paid, but future versions of `pyspark` will have improved support for data transfer from the JVM to Python using Apache Arrow.* The downside to this approach is that it's more complicated to deploy on a Spark cluster (e.g., using YARN) than a `pyspark.sql`-based approach would be.

I don't have tests included yet, and I have included `pyspark` as a dependency. My reading of the code suggests that both `dask` and `pyspark` should be installable as extras, rather than `datashader` core dependencies. I wanted to thoughts on the concept and eyes on the implementation early, in case large changes are necessary.

\* There is some Arrow support in the current PySpark release (2.3). However, the two exposed methods (1) don't allow aggregation and (2) require a full dataframe shuffle prior to aggregation. I tried to hack (1) to work by using a `pyspark.Accumulator`, but it's not available to the Python process running the Arrow UDF. (2) can be used, but I feel that the required shuffle is a larger performance penalty than RDD serialization. 

**Additional** -- I *think* I used the intended approach to extend `datashader` in this PR. This required bootstrapping existing code onto Spark using a non-preferred interface with lower performance characteristics. And this is only possible with something like Spark that supports Python computation pretty well.

There is a lot of data in the world in various storage systems. Data scientists often access these data through Python clients or APIs, but the amount of data they can bring back is limited by client system size.  I'd like to gauge interest in refactoring `datashader` a bit to allow a more extensible API so that "code can be taken to the data." For example, a better Spark implementation would be to use the SQL API to perform mapping and aggregation. This would improve performance and be much simpler to deploy. You could similarly envision an RDBMS implementation using some Python API (e.g., `blaze`) or client.


datashader/core.py
```diff
@@ -246,8 +260,9 @@ def trimesh(self, vertices, simplices, mesh=None, agg=None, interp=True, interpo
             elif interpolate == 'nearest':
                 interp = False
             else:
-                raise ValueError('Invalid interpolate method: options include {}'.format(['linear','nearest']))
-            
+                raise ValueError('Invalid interpolate method: options include {}'
+                                 .format(['linear', 'nearest']))
+
```
jbednar: It would be very helpful to revert whitespace and source-code formatting changes like these (here and in other cases below), because they make it much more difficult to see the substantive changes being proposed.  It's fine to change whitespace if it's not well formatted, but it should be in a separate PR that we can verify isn't changing anything.

datashader/core.py
```diff
@@ -404,7 +424,7 @@ def raster(self,
                       us_method=interpolate, fill_value=fill_value)
         if array.ndim == 2:
             source_window = array[rmin:rmax+1, cmin:cmax+1]
-            if isinstance(source_window, Array):
+            if has_dask and isinstance(source_window, Array):
```
jbednar: Seems like this change could be avoided by defining Array and DaskDataFrame as new empty/stub classes if Dask isn't available; that way there would be no need to check for it here or below.

datashader/dask.py
```diff
@@ -71,7 +71,7 @@ def chunk(df):
     dask_dataframe_keys = getattr(df, '__dask_keys__', None) or df._keys
     keys = dask_dataframe_keys()
     keys2 = [(name, i) for i in range(len(keys))]
-    dsk = dict((k2, (chunk, k)) for (k2, k) in zip(keys2, keys))
+    dsk = {k2: (chunk, k) for k2, k in zip(keys2, keys)}
```
jbednar: This seems like a stylistic change, in which case please revert it, but if it's needed e.g. for python2 then it's fine.

requirements.txt
```diff
@@ -1,10 +1,14 @@
...
 pandas >= 0.20.3
+param >=1.5.0,<2.0
 pillow
+scikit-image
 toolz
 xarray >= 0.9.6
-colorcet
-param >=1.5.0,<2.0
+
+pyspark>=2.0
```
jbednar: I don't want to require users to install pyspark, so this would somehow have to be an optional dependency.

datashader/utils.py
```diff
@@ -469,10 +479,127 @@ def mesh(vertices, simplices):
...
+        return np.object
+    if isinstance(data_type, T.ShortType):
+        return np.int16
+    if isinstance(data_type, T.StructType):
+        return np.object
+    if isinstance(data_type, T.StringType):
+        return np.object
+    if isinstance(data_type, T.TimestampType):
+        return np.datetime64
+    raise TypeError("Unrecognized PySpark type %r" % data_type)
```
jbednar: I'd prefer if this were a dictionary plus a line or two of code, not a bunch of ifs.

datashader/utils.py
```diff
@@ -469,10 +479,127 @@ def mesh(vertices, simplices):
...
+    Create an empty, typed dataframe from a pandas schema
+    """
+    return pd.DataFrame({name: np.empty(0, dtype=dtype) for name, dtype in schema.items()})
+
+
+def serialized_rows_to_pandas(pandas_schema, nullsafe_schema, rows):
+    """
+    Read a serialized collection of rows (an iterator of pyspark.sql.Row objects) as a pandas
+    dataframe
+    """
```
jbednar: The docstring says what it does, but not why anyone would want or need this.  Plus the name should clarify that it's about spark only; it sounds like it's about the Pandas backend.

datashader/core.py
jbednar: (And similarly for PySpark dataframes).

datashader/utils.py
```diff
@@ -469,10 +479,127 @@ def mesh(vertices, simplices):
...
+        'If no vertex weight column is provided, a triangle weight column is required.'
 
-    if isinstance(vertices, dd.DataFrame) and isinstance(simplices, dd.DataFrame):
+    if has_dask and isinstance(vertices, DaskDataFrame) and isinstance(simplices, DaskDataFrame):
         return _dd_mesh(vertices, simplices)
 
     return _pd_mesh(vertices, simplices)
+
+
+def pyspark_type_to_pandas_type(data_type):
```
jbednar: Is there a reason these functions are in utils.py, not pyspark.py?  Seems like it would be easier to maintain if all spark code were in one place, so people can see what needs changing.

jbednar:
> * There is some Arrow support in the current PySpark release (2.3). However, the two exposed methods (1) don't allow aggregation and (2) require a full dataframe shuffle prior to aggregation. I tried to hack (1) to work by using a pyspark.Accumulator, but it's not available to the Python process running the Arrow UDF. (2) can be used, but I feel that the required shuffle is a larger performance penalty than RDD serialization.

Arrow is under very active development, so if you have suggestions for how to make it support what you need to do, please make sure they hear them!

patrick-nicholson:
Thanks for the review James. I'll be working on these over the course of the week.

datashader/core.py
patrick-nicholson: Reverting formatting changes for the purpose of this PR.

datashader/core.py
patrick-nicholson: I'm reverting this change for `dask`. I have implemented the suggested change for `pyspark`.

datashader/utils.py
patrick-nicholson: I thought it could go either way, but to isolate changes, I've moved everything into `pyspark.py`.

datashader/reductions.py
```diff
@@ -314,6 +327,59 @@ def finalize(bases, **kwargs):
...
+    def _combine(aggs):
+        return aggs.sum(axis=0, dtype='i4')
+
+    def _build_finalize(self, dshape):
+        cats = list(self.dtype.categories)
+        def finalize(bases, **kwargs):
+            dims = kwargs['dims'] + [self.column]
+            coords = kwargs['coords'] + [cats]
+            return xr.DataArray(bases[0], dims=dims, coords=coords)
+        return finalize
```
jbednar: This all seems ok, but as noted in https://github.com/bokeh/datashader/issues/140 , count_cat is unnecessarily limited to "count"; it should instead be applying any of the other reductions, but per category.  We've never implemented that functionality, but it would be a shame to make a special-purpose spark-specific version of count_cat when it could just as easily be solving the general case.  If you want to do that, there would need to be a separate PR to generalize count_cat for the existing backends, when I could merge and then here there could be a spark-specific version.

patrick-nicholson: This is an anti-pattern to work around the validation in `count_cat`. PySpark doesn't have a categorical type, so I had to do something to expose creation of a categorical column. Given the ability to cast to categorical, the implementation is the same as Pandas.

patrick-nicholson:
@jbednar I think I have taken this as far as I can inside the current API. All PySpark tests are passing. I can't figure out why some of the `trimesh` tests are failing, I didn't make any changes. Let me know what else you need for this PR.

patrick-nicholson:
Uhhh I think I just realized there's a much cleaner way to do this. Closing for now.

patrick-nicholson:
Nevermind, it's late and I remembered why I had to go this way. Sorry for the spam.

patrick-nicholson:
Just picking this back up after a busy month. I added a usage notebook at `examples/topics/pyspark-notes.ipynb`. If you read through it, you'll see the big performance gap between the Spark implementation in this PR and Dask. 

This code works, and I've tested it on a large cluster with 25 billion data points, but I'm not sure if it makes sense to merge this given the complications with deployment and obvious performance downsides. I don't know how to implement the orders of magnitude faster Spark-based computations without significant changes to the existing `datashader` codebase.

jbednar:
We'd be very happy to have something showing how to work with PySpark, but closing this for now until something's ready to be mergeable that would be more useful.

Issue #592, opened 2018-04-29
No data within min-max range resulting in black PNG


tropicalpb:
When there are no applicable data (e.g. after query conditions) within the min-max range, instead of a clear png, a black png is produced when running on linux.  I tested on a few Linux distro and the results are same.  To reproduce, here is the relevant code snippet

cvs = ds.Canvas(plot_width=100,
                    plot_height=100,
                    x_range=(100, 200),
                    y_range=(100, 200))

    # intentionally create a frame with data outside of range
    row = pd.DataFrame({'x': [5], 'y': [5] })
    row = dd.from_pandas(row, npartitions=1)
    pts = cvs.points(row, 'x', 'y', ds.mean('x'));
    img = tf.shade(pts, cmap=['red'])
    img = tf.spread(img, px=1);

tropicalpb:
I put in a workaround

    img = img.to_pil();
    pixdata = img.load()
    newdata = []
    iw, ih = img.size
    for y in range(ih):
        for x in range(iw):
            if pixdata[x, y][0] == 0 and pixdata[x, y][1] == 0 and pixdata[x, y][2] == 0:
                newdata.append((0, 0, 0, 0))
            else:
                newdata.append(pixdata[x, y])

    img.putdata(newdata)
    img_io = BytesIO()
    img.save(img_io, 'png')

jbednar:
Thanks for the bug report!  I suspect it would be this line from datashader/transfer_functions.py:

```
def _interpolate(agg, cmap, how, alpha, span, min_alpha, name):
    ...
        if len(masked) == 0:
            return Image(agg.data.astype(np.uint32), coords=agg.coords, dims=agg.dims, attrs=agg.attrs, name=name)
```

Here, it's constructing a special Image to return just in the case where there is no unmasked data, to avoid later divide-by-zero errors.  As you are already set up to have a specific use case, can you try changing this to return an all-clear image and verify that it works?  If so and you make a PR, I'd be happy to merge it.

tropicalpb:
Yes, it is this special image line.  Fixing seems to be a tricky as Image is a DataArray to PIL converter class.  One way is to pass in a special mask flag, or within to_pil() do a bitwise_and checking for all zeros.  Something like:

def to_pil
.....
     # bitwise alpha and check for all zeros
     if np.bitwise_and(arr, np.full(shape=arr.shape, fill_value=int('00111111', 16), dtype=np.uint32)).any():
           return fromarray(arr, 'RGBA')
[...]

PR #593, opened 2018-05-06
fix for issue 592 black png

tropicalpb:
fix for a special Image to return just in the case where there is no unmasked data returning black png


jbednar:
Thanks for the contribution!

jbednar:
Fixes #592.

jbednar:
Fixes #592

PR #600, opened 2018-05-22
Add build stage for pip release to travis

ceball:


ceball:
I don't see any reason why this shouldn't be merged. Once merged, I'll make the pypi upload happen. (In future, it will happen when release tag is pushed.)

PR #597, opened 2018-05-17
Add build stage for conda release to .travis.yml

philippjfr:
Adds a build stage to travis to release to conda on tags, pip not yet set up due to missing credentials.

PR #601, opened 2018-05-22
For conda, test against pyviz channel packages for releases 

ceball:
(rather than packages on pyviz/label/dev; fixup to #597)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,7 @@ env:
   global:
     - PYENV_VERSION=3.6
     - CHANS_DEV="-c pyviz/label/dev -c conda-forge"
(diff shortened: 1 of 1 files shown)


ceball:
This is about pyviz packages getting their pyviz dependencies from pyviz/label/dev during development, but pyviz/label/main for releases.

(In future, I might make it so this happens automatically, i.e. so pyviz projects don't need to specify separate 'dev' and 'release' channel lists unless they need something different from the default.) 

The next step after this is to remove conda-forge from the list of channels (after moving topics examples to pyviz), in line with the idea that we'll test pyviz projects on top of defaults, and separately use c-f CI to test on top of c-f (https://github.com/pyviz/pyct/issues/25).

Issue #594, opened 2018-05-08
Need to complete remaining items from #564


jbednar:
Copied from #564:

 * [ ] pillow was pinned to 4.3.0 on appveyor; need to test ds pil usage on windows. (where did problem show up - is it covered by unit tests + running the limited set of examples, which is what happens on appveyor?) 
  * [ ] now we dropped the examples package, should move topics examples and dependencies to pyviz (maybe future PR? jbednar: yes and yes)
  * [ ] now we dropped the examples package, how do users know to get holoviews - via docs? (holoviews is imported in user guide notebooks , but isn't a required dependency.) (jbednar: mention in install instructions, as long as it's just a couple of packages to install)
  * [ ] how does taxi_prepreprocessing_example relate to download_sample_data.py? And was it intended for running as a script (because that's what it is now...) (it's an example, and isn't needed for any other purpose as the output of it is what we are providing for download)
  * [ ] To do for CB:  Travis file modifications before merge, once testing on this branch is done.

PR #602, opened 2018-05-22
Removed out of date requirements.txt

ceball:
(Part of #594)

The dependencies are now only specified in setup.py.

In the future, we could have a discussion about where it's best to specify dependencies, but we decided to start by having the generic/abstract dependencies in one place only in a form that can be used by pip and conda. Apart from using the setup.py dependencies for pip and conda, they can also be used to generate other types of dependency specifications e.g. an explicit examples environment (for pip or conda ecosystems).


ceball:
The appveyor branch failure can be ignored; something went wrong with the particular machine while installing the environment for one python version (the other python version succeeded, and both succeeded for the pr build).

PR #631, opened 2018-08-08
Silenced inappropriate warnings

jbednar:
In Datashader, NaN is used to indicate "no data" in a pixel, and so it is common for there to be aggregations that are all NaN, indicating that there is no data in a given chunk of the output raster or the entire raster.  By default, Numpy warns when an "all-NaN slice" is encountered, because in many cases people use Numpy's NaN-aware functions (e.g. `np.nanmin`) when they want a numerical value as a result, rather than NaN.  Here, we want the minimum (or other aggregation) of any bin to be NaN if there was no data in that bin, correctly reporting the lack of data rather than indicating any problem with the numerical calculations.  Thus it's appropriate to turn off Numpy's NaN warnings as done here.

Note that it's possible that they should be turned off at a different *location* in the code. In this PR, the warnings are disabled when calling `bypixel`, which applies to all of the pipeline, as I didn't find a good spot to turn them off just for aggregations.  I suspect that there is nowhere such a warning would be relevant for Datashader, though, hence disabling the warning at such a global level.

This PR also disables warnings for division by zero and invalid values in the `std` and `var` reductions specifically, which appear to occur for empty bins.  The default behavior of Numpy (apart from the warnings) appears to be fine in those cases, so it seems appropriate to disable the "divide by zero encountered" and "invalid value encountered" warnings here as well.

Also added a few recently added reduction operators to the main namespace, for API consistency.

### DIFF
--- a/datashader/__init__.py
+++ b/datashader/__init__.py
@@ -4,8 +4,7 @@ import param
 __version__ = str(param.version.Version(fpath=__file__, archive_commit="$Format:%h$",reponame="datashader"))
 
 from .core import Canvas                                 # noqa (API import)
-from .reductions import (count, any, sum, min, max,      # noqa (API import)
-                         mean, std, var, count_cat, summary)
+from .reductions import *                                # noqa (API import)
 from .glyphs import Point                                # noqa (API import)
 from .pipeline import Pipeline                           # noqa (API import)
 from . import transfer_functions as tf                   # noqa (API import)
--- a/datashader/core.py
+++ b/datashader/core.py
@@ -535,7 +535,11 @@ def bypixel(source, canvas, glyph, agg):
     glyph.validate(schema)
     agg.validate(schema)
     canvas.validate()
-    return bypixel.pipeline(source, schema, canvas, glyph, agg)
+
+    # All-NaN objects (e.g. chunks of arrays with no data) are valid in Datashader
+    with np.warnings.catch_warnings():
+        np.warnings.filterwarnings('ignore', r'All-NaN (slice|axis) encountered')
+        return bypixel.pipeline(source, schema, canvas, glyph, agg)
 
 
 bypixel.pipeline = Dispatcher()
--- a/datashader/reductions.py
+++ b/datashader/reductions.py
@@ -221,8 +221,9 @@ class m2(FloatingReduction):
 
     @staticmethod
     def _combine(Ms, sums, ns):
-        mu = np.nansum(sums, axis=0) / ns.sum(axis=0)
-        return np.nansum(Ms + ns*(sums/ns - mu)**2, axis=0)
+        with np.errstate(divide='ignore', invalid='ignore'):
+            mu = np.nansum(sums, axis=0) / ns.sum(axis=0)
+            return np.nansum(Ms + ns*(sums/ns - mu)**2, axis=0)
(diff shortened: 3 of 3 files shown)


philippjfr:
Looks fine to me. Not sure what's up with the tests.

jbednar:
@ceball, it looks like holoviews and other libraries needed for running the .ipynb files aren't installed during the tests, even though nbsmoke is running on them.  Did the testing configuration get mixed up somehow?

ceball:
I think you're referring to warnings relating to the notebooks. Those warnings are unfortunately always present at the moment: they say that datashader doesn't declare dependencies on streamz, holoviews, geoviews, etc, yet the notebooks use those. There's only one datashader conda package, so a user needs to install those dependencies manually to be able to run the examples. 

I have a task already to update the documentation and clean up the dependency declarations so that even though there isn't a conda package which includes those dependencies, tools like pyctdev can know to install them.

Separately, I have another task to improve the output of nbsmoke: shortening tracebacks (done but not pushed), getting color/better formatting on travis, and trying to make sure there's a final summary which everyone knows they can look at understand what happened (rather than having to figure it out from reams of output).

Anyway, the failures are like this (https://travis-ci.org/bokeh/datashader/jobs/413814584#L1822):

```
=================================== FAILURES ===================================
___________________ [doctest] datashader.reductions.summary ____________________
495     Computes all aggregates simultaneously, output is stored as a
496     ``xarray.Dataset``.
497 
498     Examples
499     --------
500     A reduction for computing the mean of column "a", and the sum of column "b"
501     for each bin, all in a single pass.
502 
503     >>> import datashader as ds
504     >>> red = ds.summary(mean_a=ds.mean('a'), sum_b=ds.sum('b'))
UNEXPECTED EXCEPTION: AttributeError("module 'datashader' has no attribute 'summary'",)
  File "/home/travis/miniconda/envs/test-environment/lib/python3.6/doctest.py", line 1330, in __run
    compileflags, 1), test.globs)
  File "<doctest datashader.reductions.summary[1]>", line 1, in <module>
AttributeError: module 'datashader' has no attribute 'summary'
/home/travis/build/bokeh/datashader/datashader/reductions.py:504: UnexpectedException
___________________________ test_multiple_aggregates ___________________________
    def test_multiple_aggregates():
        agg = c.points(ddf, 'x', 'y',
>                      ds.summary(f64_std=ds.std('f64'),
                                  f64_mean=ds.mean('f64'),
                                  i32_sum=ds.sum('i32'),
                                  i32_count=ds.count('i32')))
E       AttributeError: module 'datashader' has no attribute 'summary'
datashader/tests/test_dask.py:153: AttributeError
___________________________ test_multiple_aggregates ___________________________
    def test_multiple_aggregates():
        agg = c.points(df, 'x', 'y',
>                      ds.summary(f64_std=ds.std('f64'),
                                  f64_mean=ds.mean('f64'),
                                  i32_sum=ds.sum('i32'),
                                  i32_count=ds.count('i32')))
E       AttributeError: module 'datashader' has no attribute 'summary'
datashader/tests/test_pandas.py:147: AttributeError
```

I've only glanced at it, but I can imagine your changes causing that. You could change the doctest or restore the original behavior (no comment on which to do, I'm just skimming).

PR #634, opened 2018-08-14
Optimized dask datashape detection

philippjfr:
Leads to a 2-4x speedup in categorical aggregations and probably even larger improvements for out-of-core aggregation.

- [x] Fixes https://github.com/bokeh/datashader/issues/633

### DIFF
--- a/datashader/core.py
+++ b/datashader/core.py
@@ -505,10 +505,10 @@ def bypixel(source, canvas, glyph, agg):
     glyph : Glyph
     agg : Reduction
     """
-    # Avoid datashape.Categorical instantiation bottleneck
-    # by only retaining the necessary columns:
-    # https://github.com/bokeh/datashader/issues/396
-    if categorical_in_dtypes(source.dtypes.values):
+    if isinstance(source, pd.DataFrame):
+        # Avoid datashape.Categorical instantiation bottleneck
+        # by only retaining the necessary columns:
+        # https://github.com/bokeh/datashader/issues/396
         # Preserve column ordering without duplicates
         cols_to_keep = OrderedDict({col: False for col in source.columns})
         cols_to_keep[glyph.x] = True
@@ -521,14 +521,12 @@ def bypixel(source, canvas, glyph, agg):
                     cols_to_keep[subagg.column] = True
         elif agg.column is not None:
             cols_to_keep[agg.column] = True
-        src = source[[col for col, keepit in cols_to_keep.items() if keepit]]
-    else:
-        src = source
-
-    if isinstance(src, pd.DataFrame):
-        dshape = dshape_from_pandas(src)
-    elif isinstance(src, dd.DataFrame):
-        dshape = dshape_from_dask(src)
+        cols_to_keep = [col for col, keepit in cols_to_keep.items() if keepit]
+        if len(cols_to_keep) < len(source.columns):
+            source = source[cols_to_keep]
+        dshape = dshape_from_pandas(source)
+    elif isinstance(source, dd.DataFrame):
+        dshape = dshape_from_dask(source)
     else:
         raise ValueError("source must be a pandas or dask DataFrame")
     schema = dshape.measure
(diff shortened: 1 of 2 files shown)


philippjfr:
This change does not work for categorical data after all.

jcrist:
This might help explain whats going on: http://dask.pydata.org/en/latest/dataframe-design.html#categoricals

Dask uses an empty pandas dataframe to store dtypes/metadata. For categoricals, we sometimes also store the categories. This is a distinction between "known" and "unknown" categories (see the link above).

In this case you know it's a categorical column, but you don't know what the categories are. The previous behavior was incorrect, because you can't assume that the categories in partition 1 are representative of the categories in all partitions for "unknown" categories.

Instead, when a categorical column is desired, for dask you should do something like:

```
c = 'categorical_column_name'
# Convert unknown categorical column to known categories
if not df[c].cat.known:
    df = df.assign(**{c: df[c].cat.as_known()})
```

Note that converting unknown categories to known categories is expensive, as it involves touching every partition. No strong thoughts if this should be done automatically by datashader if unknown categories are found, or if an informative error should be raised telling the user how to fix this. `df.categorize(columns=[c])` would also be equivalent.

If you know the categories beforehand but dask says they're unknown, you can fix this cheaply by using

```
df[c] = df[c].set_categories(categories)
```

philippjfr:
Thanks @jcrist. Your description matches what I thought about categories, however it does seem like ``.head()`` does infer all the categories despite the fact that the chunk that is actually loaded only contains a single category:

screen shot 2018-08-14 at 10 45 45 pm

Is this maybe specific behavior perhaps due to parquet or is it something we can rely on in general?

jcrist:
> Is this maybe specific behavior perhaps due to parquet or is it something we can rely on in general?

This is parquet files written by fastparquet specific (the categorical data is stored in the parquet metadata, which is not a universal thing to do), and not something generic that you can rely on.

philippjfr:
Okay, thanks. I suppose we could speculatively inspect ``column.head(0)`` of the categorical columns to see if the categories are defined to at least take advantage of that, unless you think there will be a large overhead.

jcrist:
Loading even zero rows loads the whole partition (potentially expensive), and all categorical column partitions will result in categorical data even if the `.cat.categories` isn't exhaustive. You really need to ensure that if you're using a categorical column that the categories in that column are known (`assert df[mycol].cat.known`). Note that `as_known` or `categorize` are no-ops if the categories in that column are already known.

philippjfr:
Thanks, really appreciate your expertise.

philippjfr:
I've opened a separate issue for the categorical issue discussed here since it has considerable performance implications. This PR will instead focus on the optimization of the datashape detection.

datashader/core.py
```diff
@@ -7,10 +7,10 @@
 from xarray import DataArray, Dataset
 from collections import OrderedDict
 
-from .utils import Dispatcher, ngjit, calc_res, calc_bbox, orient_array, compute_coords, get_indices, dshape_from_pandas, dshape_from_dask, categorical_in_dtypes
-from .resampling import resample_2d
+from .utils import Dispatcher, ngjit, calc_res, calc_bbox, orient_array, compute_coords, 
+from .utils import get_indices, dshape_from_pandas, dshape_from_dask
```
philippjfr: Thanks, forgot to add that to my commit apparently.

datashader/core.py
```diff
@@ -514,7 +514,7 @@ def bypixel(source, canvas, glyph, agg):
         cols_to_keep[glyph.x] = True
         cols_to_keep[glyph.y] = True
         if hasattr(glyph, 'z'):
-            cols_to_keep[glyph.z] = True
+            cols_to_keep[glyph.z[0]] = True
```
jbednar: Where did this come from?  Seems odd that it shows up in this PR's tests?

philippjfr: Previously this code was called conditionally, i.e. only for dataframes with categorical columns and the tests weren't testing that.

jbednar: Ok, thanks.

PR #642, opened 2018-09-04
Updates for move to pyviz org

jlstevens:
This PR is to fix the breakages due to moving datashader from https://github.com/bokeh/datashader to https://github.com/pyviz/datashader. I've updated the README and the DNS records but there will still be some CI issues to fix.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,9 +1,9 @@
 Datashader
 ----------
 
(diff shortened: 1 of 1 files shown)


jlstevens:
Travis seems to be working without needing any changes. Appveyor will need configuring and the waffle board is there although the badge doesn't seem to work for some reason?

jlstevens:
@jbednar I would recommend merging this right away: I can work on getting appveyor working separately.

PR #643, opened 2018-09-06
update environment activation instructions

jo:
For me I needed to run `conda` in order to activate the environment, it was also stated in the `conda env create` output:

```
# To activate this environment, use
#
#     $ conda activate ds
#
# To deactivate an active environment, use
#
#     $ conda deactivate
```

```
$ conda --version
conda 4.4.11
```

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -23,7 +23,7 @@ Then run the following commands in your terminal (command) prompt, from wherever
 
 ```bash
 1. conda env create --file environment.yml
-2. source activate ds
+2. conda activate ds
 3. datashader examples
 3. cd datashader-examples
 ```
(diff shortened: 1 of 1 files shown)


ceball:
Whether you should type "source activate x", "conda activate x", or "activate x" depends unfortunately on a combination of: your conda version, the method by which you installed conda, your platform.

It would be great if we could just say it's "conda activate x". Maybe we can now? (I'm not sure, practically speaking.)

ceball:
I'm a bit confused.

The [conda docs](https://conda.io/docs/user-guide/tasks/manage-environments.html#activating-an-environment) say `source activate` (mac/linux) and `activate` (windows).

However, I feel sure I've been typing `conda activate` on all platforms in recent months. 

Looking up through my linux/bash shell history, I see I sometimes get an error for `conda activate`: `CommandNotFoundError: Your shell has not been properly configured to use 'conda activate'.`

My gut feeling is that we should just say to do `conda activate ds` on all platforms (not just mac/linux), along with a note: "For older versions of conda, you may instead need to do `source activate ds` (mac/linux) or `activate ds` (windows)." (It's not as easy as suggesting that people should upgrade conda if `conda activate` doesn't work, because it also depends how they originally installed conda. Upgrading to the latest conda, but having it on your path as previously recommended instead of sourcing conda.sh, means `conda activate` won't work.)

Alternatively, instead of just specifying the command, we could say "2. Activate the ds environment as prompted by the `conda env create` command (typically `conda activate ds`).

We should probably ask for a recommendation from the conda team, and check if conda's documentation is out of date.

jbednar:
`conda activate` was implemented sometime early in 2018, but I think it's only relatively recently that it's finally been fully released and most people upgraded so that it's safe as a blanket recommendation.  So I'm happy with Chris's gut feeling here.

jo:
Hi, thanks for the discussion!

I modified the commit to include the note by @ceball. I like it more than referring to the `conda env create` command output, because its more explicit. Feel free to merge or do the change by yourself, or ping me to make further changes, eg. change commit message or so.

Thousand thanks for this project!

jbednar:
That's frustrating.  I myself have never gotten it working in my own environment, and in fact trying to do so messed up `source activate envt` as well, so that I have to do a workaround. :-(

Issue #570, opened 2018-03-12
Datashader crashes python with "noisy" timeseries data.


bardiche98:
It appears as if datashader renders low noise time series data such as the following no problem:

![low_noise](https://user-images.githubusercontent.com/35264199/37307901-3bf957de-2634-11e8-9d3b-18afd17a1acf.PNG)

However, when attempting to zoom-in on high-noise data such as the following:

![high_noise](https://user-images.githubusercontent.com/35264199/37307944-5f4ec462-2634-11e8-931c-af4d581b5902.PNG)

python bombs out with an uncaught win32 exception:

![python_crash_image](https://user-images.githubusercontent.com/35264199/37307965-7258e272-2634-11e8-9eef-b5d536127788.PNG)

I have attached a python notebook which shows this behavior (at least on my machine):

Environment:
* Windows 7
* new data shader environment (as of 2018-03-12)
* * datashader 0.6.5
* * python 3.6
* * jupyter 1.0.0

The datashader envrironment was generated using the attached yml file from the datashader examples directory.
[](url)

bardiche98:
environment.yml content:
--------------
name: ds
channels:
 - bokeh
 - conda-forge
 - ioam

dependencies:
 - attrs
 - beautifulsoup4
 - bokeh
 - cartopy
 - colorcet
 - conda-forge::graphviz
 - conda-forge::pytest
 - conda-forge::pytest-benchmark
 - conda-forge::python-graphviz
 - dask>=0.15.4
 - datashader
 - dill
 - distributed
 - fastparquet
 - flake8
 - geoviews
 - ioam::holoviews>=1.8.3
 - ipython
 - iris
 - jupyter
 - jupyter_dashboards
 - krb5
 - matplotlib
 - nbconvert
 - nbformat
 - networkx>=2.0
 - numba
 - numpy
 - pandas
 - param>=1.5.1
 - paramnb
 - pyproj
 - pytables
 - python-snappy
 - python=3.6
 - rasterio
 - requests
 - scikit-image
 - scipy
 - shapely
 - snappy
 - statsmodels
 - tblib
 - xarray
 - yaml
 - pip:
   - cachey
   - streamz==0.2.0
   - webargs

bardiche98:
Here is the python notebook... just remove the .txt
----------

[datashader_tseries_high_noise_error.ipynb.txt](https://github.com/bokeh/datashader/files/1804496/datashader_tseries_high_noise_error.ipynb.txt)

[...]

PR #644, opened 2018-09-06
Fix mapping onto pixel for line

ceball:
Previously, when a line was clipped, the computed end point on the bounds could be mapped to a pixel outside the bounds because of floating point error. This showed up as #570.

There should be no change for points or triangles. However: 
  * triangles and lines can no longer share a 'map onto pixel' function
  * I've added some comments to points and triangles.

The code changes and comments should be reviewed carefully.

(The various 'build map for x' (and similar) functions seem a bit clumsy; should consider reorganizing the code a bit in the future. Have avoided doing that here so as not to confuse things.)

### DIFF
--- a/datashader/glyphs.py
+++ b/datashader/glyphs.py
@@ -162,33 +162,8 @@ class Line(_PointLike):
     """
     @memoize
     def _build_extend(self, x_mapper, y_mapper, info, append):
-
-        @ngjit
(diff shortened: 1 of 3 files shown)


jbednar:
Looks great.  Happy to see it merged once you check for other similar issues.

PR #647, opened 2018-09-09
WIP: Packaging updates

ceball:
* Use pyproject.toml for declaring build dependencies
* Use param for setup.py version (should be no change in behavior: same autover functions as before, they're just in param)
* Use pyct for packaging examples (should be no change in behavior: same hv functions as before, they're just in pyct).
* Use pyct for user example commands (should be little or no change in behavior: was previously using a prototype of pyct, pvutil)
* Travis tag detection now matches pyviz standard tags (shouldn't have any practical effect for datashader)
* Simplified travis release vs dev packages (re-used more code; should be no change to behavior).

Still more work to do in the future:
  * use setup.cfg to declare setup data
  * use setup.cfg to define conda package(s) (next release of pyctdev will include a default conda recipe template, which should be able to replace the one in this repository and generate multiple packages if that's ever required for datashader)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -84,7 +84,7 @@ jobs:
       before_install: pip install pyctdev && doit ecosystem=pip ecosystem_setup
       install:
         - unset PYENV_VERSION && pyenv global 3.6 2.7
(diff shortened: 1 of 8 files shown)


jbednar:
This all looks good so far...

PR #662, opened 2018-10-06
Updated attractors notebook to store colormaps

jbednar:


stonebig:
do you still need that testpath<0.4 ?

Issue #666, opened 2018-11-21
Tests have started failing


ceball:
Looks like the cron/weekly tests started failing about two weeks ago.

https://travis-ci.org/pyviz/datashader/jobs/451558963

```
________________ ERROR collecting datashader/tests/test_dask.py ________________
datashader/tests/test_dask.py:12: in <module>
    set_options(get=get_sync)
../../../miniconda/envs/test-environment/lib/python3.6/site-packages/dask/context.py:18: in set_options
    raise TypeError("The dask.set_options function has been deprecated.\n"
E   TypeError: The dask.set_options function has been deprecated.
E   Please use dask.config.set instead
E   
E     Before: with dask.set_options(foo='bar'):
E                 ...
E     After:  with dask.config.set(foo='bar'):
E                 ...
________________ ERROR collecting datashader/tests/test_dask.py ________________
datashader/tests/test_dask.py:12: in <module>
    set_options(get=get_sync)
../../../miniconda/envs/test-environment/lib/python3.6/site-packages/dask/context.py:18: in set_options
    raise TypeError("The dask.set_options function has been deprecated.\n"
E   TypeError: The dask.set_options function has been deprecated.
E   Please use dask.config.set instead
E   
E     Before: with dask.set_options(foo='bar'):
E                 ...
E     After:  with dask.config.set(foo='bar'):
E                 ...
!!!!!!!!!!!!!!!!!!! Interrupted: 2 errors during collection !!!!!!!!!!!!!!!!!!!!
```

From a quick glance, looks like it might be a simple fix?

I haven't got round to sorting out build alerts yet, unfortunately, so just creating an issue to note the failures for now...

SchoolGuy:
I am currently preparing a fix for that. Will open a PR as soon as it is resolved locally.

PR #667, opened 2018-11-22
Fix the broken test because of the dask api-change.

SchoolGuy:
Fixes:  #666

### DIFF
--- a/datashader/tests/test_dask.py
+++ b/datashader/tests/test_dask.py
@@ -1,5 +1,5 @@
 from dask.local import get_sync
-from dask.context import set_options
+from dask.context import config
 import dask.dataframe as dd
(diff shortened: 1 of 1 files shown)


philippjfr:
Thanks, looks good to me.

jbednar:
Thanks!  I guess we don't need to add any version pinning or version checks, because we're only testing with the latest Dask, and this code is only in our tests, not our actual codebase?

SchoolGuy:
I can't say because my goal was just to fix the test. Problem is/was when setting up the dev-environment for me that my PyCharm complained about a lot of missing packages which sometimes resulted in tests failing which are actually not failing. If you like I open another PR for that. (@philippjfr @jbednar )

PR #671, opened 2018-12-17
Removed numba jit from smooth function

philippjfr:
- [x] Fixes https://github.com/pyviz/datashader/issues/670


jbednar:
Have you tested whether this affects performance?

philippjfr:
Not yet, I'll do that now.

philippjfr:
Can't see any difference, with or without the decorator I get 240-250 ms on the example in the linked issue.

jbednar:
Ok, sounds good.  Looks like the tests are calling something deprecated in dask?  "The dask.set_options function has been deprecated. Please use dask.config.set instead"?

philippjfr:
I thought we had already don that.

philippjfr:
Ah it's just one test. Want me to update that in this PR?

jbednar:
Thanks.  I don't think we need to pin dask as a result, since this is just for the test suite, which isn't tested with older dasks.

PR #673, opened 2018-12-19
Updated the user guide

jbednar:
Avoids warnings for old HoloViews option setting, and adds other minor updates to the user guide and roadmap.  Now requires a recent version of HoloViews, which should be specified in setup.py once released.


jlstevens:
PR #693 applies the changes in this PR, ignoring option related changes that have already been applied to master. That PR can be reviewed and compared to this one for the option changes but otherwise, I think that PR supersedes this one.

PR #675, opened 2018-12-20
Allow the use of xarray objects as the input to datashader 2D functions

### LINKED ISSUE #201: Using xarray.dataset as input
It seems like it should be straightforward to add xarray.dataset as an input option to datashader. If the capability is already there, then maybe just adding a sentence or two to the docs would do it. 

I have an example with a work-around here in box 7: http://nbviewer.jupyter.org/gist/jsignell/a6bf9d8545837e18dcf8fef66d556a76

jsignell:
Closes #201

### DIFF
--- a/datashader/core.py
+++ b/datashader/core.py
@@ -147,7 +147,7 @@ class Canvas(object):
 
         Parameters
         ----------
-        source : pandas.DataFrame, dask.DataFrame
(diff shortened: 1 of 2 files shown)


jsignell:
I don't have merge access. But this is good to go!

philippjfr:
Looks good to me too, I'll merge.

Issue #618, opened 2018-07-12
Dashboard example is broken in latest release


jbednar:
```
(pyviz) bash-3.2$ python dashboard/dashboard.py -c dashboard/census.yml
python dashboard/dashboard.py -c dashboard/census.yml
Traceback (most recent call last):
  File "dashboard/dashboard.py", line 17, in <module>
    from bokeh.models import (Range1d, ImageSource, WMTSTileSource, TileRenderer, DynamicImageRenderer, Row, Column)
ImportError: cannot import name 'ImageSource'
(pyviz) bash-3.2$ 
```

harrismia:
Hi,   Thanks for making datashader available.  I am just getting started and would like to see what the dashboards are all about.   I am running into this error.  Do you have any plans to fix it.  Thanks!

jbednar:
The dashboard will be replaced entirely with one based on [Panel](http://panel.pyviz.org).  See [pyviz.org tutorial 1](http://pyviz.org/tutorial/01_Workflow_Introduction.html#Branching-out:-large-data,-geo-data,-custom-controls) for an example Datashader-based dashboard, though not yet with the full set of features of the older (but broken) one.

harrismia:
Thanks, I will take a look.   I am planning to visualize some large genomic datasets and am hoping to use datashader.   I am an experienced Java developer but new to datashader and a Python beginner.   Do you have any recommendations?  Would pyviz.org be the best place to start ?    I am very impressed with the datashader package and have gotten most of the example notebooks working.  Now I’m ready to start working with my own data.

jbednar:
Datashader.org focuses on Datashader itself, but it works best in conjunction with other tools, for which pyviz.org gives the best overview.  If you come up with any good genomic data plotting examples, I'd love to publish those, as several other people have also indicated that they are using Datashader for genomic or proteomic data (but none have so far made anything public).

Issue #252, opened 2016-10-12
Hover not showing up in dashboard.py nyc taxi example


vollkorn1982:
I'm trying to run the nyc taxi example like this:

```
python dashboard.py -c nyc_taxi.yml
```

When I move the mouse cursor over the displayed data points, no mouse hover shows up.

The hover tool symbol is showing up on the right hand side and I can check and uncheck the box which pops up, when I click on it, but that does not change anything.

Hover works fine with the examples from the bokeh package.

I am on a 64 bit Ubuntu 16.04 and have a python 3.5.2 virtualenv in which I installed the current datashader master branch from git and bokeh 0.12.3 via pip.

EDIT: datashader was installed from git master, not pip

jbednar:
Unfortunately, the Bokeh API keeps changing, and we have not yet been able to update the dashboard example to work with the latest Bokeh.   Hover information should appear with the dashboard if you use Bokeh 0.11 (``conda install bokeh=0.11.1``), but it won't appear for any later Bokeh version up to 0.12.4.  On the other hand, hover information is working fine for the hover_layer.ipynb example, for the latest Bokeh release.  

I haven't personally looked into what the changes were or what we need to do to update the dashboard, but if you wanted to check it out, just compare the code inside dashboard.py with hover_layer.ipynb, to see if you can see what to change about dashboard.py to make it work the same way.  If you can find the problem, we'd be happy to accept a pull request for dashboard.py!  You shouldn't need to understand datashader itself to do this, just how to use bokeh, since hover_layer.ipynb already shows how to do the datashader side of the calculations.

tommycarstensen:
@vollkorn1982 Any updates on this? Thanks.

jbednar:
We have not investigated the dashboard in particular, but you can get hover working using HoloViews 1.7 (due for release this week) as described in issue #126.

philippjfr:
This issue references the old Bokeh based dashboard but we've long switched to a Panel based version. In future we will use Bokeh support for colormapping so that we can use hover and colormaps properly.

Issue #221, opened 2016-08-18
Fix/Upgrade dashboard example to Bokeh 0.12.1


brendancol:

jbcrail:
Even though the NYC taxi example is usable, the hover and legend are still not working.

philippjfr:
Bit confused why this was bumped for so long, the old Bokeh based dashboard was replaced with a Panel version which does support more recent versions of Bokeh.

jbednar:
That's odd; #676 says that it closes this issue, but I guess Github's robots didn't parse that. Yes, it should be closed, thanks.

PR #676, opened 2018-12-23
Panel-based Datashader dashboard

### LINKED ISSUE #358: Simplify dashboard example
`examples/dashboard/dashboard.py` evolved over time, alongside Bokeh server, and the result is now fairly complex and difficult to follow, as well as difficult to maintain.  We should make a new, clean implementation in HoloViews (to show how the basic approach can work at a high level), along with a new, clean pure Bokeh version (to provide a simpler starting point for pure Bokeh work than our current version).

jbednar:
This PR closes #358, #618, and #221 by providing a [Panel](http://panel.pyviz.org)-based dashboard to replace the current pure Bokeh version that has been broken for several releases now, with the Intake support provided by @jsignell.  The old Bokeh dashboard was over 500 lines of code that I was never able to update or even keep running, hence the long period of breakage.  This new version fits into 50 lines of code (so far) and has nearly the same functionality, once the to-do items below are addressed:

- [x] opacity (of the data) and spread_size are not hooked up to the control
- [x] the widgets aren't in any order yet
- [ ] it would be nice to separate some widgets to be on top of the plot and others down the side, if there is a clear semantic distinction
- [x] colormaps should be updated when the plot changes
- [ ] opensky data is not tested
- [x] osm data is not tested and is not yet listed in the .yml file
- [ ] the basemap argument doesn't update the map tiles, due to an open Bokeh bug in caching tiles
- [x] should probably add a logo
- [x] Need to allow field (fare, etc.) to be selected

There are also other to-do items that weren't supported in the last working version of the dashboard:

- [ ] Hover info (#252)
- [ ] Colorbar and/or categorical legend


examples/dashboard.yml
```diff
@@ -0,0 +1,100 @@
+sources:
+  nyc_taxi:
+    description: Large version of nyc taxi dataset
+    driver: parquet
+    args:
+      urlpath: '{{ CATLOG_DIR }}data/nyc_taxi_wide.parq'
```
philippjfr: Bit confused here, ``nyc_taxi_wide.parq`` is still not listed in the datashader datasets.yaml, it is only in the pyviz on.

examples/dashboard.yml
```diff
@@ -0,0 +1,100 @@
...
+          label: Pick-up Hour
+        fare_amount:
+          label: Fare
+          unit: $
+        passenger_count:
+          label: Passenger Count
+        counts:
+          label: Ride counts
+      plot:
+        xlim: !!python/tuple [-8240227.037, -8231283.905]
```
philippjfr: I think I've arranged it so you can simply provide lists and they are cast appropriately.

examples/dashboard.yml
```diff
@@ -0,0 +1,100 @@
...
+            b: green
+            a: red
+            h: orange
+            o: saddlebrown
+  
+  opensky:
+    description: OpenSky Flight Paths
+    driver: netcdf
+    args:
+      urlpath: '{{ CATLOG_DIR }}data/opensky.h5'
```
philippjfr: Is ``{{ CATLOG_DIR }}`` really the correct way to spell this? I've always used ``{{ CATALOG_DIR }}/``?

philippjfr:
I've pushed updates to apply opacity and spread_size, however control over global opacity required a small fix in holoviews, which we can push out as part of a 1.11.1 release (see https://github.com/ioam/holoviews/pull/3371).

examples/dashboard.yml
philippjfr: Actually I'm probably wrong about that, I think that was explicitly for dimension ranges.

examples/dashboard.yml
jbednar: Good point. I think we should probably delete the .csv version; .csv is really not a recommended format for Datashader.

examples/dashboard.yml
jbednar: Would be good to provide support for such casting if you can, to avoid this awkwardness.

examples/dashboard.yml
jbednar: I think it looks more natural with the /, but it seems like both ways work?

jbednar: Oh, you mean CAT LOG?  That seems like a typo.

jbednar: This one hasn't been tested, so I'll just go ahead and fix that.

philippjfr: I meant both things, glad it's a typo though.

philippjfr: >This one hasn't been tested, so I'll just go ahead and fix that.

The typo appears in all of them.

jbednar: Odd; not sure how it's ever worked, then. "CATLOG" doesn't appear in the intake docs or repo.

jbednar: (Scratches head) Odd.  I've fixed that and pushed.

philippjfr: The variable was probably just replaced with an empty string so it would work as long as the dashboard.ipynb is in the same place as the data directory.

examples/dashboard.yml
jbednar: Oh, it's definitely YAML's fault, or at least JSON's fault (from which YAML inherits its scope).  But the fact remains that it's awkward for Python users to write code like the above, so we should probably accept a list as well as a tuple unless we are already using the list/tuple distinction for some semantic purpose in this case.

jbednar:
Current version:

image

Still just 54 lines.  Need to add support for selecting which field to plot.

examples/dashboard.yml
```diff
@@ -0,0 +1,120 @@
...
+            b: green
+            a: red
+            h: orange
+            o: saddlebrown
+  
+  opensky:
+    description: OpenSky Flight Paths
+    driver: netcdf
+    args:
+      urlpath: '{{ CATALOG_DIR }}/data/opensky.h5'
```
jsignell: I think you need to add a chunks argument, but you can set it to empty:

```suggestion
      urlpath: '{{ CATALOG_DIR }}/data/opensky.h5'
      chunks: {}
```

examples/dashboard.yml
```diff
@@ -0,0 +1,121 @@
...
+          label: Fare
+          unit: $
+        passenger_count:
+          label: Passenger Count
+        counts:
+          label: Ride counts
+      plot:
+        xlim: !!python/tuple [-8240227.037, -8231283.905]
+        ylim: !!python/tuple [4974203.152, 4979238.441]
+        kind: points
```
jsignell: If you need to have access to other columns later, then we can include them in the original points plot

```suggestion
        kind: points
        hover_cols: ['dropoff_hour', 'pickup_hour', 'passenger_count']
```

examples/dashboard.ipynb
```diff
@@ -0,0 +1,139 @@
...
+    "    def labels(self):\n",
+    "        return gts.StamenLabels.options(level='annotation', alpha=1 if self.show_labels else 0)\n",
+    "    \n",
+    "    @pm.depends('data_opacity')\n",
+    "    def apply_opacity(self, shaded):\n",
+    "        return shaded.opts(alpha=self.data_opacity, show_legend=False)\n",
+    "\n",
+    "    def view(self,**kwargs):\n",
+    "        points   = hv.DynamicMap(self.points)\n",
+    "        agg      = rasterize(points, width=600, height=400)\n",
```
jsignell: To aggregate on another field, you need to add an aggregator to the rasterize call. The field can be another param stream. 

```suggestion
    "        agg      = rasterize(points, width=600, height=400, aggregator=rd.mean('passenger_count'))\n",
```

examples/dashboard.ipynb
```diff
@@ -0,0 +1,139 @@
...
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "gopts = hv.opts.WMTS(width=800, height=500, xaxis=None, yaxis=None, bgcolor='black', show_grid=False)\n",
+    "cmaps = {n: colorcet.palette[n] for n in ['fire', 'bgy', 'bgyw', 'bmy', 'gray', 'kbc']}\n",
+    "norms = {'Histogram_Equalization': 'eq_hist', 'Linear': 'linear', 'Log': 'log', 'Cube root': 'cbrt'}\n",
+    "maps  = ['CartoMidnight', 'StamenWatercolor', 'StamenTonerBackground', 'EsriImagery', 'EsriUSATopo', 'EsriTerrain']\n",
+    "bases = {name: ts.relabel(name) for name, ts in gts.tile_sources.items() if name in maps}\n",
+    "plots = {source.metadata['plots'][p].get('label', p):p for p in source.plots}\n",
```
jsignell: So just limit the list of fields to ones that you want to allow aggregation over and have one just be counts or something if you want to allow that. 

This will get a dict of labelled selectors for all the available aggregator fields. 

```suggestion
    "plots = {source.metadata['plots'][p].get('label', p):p for p in source.plots}\n",
    "fields = {k: v.get('label', k), for k, v in source.metadata['fields'].items()}\n",
```

jbednar:
I've made a number of changes in the above commits:

- Added `nyc_taxi_50k` to datasets.yml as a quick test case to use
- Updated the .yml file to consistently list "counts" as the first item in the list of plottable fields, as it makes a good default.  
- Removed any mention of "fare" from the .yml, because the nyc_taxi parq files don't currently include the fare column.  (At some point I need to make new Parquet files including fare and also adding JFK to the visible range of points.)
- Added an "agg_fn" parameter that allows the user to select min, max, mean, etc.

All of that seems fine so far. I've also made some changes that don't quite work yet:

- [ ] I changed the `.points` method to `.rasterize` so that it could specify the agg_fn, which works but required me to make `rasterize` no longer be dynamic, which means that RangeXY is no longer tied to this object and zooming no longer updates the datashaded plot.
- [ ] I added a "field" parameter, which is populated correctly but doesn't seem to have any effect yet, despite being applied in the same method as the agg_fn and having the same dependency structure.

jbednar:
Thanks, @philippjfr for re-connecting RangeXY and for realizing that the field parameter _is_ working, it just has no noticeable effect for a counts aggregator; e.g. switch to `Mean` to see the field having an effect. I've renamed `.points` to `.elem` because the plot isn't always a points plot; depends what's declared in the data catalog.  

I also reorganized the code a bit and added a very extensive section at the end unpacking how it works.  It would be nice not to have to use three different mechanisms for getting the DynamicMaps involved, but I'll open a separate issue about that.  Meanwhile, I think it's a very concise and *mostly* understandable way to express some quite complicated relationships between user-visible controls and computations, and it's a massive improvement over the old dashboard, so it should be ready to merge when tests pass.

examples/dashboard.yml
jbednar: Fixed in https://github.com/pyviz/hvplot/pull/248.

PR #680, opened 2019-01-05
Fix handling of the span argument to tf.shade

### LINKED ISSUE #368: 'shade' function doesn't correctly implement 'span'
When I used the span argument to shade, for example in the following callback function for Bokeh:
```
def create_image(x_range, y_range, w, h):
    cvs = ds.Canvas(x_range=x_range, y_range=y_range, plot_width=200, plot_height=200)
    agg = cvs.points(df, 'x', 'y', ds.min('z'))
    img = tf.shade(agg, cmap=chi2cmap, how='linear', span=[mn,mx])
    #return tf.dynspread(img, threshold=0.9, max_px=10)
    return img
```
then I expect the colormap to be absolutely normalised according to the min and max 'z' values supplied in 'span'. However, Datashader doesn't do this, it seems to offset the colors by the minimum value of 'z' in the data. This can be observed by dynamically zooming with bokeh, one seems the colors change rather than remain constant.

Looking at the code, I think it is because of the following lines in `https://github.com/bokeh/datashader/blob/master/datashader/transfer_functions.py`:

```
def _interpolate(agg, cmap, how, alpha, span):
    if agg.ndim != 2:
        raise ValueError("agg must be 2D")
    interpolater = _normalize_interpolate_how(how)
    data = agg.data
    if np.issubdtype(data.dtype, np.bool_):
        mask = ~data
        interp = data
    else:
        if np.issubdtype(data.dtype, np.integer):
            mask = data == 0
        else:
            mask = np.isnan(data)

        masked = data[~mask]
        if len(masked) == 0:
            return Image(agg.data.view(np.uint32), coords=agg.coords, dims=agg.dims, attrs=agg.attrs)

        offset = masked.min()

        interp = data - offset
        
    data = interpolater(interp, mask)
...
```

Here it seems that some offset is indeed applied according the to minimum of the dataset. I assume this is what is screwing up the interpretation of 'span'. Probably the offset needs to be added back in after the interpolation is applied?

jonmmease:
This PR closes #368 

Tests are added to ensure that:
 1) Manually setting `span` to the min/max of the aggregated data is the same as setting `span` to `None`
 2) When `span` is set, removing the smallest or largest aggregation values alters only the single
    corresponding pixel in the resulting image
 3) Unmasked aggregation values that fall outside of `span` are clipped to the span
 4) The `span` can be specified using either floats or integers

To make these tests pass, the implementation is updated so that:
 1) When `span` is provided, the dataset should be shifted by `span[0]` and clipped
to the `span` before the interpolation transform is applied. When `span` is computed
automatically, the dataset is shifted by the minimum value and no clipping is necessary.
 2) The transformed span should be computed by shifting the `span` by `span[0]` before
the transformation.

~**Note:** Waiting on https://github.com/pyviz/datashader/pull/682 for travis test fixes~

### DIFF
--- a/datashader/tests/test_transfer_functions.py
+++ b/datashader/tests/test_transfer_functions.py
@@ -55,6 +55,53 @@ eq_hist_sol = {'a': np.array([[0, 4291543295, 4288846335],
 eq_hist_sol['c'] = eq_hist_sol['b']
 
 
+def check_span(x, cmap, how, sol):
+    # Copy inputs that will be modified
+    sol = sol.copy()
+    x = x.copy()
+
+    # All data no span
+    img = tf.shade(x, cmap=cmap, how=how, span=None)
+    assert img.equals(sol)
+
+    # All data with span
+    img = tf.shade(x, cmap=cmap, how=how, span=float_span)
+    assert img.equals(sol)
+
+    # Decrease smallest. This value should be clipped to span[0] and the
+    # resulting image should be identical
+    x[0, 1] = 10
+    x_input = x.copy()
+    img = tf.shade(x, cmap=cmap, how=how, span=float_span)
+    assert img.equals(sol)
+
+    # Check that clipping doesn't alter input array
+    assert x.equals(x_input)
+
+    # Increase largest. This value should be clipped to span[1] and the
+    # resulting image should be identical
+    x[2, 1] = 18
+    x_input = x.copy()
+    img = tf.shade(x, cmap=cmap, how=how, span=float_span)
+    assert img.equals(sol)
+
+    # Check that clipping doesn't alter input array
+    assert x.equals(x_input)
+
+    # zero out smallest. If span is working properly the zeroed out pixel
+    # will be masked out and all other pixels will remain unchanged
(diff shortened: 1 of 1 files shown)


ceball:
From a quick glance at the screenshot, it looks like a problem with one of the temporary workarounds in the pyviz build/test system after a change in an underlying external tool/repository. I.e. not related to datashader (or your PR). I'll confirm that (and fix it if so) this weekend, and get back to you.

In case it sounds scary that we have our own build/test system: we just call external tools (e.g. conda build) to do the actual work, but have some temporary workarounds for issues in those underlying tools (issues which hopefully get addressed over time, meaning we can lose the workarounds).

ceball:
The build environment issue should be fixed in #682.

jbednar:
Thanks so much for taking this on! My eyes glaze over and my brain starts overheating when I look at this part of the code, but the tests look good to me at least.  Are there any tests to cover the three different type of colormaps (list, string/tuple, and callable (matplotlib cmap))?  The logic looks different in those three cases and I'm not certain they will have the same results as they should, when used with span.

PR #682, opened 2019-01-05
Set minimum version of pyctdev 

ceball:
Avoids installation of much older version of pyctdev (behavior that started with conda 4.5.12, which was released 3 days ago; does not happen with 4.5.11).


ceball:
This change or an equivalent will probably need to be applied to several pyviz projects.

See also https://github.com/pyviz/pyct/issues/72

PR #685, opened 2019-01-10
Updated the notebooks to use current HoloViews syntax

jlstevens:
I've only done the first user guide so far but I found a lot more things that needed improvement than just the removal of the holoviews magics. One thing that isn't fixed yet in that user guide is a workaround for the broken markdown rendering in the current notebook version.


jlstevens:
While updating the gerrymandering notebook, I get this error:

image

@philippjfr says he can reproduce the issue so maybe the shapefile is no longer valid or incompatible?

jlstevens:
A couple of notes about the network_packets notebook:

* I got a rid of a lot of annoying padding related logic.
* It was using `color_index='weight'` which seems wrong as that would map the nodes (`node_color`) to weights and not the edges? Now using `edge_color` instead...
* `node_hover_size` didn't seem to be accepted to I changed it to `node_size`. This is probably a bug as I do think the former was intended.

jlstevens:
@jbednar Ready to squash merge. 

Note that two commits that were meant to be part of this PR accidentally ended up directly on master (updates to the UK researchers notebook and the removal of the notebooks now on EarthML)

examples/topics/network_packets.ipynb
```diff
@@ -106,7 +107,10 @@
     "\n",
     "hv.extension('bokeh')\n",
     "\n",
-    "%opts RGB Graph Nodes [bgcolor='black' width=800 height=800 xaxis=None yaxis=None]"
+    "opts.defaults(\n",
+    "    opts.Graph(bgcolor='black', width=800, height=800, xaxis=None, yaxis=None),\n",
+    "    opts.Nodes(bgcolor='black', width=800, height=800, xaxis=None, yaxis=None),\n",
+    "    opts.RGB(  bgcolor='black', width=800, height=800, xaxis=None, yaxis=None))"
```
jbednar: I'd vote for:

```
o = dict(bgcolor='black', width=800, height=800, xaxis=None, yaxis=None)
opts.defaults(opts.Graph(**o), opts.Nodes(**o), opts.RGB(**o))
```

examples/topics/osm-1billion.ipynb
```diff
@@ -164,14 +164,14 @@
...
     "import dask.dataframe as dd\n",
     "from holoviews.operation.datashader import datashade, dynspread\n",
     "hv.extension('bokeh')\n",
     "\n",
-    "sizes  = dict(width=plot_width,height=plot_height)\n",
-    "opts   = dict(xaxis=None, yaxis=None, bgcolor=\"black\", **sizes)\n",
+    "rgb_opts = opts.RGB(width=plot_width,height=plot_height, xaxis=None, yaxis=None, bgcolor=\"black\")\n",
     "points = hv.Points(df, ['x', 'y']).redim.range(x=bounds[\"x_range\"],y=bounds[\"y_range\"])\n",
-    "dynspread(datashade(points, cmap=[\"darkblue\", \"lightcyan\"], **sizes)).options(**opts)"
+    "dynspread(datashade(points, cmap=[\"darkblue\", \"lightcyan\"], **sizes)).options(rgb_opts)"
```
jbednar: Why .options and not .opts?

examples/user_guide/1_Plotting_Pitfalls.ipynb
```diff
@@ -113,8 +121,8 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "%%opts Points (s=50 alpha=0.1)\n",
-    "blues + reds + reds*blues + blues*reds"
+    "layout = blues + reds + reds * blues + blues * reds\n",
+    "layout.opts(opts.Points(s=50, alpha=0.1))"
```
jbednar: PEP8 says to leave the spacing around * like it was: https://www.python.org/dev/peps/pep-0008

jbednar: "If operators with different priorities are used, consider adding whitespace around the operators with the lowest priority(ies)."

examples/topics/osm-1billion.ipynb
jlstevens: Thanks, missed that.

jlstevens:
I suppose we might not want to merge until we figure out the shape file issue.

jbednar:
I think the shape file issue is orthogonal, so I'd go for it.  That way I can rebase my other PR and have my working version of the shapefile code but based on current master, which is probably a better starting point for figuring it out since I'll simultaneously be working on the dashboard next week.  So let me know when you are done with the above and I'll just merge.

jbednar:
(That said, if you can figure out how to avoid the shapefile issue, please do!)

examples/user_guide/1_Plotting_Pitfalls.ipynb
jlstevens: To be honest, I think explicit parentheses are often the most readable thing. To maximize readability I would do:

```
layout = blues + reds + (reds * blues) + (blues * reds)
```

But otherwise, without the parentheses I do agree with PEP8 in this case.

examples/topics/osm-1billion.ipynb
jlstevens: Fixed in 46c68de

examples/user_guide/1_Plotting_Pitfalls.ipynb
jbednar: Ok.  I'd only use parentheses if the items involved are somehow complex, but "reds" and "blues" aren't enough to justify parens, for me...

examples/topics/network_packets.ipynb
philippjfr: +1, we've had some disagreement about this but in holoviews I've generally favored an approach like this.

examples/user_guide/1_Plotting_Pitfalls.ipynb
jlstevens: I improved the readability of these expressions in 3de0e14 using parentheses. I prefer this format but if you really don't like it, I'll have to revert to the mixed spacing around the operators.

jbednar: I don't prefer it but don't mind; just leave it as you have it now.

examples/topics/network_packets.ipynb
jlstevens: I really prefer to avoid `**` when possible. I would prefer:

```python
o = opts.Options(bgcolor='black', width=800, height=800, xaxis=None, yaxis=None)
opts.defaults(opts.Graph(o), opts.Nodes(o), opts.RGB(o))
```

But this isn't currently supported (but it could be). As it is, I'll implement your suggestion even though I don't hugely like it (but the current redundant version isn't great either)

jlstevens: Fixed in 71f5ee6

jlstevens:
I've now addressed the comments. I would also prefer to have this PR merged sooner rather than later as I doubt I can figure out the shape file issue *that* quickly.

examples/topics/network_packets.ipynb
jbednar: Sure, consider adding support for your proposed syntax to HoloViews, which could at least check if those options are valid for *some* backend. But meanwhile your fix is fine, thanks.

jbednar:
Looks like the one failing test was a transient, so I will go ahead and merge.

PR #687, opened 2019-01-13
Add pandas ExtensionArray for storing homogeneous ragged arrays

jonmmease:
## Overview
This PR introduces a pandas [`ExtensionArray`](http://pandas.pydata.org/pandas-docs/version/0.24.0rc1/extending.html#extensionarray) for storing a column of homogeneous ragged 1D arrays.  The Datashader motivation for ragged arrays is to make it possible to store variable-length lines (fixing problems like https://github.com/pyviz/datashader/issues/464) and eventually polygons (https://github.com/pyviz/datashader/issues/181) as elements of a column in a DataFrame.  Using one such shape per row makes it simpler to store associated columns of data for use with selections and filtering, hovering, etc.

This PR currently contains only the extension array and associated testing.

## Implementation
`RaggedArray` is a subclass of `pandas.api.extension.ExtensionArray` with a `RaggedDtype` that is a subclass of `pandas.api.extension.ExtensionDtype`.  `RaggedDtype` takes advantage of the `@register_extension_dtype` decorator introduced in pandas 0.24rc1 to register itself with pandas as a datatype named `'ragged'`.

**NOTE:** This branch currently requires pandas 0.24rc1

A ragged array of length `n` is represented by three numpy arrays:
 - `mask`: A boolean array of length `n` where values of `True` represent missing/NA values
 - `flat_array`: An array with the same datatype as the ragged array element and with a length equal to the sum of the length of all of the ragged array elements.
 - `start_indices`: An unsigned integer array of length `n` of indices into `flat_array` corresponding to the start of the ragged array element. For space efficiency, the precision of the unsigned integer is chosen to be the smallest available that is capable of indexing the last element in `flat_array`.

## Example Usage
```
In[1]: from datashader.datatypes import RaggedArray
In[2]: ra = RaggedArray([[1, 2], [], [10, 20], None, [11, 22, 33, 44]])
In[3]: ra
Out[3]: 
<RaggedArray>
[            array([1., 2.]),    array([], dtype=float64),
           array([10., 20.]),                        None,
 array([11., 22., 33., 44.])]
Length: 5, dtype: <class 'datashader.datatypes.RaggedDtype'>

In[4]: ra.flat_array
Out[4]: array([ 1.,  2., 10., 20., 11., 22., 33., 44.])

In[5]: ra.start_indices
Out[5]: array([0, 2, 2, 4, 4], dtype=uint8)

In[6]: ra.mask
Out[6]: array([False, False, False,  True, False])

In[7]: pd.array([[1, 2], [], [10, 20], None, [11, 22, 33, 44]], dtype='ragged')
Out[7]: 
<RaggedArray>
[            array([1., 2.]),    array([], dtype=float64),
           array([10., 20.]),                        None,
 array([11., 22., 33., 44.])]
Length: 5, dtype: <class 'datashader.datatypes.RaggedDtype'>

In[8]: rs = pd.Series([[1, 2], [], [10, 20], None, [11, 22, 33, 44]], dtype='ragged')
In[9]: rs
Out[9]: 
0              [1. 2.]
1                   []
2            [10. 20.]
3                 None
4    [11. 22. 33. 44.]
dtype: ragged

In[10]: ragged_subset = rs.loc[[0, 1, 4]]
In[11]: ragged_subset
Out[11]: 
0              [1. 2.]
1                   []
4    [11. 22. 33. 44.]
dtype: ragged

In[12]: ragged_subset.array.mask
Out[12]: array([False, False, False])

In[13]: ragged_subset.array.flat_array
Out[13]: array([ 1.,  2., 11., 22., 33., 44.])

In[14]: ragged_subset.array.start_indices
Out[14]: array([0, 2, 2], dtype=uint8)
```


datashader/datatypes.py
```diff
@@ -0,0 +1,361 @@
...
+                # Update flat array
+                self._flat_array[next_start_ind:next_start_ind+n] = array_el
+
+                # increment next start index
+                next_start_ind += n
+
+        # This is a workaround (hack?) to keep pandas.lib.infer_dtype from
+        # "raising cannot infer type" ValueError error when calling:
+        # >>> pd.Series([[0, 1], [1, 2, 3]], dtype='ragged')
+        self._values = self._flat_array
```
jonmmease: Hack to work around `ValueError: cannot infer type for <class 'NoneType'>` in `pandas._libs.lib.infer_dtype`

jonmmease:
Hi @TomAugspurger, I was wondering if you'd have a little time to look this over.  Any feedback/thoughts you have on the extension array approach would be appreciated, but there are two specific things I wanted to ask you about that I think are related to pandas 0.24rc1.

1) It looks like Datashader had been using `categorical_series.cat.categorical.ordered` to check for an ordered categorical and it seems the `categorical` property of the `CategoricalAccessor` has been removed in 0.24rc1.  Changing this to `categorical_series.cat.ordered` works fine and is more direct, but I just wanted to double check whether this change in pandas was intentional.

2) I'm using the new `register_extension_dtype` decorator to register the `RaggedDtype` with pandas as `'ragged'`. And this makes it possible to construct the ragged array using the new `pd.array` constructor. e.g.:

```
pd.array([[1, 2], [], [10, 20], None, [11, 22, 33, 44]], dtype='ragged')

<RaggedArray>
[            array([1., 2.]),    array([], dtype=float64),
           array([10., 20.]),                        None,
 array([11., 22., 33., 44.])]
Length: 5, dtype: <class 'datashader.datatypes.RaggedDtype'>
```

However, I ran into an error when trying to do the same with a pandas `Series`

```
pd.Series([[1, 2], [], [10, 20], None, [11, 22, 33, 44]], dtype='ragged')

  File "pandas/_libs/lib.pyx", line 1201, in pandas._libs.lib.infer_dtype
TypeError: Cannot convert RaggedArray to numpy.ndarray
During handling of the above exception, another exception occurred:
  File "/anaconda3/envs/datashader_dev/lib/python3.6/site-packages/IPython/core/interactiveshell.py", line 3267, in run_code
    exec(code_obj, self.user_global_ns, self.user_ns)
  File "<ipython-input-4-3fd52f24dd69>", line 1, in <module>
    pd.Series([[1, 2], [], [10, 20], None, [11, 22, 33, 44]], dtype='ragged')
  File "/anaconda3/envs/datashader_dev/lib/python3.6/site-packages/pandas/core/series.py", line 262, in __init__
    raise_cast_failure=True)
  File "/anaconda3/envs/datashader_dev/lib/python3.6/site-packages/pandas/core/internals/construction.py", line 674, in sanitize_array
    inferred = lib.infer_dtype(subarr, skipna=False)
  File "pandas/_libs/lib.pyx", line 1208, in pandas._libs.lib.infer_dtype
ValueError: cannot infer type for <class 'NoneType'>
```

Looking at the source for `lib.infer_dtype`

https://github.com/pandas-dev/pandas/blob/48c3ce5b5b89fa53b3d4ac9c23d4c22da5e86493/pandas/_libs/lib.pyx#L1201

I surmised that I could work around the error by setting a `._values` property in my `RaggedDtype` constructor to a numpy array.  But this doesn't seem like the *right* thing to do.  To reproduce the error, comment out this line at the bottom of the `RaggedArray` constructor.

Thanks! And thanks for all of your efforts on the `ExtensionArray` system.  This is awesome work and really opens up a whole new world for pandas!

TomAugspurger:
Opened https://github.com/pandas-dev/pandas/issues/24751 for the categorical issue. That was a byproduct of a refactor (but the old way is being deprecated so datashader will want to update).

> But this doesn't seem like the right thing to do.

Right. pandas should rather unbox the Series / Index to the array before attempting to infer the dtype. I'll have to look  bit more closely at the infer_dtype failure. In theory, this should reproduce the failure

```python
In [13]: from pandas.tests.extension.arrow.bool import *

In [14]: pd.array([True, False], dtype='arrow_bool')
Out[14]:
ArrowBoolArray(<pyarrow.lib.ChunkedArray object at 0x114371060>
[
  [
    true,
    false
  ]
])

In [15]: pd.Series([True, False], dtype='arrow_bool')
Out[15]:
0     True
1    False
dtype: arrow_bool
```

It's possible that the raggedness of the input is sending things down a different code path (but it shouldn't, when a dtype is specified).

datashader/datatypes.py
```diff
@@ -0,0 +1,361 @@
+import numpy as np
+from pandas.api.extensions import ExtensionDtype, ExtensionArray
+from pandas.core.dtypes.dtypes import register_extension_dtype
```
TomAugspurger: This should be imported from `pandas.api.extensions` as well. Pandas considers `pandas.core` to be private.

jonmmease:
Thanks for pointing out the testing docs, I had missed those. I’ll implement these this evening. Thanks a lot for your help!

jonmmease:
@TomAugspurger Ok, I added the test suite provided by pandas and that helped a lot.  I'm not exactly sure what made the difference, but once I got the pandas test suite passing I didn't need the `._values` hack to get my original tests passing.

One change that might have been the culprit was that the `dtype` property on my `ExtensionArray` subclass was returning my `ExtensionDtype` *class*, rather than an instance of the class.  And the test suite helped me find that 🎉 

I would definitely appreciate any other comments you have on the extension classes, but I think I've made it through what I was stuck on.  Thanks again!

datashader/datatypes.py
```diff
@@ -0,0 +1,364 @@
+import numpy as np
+from pandas.api.extensions import (
+    ExtensionDtype, ExtensionArray, register_extension_dtype)
+from numbers import Integral
+
+
+@register_extension_dtype
+class RaggedDtype(ExtensionDtype):
```
TomAugspurger: Does it make sense to parametrize this by the underlying `dtype` of the data? e.g. a `RaggedDtype[int]`, `RaggedDtype[float]`? Being able to tell the dtype of an output by just looking at the type, and not the values, is often helpful.

TomAugspurger: Pandas typically calls this the `.subdtype` (see SparseArray or IntervalArray).

datashader/datatypes.py
```diff
@@ -0,0 +1,364 @@
...
+
+class RaggedArray(ExtensionArray):
+    def __init__(self, data, dtype=None):
+        """
+        Construct a RaggedArray
+
+        Parameters
+        ----------
+        data
+            List or numpy array of lists or numpy arrays
```
TomAugspurger: Seems that a `dict` of the correct schema is also accepted. Or is that intended to be "private"? Could that be placed in an alternative constructor `_from_dict`?

It may also be useful to accept a `RaggedArray` here..

datashader/datatypes.py
```diff
@@ -0,0 +1,364 @@
...
+        """
+        if isinstance(item, Integral):
+            if item < -len(self) or item >= len(self):
+                raise IndexError(item)
+            elif self.mask[item]:
+                return None
+            else:
+                # Convert negative item index
+                if item < 0:
+                    item = 5 + item
```
TomAugspurger: Sily question: why is this 5, and not `len(self)`?

datashader/datatypes.py
```diff
@@ -0,0 +1,364 @@
...
+        """
+        Length of this array
+
+        Returns
+        -------
+        length : int
+        """
+        return len(self._start_indices)
+
+    def __getitem__(self, item):
```
TomAugspurger: What do you think should be the return value here (using the RaggedArray from your original post?

```python
In [25]: ra[[0]]
KeyError                                  Traceback (most recent call last)
<ipython-input-25-9a9c4b76d0ab> in <module>
----> 1 ra[[0]]

~/sandbox/datashader/datashader/datatypes.py in __getitem__(self, item)
    184             return RaggedArray(data, dtype=self.flat_array.dtype)
    185         else:
--> 186             raise KeyError(item)
    187
    188     @classmethod

KeyError: [0]
```

I believen that should be (possibly)

```python
In [29]: np.atleast_2d(ra[0])
Out[29]: array([[1., 2.]])
```

though still thinking things through.

datashader/datatypes.py
```diff
@@ -0,0 +1,364 @@
...
+        elif isinstance(item, np.ndarray) and item.dtype == 'bool':
+            data = []
+
+            for i, m in enumerate(item):
+                if m:
+                    data.append(self[i])
+
+            return RaggedArray(data, dtype=self.flat_array.dtype)
+        else:
+            raise KeyError(item)
```
TomAugspurger: Probably an IndexError?

TomAugspurger:
Just to be clear, there are a whole  bunch of tests you can inherit, not just BaseDtypeTests. See e.g. https://github.com/pandas-dev/pandas/blob/master/pandas/tests/extension/decimal/test_decimal.py

datashader/datatypes.py
```diff
@@ -0,0 +1,364 @@
...
+
+        start_indices = np.hstack([ra.start_indices + offset
+                                   for offset, ra in zip(offsets, to_concat)])
+
+        return RaggedArray(dict(
+            mask=mask, flat_array=flat_array, start_indices=start_indices))
+
+    @property
+    def dtype(self):
+        return RaggedDtype()
```
TomAugspurger: Maybe not a big deal, but it's a bit strange for the `id` of a `.dtype` to change on each call. Probably best to set a `._dtype` in the `__init__` and return `self._dtype` here.

datashader/datatypes.py
```diff
@@ -0,0 +1,364 @@
...
+    def nbytes(self):
+        """
+        The number of bytes needed to store this object in memory.
+        """
+        return (self._flat_array.nbytes +
+                self._start_indices.nbytes +
+                self._mask.nbytes)
+
+    def astype(self, dtype, copy=True):
+        if isinstance(dtype, RaggedDtype):
```
TomAugspurger: Note: some people may hope that `.astype('category')` (or any other extension dtype) works here.

I'm not sure whether it makes sense to support that. But just thought I'd make you aware that `ExtensionArray.astype` *may* want to support ExtensionArray[T] -> ExtensionArray[S] conversions.

TomAugspurger:
It'll be good to add this to the ecosystem page at http://pandas.pydata.org/pandas-docs/version/0.24/ecosystem.html#extension-data-types once this is ready.

datashader/datatypes.py
jonmmease: Yes that makes good sense, I'll take a look at how the other extension arrays handle this.

datashader/datatypes.py
jonmmease: I was initially thinking that the `dict` form would be private so that there would be an internal fast-path for constructing RaggedArrays without going through the nested array representation.  But I went ahead and made the `mask`, `flat_array`, and `start_indices` properties public so I might as well document the `dict` constructor form so that other libraries can use the fast-path as well.

datashader/datatypes.py
jonmmease: Silly answer: Copy paste error 😊

datashader/datatypes.py
jonmmease: Yeah, I agree that is should be `array([[1., 2.]])`, I'll take a look at what's going wrong.

datashader/datatypes.py
jonmmease: Interesting. Yeah this should at least support `RaggedArray[T]` -> `RaggedArray[S]` to support casting the datatype of the underlying `flat_array`.

If it's some other extension type, would it be reasonable to default to passing `np.array(self)` to the constructor of the other extension array type?

jonmmease:
@TomAugspurger  Ok, cool. Thanks for pointing out the additional test cases that are available.  I'll try them out and report back.  I'll also look over the performance optimization overrides and see what makes sense there.  I'd also like to work out a way to avoid converting everything to tuples for factorization/sorting.

Regarding testing  dask/dask#4379, definitely!  I was actually just starting to look at what it would take to get this working in Dask, and it didn't seem straightforward 🙂  I'd be happy to try this out when you say the word.

datashader/datatypes.py
TomAugspurger: > If it's some other extension type, would it be reasonable to default to passing np.array(self) to the constructor of the other extension array type?

Pandas doesn't make any assumptions about any subclasses `ExtensionArray.__init__`. If you want to, something like

```python
dtype = pandas_dtype(dtype)

if is_extension_array_dtype(dtype):
    dtype.construct_array_type()._from_sequence(np.asarray(self))
```

should work for any ExtensionArray that can handle your scalar type (which may not be many, but best to let them error if they can't handle it.

FYI, there's discussion on this here: https://github.com/pandas-dev/pandas/issues/22384 If you have any thoughts, it'd be good to hear them.

TomAugspurger:
> I'd also like to work out a way to avoid converting everything to tuples for factorization/sorting.

Yes, I'll try to think of something here has well. Does data shader have a hard dependency on numba? That may be helpful here.

jonmmease:
Yes, datashader has a hard dependency on numba.

One approach I was considering was to have `_values_for_factorize` return a list of instances of a lightweight class that holds a reference to the `flat_array`, a start index, and a length. These classes would assume that `flat_array` is immutable during their lifetime and would provide `__hash__`, `__lt__`, etc.

jbednar:
> Not sure about sorting either. How do you define it? Lexicographic?

I can't think of any meaningful way to sort a ragged array column; seems like they will generally be sorted based on other columns (e.g. by a name column).

datashader/datatypes.py
```diff
@@ -0,0 +1,364 @@
...
+                                         if v is not None])
+
+            # Initialize representation arrays
+            self._mask = np.zeros(index_len, dtype='bool')
+            self._start_indices = np.zeros(index_len, dtype=start_indices_dtype)
+            self._flat_array = np.zeros(buffer_len, dtype=dtype)
+
+            # Populate arrays
+            next_start_ind = 0
+            for i, array_el in enumerate(data):
```
jbednar: Should this constructor be numba-ized?

jonmmease: At this point I expect that `data` will be a list of lists or a numpy array of numpy arrays (with `dtype='object'`), so I didn't expect numba to be able to do much with it (at least I didn't think I could use numba in `nopython` mode).

jbednar:
Looks good to me!  A few questions:

- Can it accept the NaN-separated data currently used for variable-length data in Datashader as an input?
- Is the flattened representation usable directly like the current NaN-separated data, or does that code need updating now?
- Does the mask array really need to be a separate data structure, or could masking be done with a special index value to avoid the separate array and handling of it?  I can't immediately see how to support missing values without the mask, but it just seems like there should be a way to avoid it.
- How easy will it be to have this code graduate out of Datashader and into a standalone project later, once we are happy with it?  Ragged array support seems useful for a much wider audience than just Datashader, e.g. for any GeoPandas user.

jonmmease:
> Can it accept the NaN-separated data currently used for variable-length data in Datashader as an input?

Not at the moment, but I don't think it would be too hard to add an alternative `RaggedArray.from_nan_separated` static factory method.  Probably with something like a `consecutive_nans` property that would control whether back-to-back `nan`s should be interpreted as `None`, `[]`, or if they should be treated as a single `nan`

> Is the flattened representation usable directly like the current NaN-separated data, or does that code need updating now?

The code will need to be updated at least a little bit since this flattened representation won't have the `nan`s in it.  The general flow of scanning all the way down a pair of long arrays will be the same, we'll just need an alternative check for when to break the line.   

> Does the mask array really need to be a separate data structure, or could masking be done with a special index value to avoid the separate array and handling of it?

Here are two approaches I considered that don't include a separate mask.

 1. Use a special sentinel value in the index array to represent missing.  Since I'm using unsigned integers for the `start_indices` array, this sentinel value could be the max value for the unsigned integer type.  This would work, the one downside is that looking up ragged element slices from the flattened array becomes more complicated.  With this approach, it's no longer possible to determine the end index of a slice by looking at the next element of `start_indices` since that element might be the sentinel value.  Looking up a slice would involve iterating forward along the `start_indicies` array until finding a non-sentinel value.  It's probably not likely, but the worst case for this scenario would essentially turn the single linear scan of the flattened array into an N^2 scan.

 2. Rather than a sentinel value, represent missing values as the start_index of the next element times -1.  This would get rid of the worst-case described above.  But it would require changing the type of the `start_indices` array from an unsigned to a signed integer, doubling the storage requirement of the `start_indices` array.

So I was concerned about hitting the worst case in scenario (1), and scenario (2) felt like a hack that would generally be less space efficient than adding a 1-byte mask array.

> How easy will it be to have this code graduate out of Datashader and into a standalone project later, once we are happy with it? Ragged array support seems useful for a much wider audience than just Datashader, e.g. for any GeoPandas user.

So far it would be very easy.  We just need to avoid introducing dependencies back on the rest of the Datashader project.

datashader/datatypes.py
TomAugspurger: Ah, yes it should indeed be a length-1 RaggedArray.

jbednar:
> The code will need to be updated at least a little bit since this flattened representation won't have the nans in it. 

What are the pros and cons of including the nans in it?

For the mask, would it not be sufficient to have a zero-length array to indicate a missing element?  Or are you concerned that it's important to be able to distinguish between "missing ragged item" and "empty ragged item"?

datashader/datatypes.py
jonmmease: Is there a non-private import for `is_extension_array_dtype`?  I'm only finding it through `pandas.core.dtypes.common`

TomAugspurger: Sorry. https://github.com/pandas-dev/pandas/pull/24771, so `pandas.core` will have to do for the RC.

jonmmease:
> What are the pros and cons of including the nans in it?
  - **Pro**: Datashader could use it without modification 🙂
  - **Con**: I think the only con is the storage of one extra scalar per ragged element (Even with the nan's there I think we would still need something like `start_indices` to efficiently locate the start of a particular element).  There's also the question of what to do with integers, although perhaps Datashader doesn't need these.

> For the mask, would it not be sufficient to have a zero-length array to indicate a missing element? Or are you concerned that it's important to be able to distinguish between "missing ragged item" and "empty ragged item"?

Yeah, I was assuming the need to distinguish between empty and missing. Though I suppose Datashader doesn't really need this distinction for lines/polygons.  Without this distinction, you're right that `start_indices` could encode missing/empty without a mask.

datashader/datatypes.py
jonmmease: Fixed in 5f46b8e85d6ddac4157ed4b047ed72a079b1aeab

TomAugspurger:
Without having thought about it deeply, I would recommend just using an empty array to represent missing.

jonmmease:
Great, I'm fine pulling out the mask array then.

jonmmease:
Ok, so I tried out the rest of the test base classes from https://github.com/pandas-dev/pandas/blob/master/pandas/tests/extension/decimal/test_decimal.py, and there are quite of lot of failures.  Here are some failure modes I noticed:

 1. Tests using `==` to compare `ExtensionArray` scalars fail on numpy arrays
 2. Lots of internal pandas broadcast/validation errors.

I'm starting to think that for this to be a completely well-behaved ExtensionArray, we'll need the "scalar" elements to not be raw numpy arrays.  Perhaps there could be a `RaggedElement` class that wraps a slice of `flat_array` and provides the magic methods that pandas needs for hashing, equality, and comparing along with something like a `.array` property to get the underlying numpy array slice out.

Any thoughts on this direction?

datashader/datatypes.py
jbednar: Ok.

datashader/datatypes.py
```diff
@@ -0,0 +1,438 @@
...
+
+def _validate_ragged_properties(data):
+    """
+    Validate that dict contains the necessary properties to construct a
+    RaggedArray.
+
+    Parameters
+    ----------
+    data: dict
+        A dict containing 'start_indices', and 'flat_array' keys
```
jbednar:
```suggestion
        A dict containing 'start_indices' and 'flat_array' keys
```

jbednar:
> Any thoughts on this direction?

Hopefully @TomAugspurger can weigh in, but it sounds reasonable to me as a way to support things Pandas needs.

TomAugspurger:
> Tests using == to compare ExtensionArray scalars fail on numpy arrays

Can you show an example of this?

TomAugspurger:
I notice that `__eq__` isn't behaving correctly right now

```python
In [16]: ra = RaggedArray([[1, 2], [], [10, 20], None, [11, 22, 33, 44]])

In [17]: ra == [1, 2]
Out[17]: False
```

I'm not sure of the exact semantics, but that should presumably be broadcast to the same shape as `ra`, where each elem is `np.array([1, 2])`, and then the equality should proceed from there. So the expected output might be

```python
np.array([True, False, False, False, False]])
```

You may want to implement `__array_ufunc__` and inherit from `NDOperatorMixin`: https://docs.scipy.org/doc/numpy-1.15.1/reference/generated/numpy.lib.mixins.NDArrayOperatorsMixin.html. That should make `__eq__` work, and hopefully fix some issues.

jonmmease:
> Can you show an example of this?

Sure, when I add the `Getitem` tests to `test_datatypes.py` with 

```
class TestRaggedGetitem(BaseGetitemTests):
    pass
```

The `test_take_sequence` test in `/pandas/tests/extension/base/getitem.py` fails with 
```
ValueError: The truth value of an array with more than one element is ambiguous. Use a.any()
```

Here are the lines from that test:

https://github.com/pandas-dev/pandas/blob/0bd454cdc9307d3a7e73403c49cc8350965628ce/pandas/tests/extension/base/getitem.py#L154-L158

So it was seeming to me that having `==` on the element type of a `RaggedArray` not return a `bool` is going to cause trouble.  I'm pretty sure I also hit this inside pandas itself at one point, but I don't recall that example at the moment.

I've made some more progress on the test suite by adding a `RaggedElement` class as the element type of `RaggedArray`.  I'll push that work up later today.

> You may want to implement __array_ufunc__ and inherit from NDOperatorMixin

Thanks for pointing that out. Yeah that makes sense.  I'll take a look.

TomAugspurger:
I see, thanks.

That test was indeed written with the assumption that the result were scalars. I see a few options

1. Make the scalars of your class some thing wrapper around an `ndarray`. I'm not sure this is the best from a user point-of-view, but it may not be too bad.
2. Update the pandas tests to replace uses of `x == y` with `self.assert_scalar_equal(x, y)`. By default that would be something like `assert isinstance(x, dtype.type) and x == y`, but subclasses could override that as part of the tests.
3. Just add skips / override the failing tests here.

I believe another group is looking into making a kind of JSONArray, so pandas will be maturing to support nested data in time. But in the interest of "getting things working", a `RaggedElement` may be the way to go for now.

jbednar:
Option 2 there seems like the best long-term approach to me, and it seems like as long as we used a version with that patch for our own Pandas testing, it would be ok if it's some time before it makes it into the official Pandas release.

jonmmease:
Ok, thanks for the feedback and discussion @jbednar and @TomAugspurger .  I just pushed up a new version that I'm feeling more comfortable with.

I did add a `RaggedElement` wrapper class, but I only use it internally for sorting, grouping, joining, etc.  So to the user it will still look like a column of ndarrays.

I also incorporated as many of the applicable pandas ExtensionArray tests as I could find.  For tests that needed a small adjustment to handle ndarrays element types, I copied them into the subclass and made the change.  I also skipped a few tests where pandas had internal exceptions when dealing with ndarray scalars. So @TomAugspurger , the test methods that I overrode should be a good summary of what i ran into.  Overall the test suite was *super* helpful, definitely uncovered lots of edge cases!

Apart from general review/cleanup, here are the three remaining chunks that I think are needed:
 - [ ] Parametrize RaggedDtype by the numpy element type
 - [ ] Add `__array_ufunc__` and inherit from `NDOperatorMixin` to support vectorized `__eq__` etc. 
 - [ ] Remove creation of `tuple` objects in the `RaggedElement` internal class.  Right now they are used as an easy way to implement `__hash__` and `__lt__`.  But these should be replaced with functions operating directly on the array buffer without copying anything.

jbednar:
Sounds great, including the three to-do items!

datashader/__init__.py
```diff
@@ -15,6 +15,9 @@
 except ImportError:
     pass
 
+# Make ragged pandas extension array available
+from . import datatypes
```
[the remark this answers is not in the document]
jbednar: Good question.  No decision on that for now; may just try/except then abort in this particular file with a message saying to upgrade pandas; probably no reason to force *everyone* to upgrade their pandas unless pandas has become incompatible with the rest of Datashader code.

datashader/datatypes.py
```diff
@@ -0,0 +1,583 @@
...
+        with numpy array values
+
+    Raises
+    ------
+    ValueError:
+        if input contains invalid or incompatible properties
+    """
+
+    # Validate start_indices
+    start_indices = data['start_indices']
```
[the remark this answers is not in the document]
jonmmease: Cleaned up in 939405b8fcbe93dec4a9a5b976a4cd3bfed805d7

datashader/datatypes.py
```diff
@@ -0,0 +1,583 @@
...
+            raise TypeError("Cannot construct a '{}' from '{}'"
+                            .format(cls, string))
+
+
+def missing(v):
+    return v is None or (np.isscalar(v) and np.isnan(v))
+
+
+class RaggedArray(ExtensionArray):
+    def __init__(self, data, dtype=None):
```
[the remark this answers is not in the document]
jonmmease: Thanks, that's a good idea. Added in 7f355d2c1430538a180e9e72dbb75ef7228ffb68

datashader/datatypes.py
```diff
@@ -0,0 +1,583 @@
...
+
+            * ndarray: A 1-d boolean NumPy ndarray the same length as 'self'
+        """
+        if isinstance(item, Integral):
+            if item < -len(self) or item >= len(self):
+                raise IndexError("{item} is out of bounds".format(item=item))
+            else:
+                # Convert negative item index
+                if item < 0:
+                    item = len(self) + item
```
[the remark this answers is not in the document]
jonmmease: 👍 9e449468fe114c17bed911e41fdd54013054f3c2

datashader/datatypes.py
```diff
@@ -0,0 +1,583 @@
...
+    def astype(self, dtype, copy=True):
+
+        dtype = pandas_dtype(dtype)
+        if isinstance(dtype, RaggedDtype):
+            if copy:
+                return self.copy()
+            return self
+
+        elif is_extension_array_dtype(dtype):
+            dtype.construct_array_type()._from_sequence(np.asarray(self))
```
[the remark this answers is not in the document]
jonmmease: 👍  a52728a2a5f7e465a7d84ff3adffcb3ee7330d20

datashader/datatypes.py
```diff
@@ -0,0 +1,583 @@
...
+            Construct for this particular dtype. This should be a Dtype
+            compatible with the ExtensionArray.
+        copy : boolean, default False
+            If True, copy the underlying data.
+
+        Returns
+        -------
+        RaggedArray
+        """
+        return RaggedArray(scalars)
```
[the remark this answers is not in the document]
jonmmease: Right now the `dtype` passed to the constructor is intended to be the element `dtype`.  Once the `RaggedDtype` is parameterized by element `dtype` then the constructor can also accept `RaggedDtype`s and we can pass this through.

jonmmease:
In 75f914d10541277c93553dc67c67ff47cd3e9cac I updated the `RaggedDtype` to be parameterized by the element datatype.  So it's now possible to
specify the element datatype in the ragged dtype string.  E.g.
 
```
>>> pd.Series([[1, 2], [2, 3, 4], None], dtype='Ragged[uint16]')
Out[13]: 
0      [1 2]
1    [2 3 4]
2        NaN
dtype: Ragged[uint16]
```
When the subtype is not specified, it defaults to `float64`

```
>>> pd.Series([[1, 2], [2, 3, 4], None], dtype='Ragged')
Out[14]: 
0       [1. 2.]
1    [2. 3. 4.]
2           NaN
dtype: Ragged[float64]
```

datashader/__init__.py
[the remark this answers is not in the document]
jonmmease: I added a guard for pandas 0.24.0 in 3fda786

datashader/datatypes.py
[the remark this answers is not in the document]
jonmmease: Done in 75f914d

datashader/datatypes.py
jonmmease: Done in 75f914d

jonmmease:
I've finished the last batch of updates that I was planning for the `RaggedArray` extension type.  It looks like we're not going to be integrating this in Datashader *quite* as soon as we thought, so this PR may hang out for a while.

@TomAugspurger thanks so much for the review and help so far. Any other feedback you have is more than welcome, and let me know when you're ready for me to try this against your Dask DataFrame extension array work.

TomAugspurger:
Sounds good, I'll give it another look tomorrow.

For Dask things, https://github.com/dask/dask/pull/4379 is in master now.

```python
In [39]: from datashader.datatypes import RaggedArray, RaggedDtype

In [40]: from dask.dataframe.extensions import make_array_nonempty

In [41]: @make_array_nonempty.register(RaggedDtype)
    ...: def _(dtype):
    ...:     return RaggedArray([[1], [1, 2]], dtype=dtype)
    ...:

In [42]: ser = pd.Series([[1, 2], [2, 3, 4], None], dtype='Ragged[uint16]')
    ...: dser = dd.from_pandas(ser, 2)

In [43]: dser
Out[43]:
Dask Series Structure:
npartitions=1
0    Ragged[uint16]
2               ...
dtype: Ragged[uint16]
Dask Name: from_pandas, 1 tasks
```

Is a start. Something things involving scalars don't work with dask.dataframe yet, and may be somewhat challenging.

jonmmease:
RaggedArray line rendering rendering added!

```python
import pandas as pd
import numpy as np
from datashader import Canvas
import datashader as ds

df_ragged = pd.DataFrame({
   'A1': pd.array([
       [1, 1.5], [2, 2.5, 3], [1.5, 2, 3, 4], [3.2, 4, 5]],
       dtype='Ragged[float32]'),
   'B1': pd.array([
       [10, 12], [11, 14, 13], [10, 7, 9, 10], [7, 8, 12]],
       dtype='Ragged[float32]'),
   'group': pd.Categorical([0, 1, 2, 1])
})
df
```
screen shot 2019-02-08 at 8 44 24 am

Set `axis=1` to aggregate one variable length line per row
```python
agg = cvs.line(df_ragged, x='A1', y='B1', axis=1)
tf.spread(tf.shade(agg))
```
download

Now use `count_cat` aggregation on the `group` column to color the variable length line segments
```python
agg = cvs.line(df_ragged, x='A1', y='B1', agg=ds.count_cat('group'), axis=1)
tf.spread(tf.shade(agg))
```
download 1

In terms of documentation, the two examples above have been added to the `Canvas.line` docstring, but I have not made any updates to the documentation notebooks.

@jbednar Pending CI tests, this is ready for review.

cc @TomAugspurger

jonmmease:
Added optimized Dask auto-range calculation logic consistent with https://github.com/pyviz/datashader/pull/717

datashader/datatypes.py
```diff
@@ -0,0 +1,1018 @@
...
+
+    def shift(self, periods=1, fill_value=None):
+        # type: (int, object) -> ExtensionArray
+        """
+        Shift values by desired number.
+
+        Newly introduced missing values are filled with
+        ``self.dtype.na_value``.
+
+        .. versionadded:: 0.24.0
```
jbednar: 0.24.0 is a pandas version, not datashader, right?  Presumably not supposed to be listed here or elsewhere below?

datashader/datatypes.py
```diff
@@ -0,0 +1,1018 @@
...
+        Returns
+        -------
+        RaggedArray
+        """
+        return RaggedArray(scalars, dtype=dtype)
+
+    @classmethod
+    def _from_factorized(cls, values, original):
+        """
+        Reconstruct an ExtensionArray after factorization.
```
jbednar:
```suggestion
        Reconstruct a RaggedArray after factorization.
```

?

datashader/datatypes.py
```diff
@@ -0,0 +1,1018 @@
...
+
+        Parameters
+        ----------
+        values : ndarray
+            An integer ndarray with the factorized values.
+        original : RaggedArray
+            The original ExtensionArray that factorize was called on.
```
jbednar:
```suggestion
            The original RaggedArray that factorize was called on.
```

?

datashader/datatypes.py
```diff
@@ -0,0 +1,1018 @@
...
+        return self._as_ragged_element_array()
+
+    def unique(self):
+        """
+        Compute the ExtensionArray of unique values.
+
+        Returns
+        -------
+        uniques : ExtensionArray
+        """
```
jbednar: I would prefer not to copy docstrings unmodified from the parent class, whether trivial (as here) or complicated (below).  Basically, if the parent class defines the semantics, I want the reader to refer to the parent class, not to this possibly outdated copy of the docstring; that way people know to go find it in the parent, rather than thinking this actually covers everything.  Conversely, if there is a docstring here, I think it should be customized to just be about RaggedArray.

datashader/datatypes.py
```diff
@@ -0,0 +1,1018 @@
...
+            NaN values to forward/backward fill. In other words, if there is
+            a gap with more than this number of consecutive NaNs, it will only
+            be partially filled. If method is not specified, this is the
+            maximum number of entries along the entire axis where NaNs will be
+            filled.
+
+        Returns
+        -------
+        filled : ExtensionArray with NA/NaN filled
+        """
```
jbednar: This docstring seems just copied from the parent class, but if there are differences in behavior from ExtensionArray, please describe those here and refer to the parent class for anything else.

datashader/datatypes.py
```diff
@@ -0,0 +1,1018 @@
...
+
+    Returns
+    -------
+    mask: ndarray
+        1D bool array of same length as input with elements True when
+        corresponding elements are equal, False otherwise
+
+    Notes
+    -----
+    This function is not numba accelerated because it, but design, inputs
```
jbednar:
```suggestion
    This function is not numba accelerated because it, by design, inputs
```

datashader/datatypes.py
```diff
@@ -0,0 +1,1018 @@
...
+        flat_array property of a RaggedArray
+    a: ndarray
+        A 2D numpy array where the length of the first dimension matches the
+        length of the RaggedArray
+
+    Returns
+    -------
+    mask: ndarray
+        1D bool array of same length as input RaggedArray with elements True
+        when corresponding elements of ra equals corresponding row of a
```
jbednar:
```suggestion
        when corresponding elements of ra equal corresponding row of `a`
```

datashader/datatypes.py
jonmmease: Removed in 1538909

datashader/datatypes.py
jonmmease: Removed in 1538909

datashader/datatypes.py
jonmmease: Updated in 1538909

datashader/datatypes.py
```diff
@@ -123,17 +123,26 @@ class RaggedDtype(ExtensionDtype):
 
     @property
     def name(self):
+        """
+        See docstring for ExtensionDtype.name
+        """
```
jbednar: With most docs/API tools, docstrings will simply be inherited as-is if you don't specify one here, so please remove these altogether unless they need to say something explicitly about how they relate to the parent class.  You can mention the parent class explicitly in the class docstring, once, with something like "Methods not otherwise documented here are inherited from ExtensionDtype; please see the corresponding method on that class for the docstring".

jbednar:
Looks good; thanks! Happy to merge once the merge conflict is addressed and the docstring stubs are removed as indicated above.

jonmmease:
Thanks for the review, should be done now

PR #688, opened 2019-01-14
Fix for the gerrymandering shape file

jlstevens:
The original link to the shape file source for the gerrymandering notebook was not reliable (see [this commit](https://github.com/pyviz/datashader/commit/5d98404c1cd093357072a8c6291acb9bfb0d9adb)) but the change introduced there did not include all the necessary files. This PR now moves the entire zip file to S3 which should be more reliable and include all the files.

Not quite ready to merge as I need to upload the zip (waiting on access permissions) at which point I'll test it is working in a clean environment.


jlstevens:
I've also removed entires that are now unused in `dataset.yml`.

jlstevens:
I've added the zip file to the bucket. I'll test now and if it works, this PR is ready to review/merge.

jlstevens:
I can confirm that the gerrymandering example is now working in a fresh environment / datashader install.

PR #689, opened 2019-01-15
Clip composite pixel values to fit into int8

jbednar:
Right now, the "add" compositor is essentially unusable, because the various channels overflow into each other if any pixel values total more than 255 after adding:

```
import pandas as pd, numpy as np, datashader as ds, datashader.transfer_functions as tf
from colorcet import fire
num=10000
np.random.seed(1)

dists = {cat: pd.DataFrame.from_items([('x',np.random.normal(x,s,num)), 
                                       ('y',np.random.normal(y,s,num)), 
                                       ('val',val), 
                                       ('cat',cat)])      
         for x,  y,  s,  val, cat in 
         [(  2,  2, 0.03, 10, "d1"), 
          (  2, -2, 0.10, 20, "d2"), 
          ( -2, -2, 0.50, 30, "d3"), 
          ( -2,  2, 1.00, 40, "d4"), 
          (  0,  0, 3.00, 50, "d5")] }

df = pd.concat(dists,ignore_index=True)
df["cat"]=df["cat"].astype("category")

cvs = ds.Canvas()
img = tf.shade(agg, name="Original image", cmap=fire)

tf.Images(img, *[tf.spread(img, px=7, shape='circle', how=how, name="Circle "+how) 
                 for how in ['source', 'over', 'saturate', 'add']])
```
image

Here there is no green value in the colormap, so any such values visible are entirely errors. This PR adds clipping to 0..255 before the channel values are packed into the 32-bit int to ensure that values only increase up to the limit:

image

Adding now behaves reasonably in this case.  However, note that `add` is only really useful for black-background colormaps increasing in intensity monotonically.  E.g. with a reversed colormap values are additively driven toward the background color, which is the opposite of what a user would want:

image

But at least it's no longer generating colors outside of the gamut from this colormap!

### DIFF
--- a/datashader/composite.py
+++ b/datashader/composite.py
@@ -1,9 +1,14 @@
+"""
+Binary graphical composition operators
+
+See https://www.cairographics.org/operators/; more could easily be added from there.
+"""
+
 from __future__ import division
 
 import numba as nb
 import numpy as np
 import os
-
 
 __all__ = ('composite_op_lookup', 'over', 'add', 'saturate', 'source')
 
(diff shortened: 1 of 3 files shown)


jbednar:
This problem wasn't previously detected, because the tests were encoding the incorrect (binary wrapped) values.  (They were also declaring FFFFFF to be black, not white, but that would only matter to a human reader!)

PR #691, opened 2019-01-18
Changed OpenSky notebook to use Parquet

jbednar:
This notebook originally used HDF5, but since it was written Parquet became a viable format, and it's appropriate here because it's all columnar data and thus doesn't particularly need HDF5.  The resulting opensky.parq file is now 200MB instead of the 600MB opensky.h5 that it replaces, though some of that savings may be from storing a smaller set of columns than before.  

This notebook still needs a good bit of updating, to use HoloViews rather than the outdated InteractiveImage. It's also annoying that the datatypes of some of the columns have to be converted after loading the file, but it wasn't clear how to save it with the types needed here (thanks to @jsignell for help debugging that).


jbednar:
The opensky.parq file is also now allowed to be distributed, by special permission from the OpenSky network (email to jbednar dated Jan 20, 2019), so I've added that to [datashader's s3 bucket](https://s3.amazonaws.com/datashader-data/opensky.parq) and to the datasets.yml file.

PR #692, opened 2019-01-21
Fix variable name collision

jacobtomlinson:
Rename `type` to `typ` as `type` is an existing python method. This follow the scheme of the method below which also uses `typ`.

Although this collision doesn't look to have caused any problems because the `type` function is not used within the method it could potentially be an issue in future and also makes debugging difficult as the code is harder to follow.

Looks like my IDE has also stripped some trailing whitespace. An unintended consequence but is good for PEP8 reasons :).

### DIFF
--- a/datashader/utils.py
+++ b/datashader/utils.py
@@ -37,15 +37,15 @@ class Dispatcher(object):
     def __init__(self):
         self._lookup = {}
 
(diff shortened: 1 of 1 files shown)


jbednar:
Looks good, thanks!

PR #694, opened 2019-01-22
Support for row-oriented coordinates in Canvas.line

jonmmease:
## Overview
This PR introduces a new `Canvas.lines` method for rendering collections of polylines (here called "lines" for short) where each polyline corresponds to a single row in an input DataFrame.  This is in contrast to the current `Canvas.line` method which renders a single polyline (potentially containing gaps if there are nan values) per DataFrame.

I propose that this method eventually be overloaded to support the following 3 use-cases. Note: This PR only implements (1), but I'm hopeful that the addition of (2) and (3) will not require much additional refactoring.

 1. Each line has the same length, `N`, and is represented by `N` columns containing the x coordinates and `N` columns containing the y coordinates.  Here's an example in the case of two-vertex line segments (N=2):
```python
cvs.lines(df, x=['x0', 'x1'], y=['y0', 'y1'])
```

2. Each line has the same length, `N`, but either the x or y vertex coordinates are constant for all lines.  A common use-case here would be time-series data, where the x values (the time bins) are constant across all lines and each row of the DataFrame contains only the y-coordinates of each line. This would nicely handle cases like the one brought up in https://github.com/pyviz/datashader/issues/286#issuecomment-281896685, without the need to concatenate and nan separate each line.  Here's what the aggregation step for this example could look like:

```python
data = ... # 10000 x 50 array representing 10000 lines over 50 time bins
cvs.lines(pd.DataFrame(data), y=list(range(50)), x_constants=np.linspace(0, 10, 50))
```

Note that the `data` numpy array is converted into a DataFrame before being passed to `cvs.lines`.  This DataFrame will have columns with integer labels that match the column index, and so the `y=list(range(50))` argument is specifying a list of the column labels where the column labels are integers.  As suggested in https://github.com/pyviz/datashader/issues/283, it would also be nice to allow DataShader to accept a numpy (or Dask) array directly without requiring this DataFrame conversion step.  I see this as a somewhat orthogonal feature that would require support across several other parts of the API.

3. Each line may have a different length.  In this case the x and y vertex coordinates could be stored in a pair of RaggedArray columns (https://github.com/pyviz/datashader/pull/687).  

**Additional consideration:**  For cases 1 and 2, we should also have a way to specify whether lines are
represented by rows (as I originally had in mind writing the descriptions above) or by columns.  Representing lines by rows is useful when the line lengths are manageable, but there are lots and lots of lines.  Specifying lines by columns is useful when the number of lines is manageable, but each line has lots and lots of points.

## Example
Drawing line segments

```python
import numpy as np
import pandas as pd
import dask.dataframe as dd
import datashader as ds
import datashader.utils as du, datashader.transfer_functions as tf

n = 100
np.random.seed(2)

x0 = np.random.uniform(size=n)
x1 = np.random.uniform(size=n)
y0 = np.random.uniform(size=n)
y1 = np.random.uniform(size=n)

pts = np.stack((x0,x1,y0,y1)).T
df = pd.DataFrame(np.stack((x0,x1,y0,y1)).T, columns=['x0', 'x1', 'y0', 'y1'])

cvs = ds.Canvas(plot_height=400,plot_width=400, x_range=[0, 1], y_range=[0, 1])
tf.shade(cvs.lines(df, ['x0', 'x1'], ['y0', 'y1'], agg=ds.count()))
```
download

## Performance

**Edit:** See updated performance results in https://github.com/pyviz/datashader/pull/694#issuecomment-456607889

Here are some performance comparisons comparing `cvs.lines` with the equivalent `cvs.line` command for the segment example above. Timing does not take into account the time to generate the nan separated DataFrame needed by `cvs.line`.

newplot 8

And here's what scaling looks like on my quad-core 2015 MBP.
newplot 9

So the current `cvs.line` approach is consistently a *little* bit faster in the single-threaded case. But if the conversion time were taken into account this difference would probably get washed out.  And for some reason the new `cvs.lines` approach seems to scale a little better in the multi-threaded case (at least for 3 and 4 cores).

## TODO
 - [x] Evaluate unifying this functionality with the current `cvs.line` method by adding a new "lines-by-rows-or-columns" argument.  Either way, sketch out the future API for the by-rows and by-columns cases.
 - [x] Rewrite last section of http://datashader.org/user_guide/3_Timeseries.html

### DIFF
--- a/datashader/glyphs.py
+++ b/datashader/glyphs.py
@@ -234,9 +234,12 @@ class Line(_PointLike):
 
 
 class LinesXY(_PointLike):
+
     def validate(self, in_dshape):
-        # TODO
-        pass
+        if not all([isreal(in_dshape.measure[xcol]) for xcol in self.x]):
+            raise ValueError('x columns must be real')
+        elif not all([isreal(in_dshape.measure[ycol]) for ycol in self.y]):
+            raise ValueError('y columns must be real')
 
     def required_columns(self):
         return self.x + self.y
--- a/datashader/glyphs.py
+++ b/datashader/glyphs.py
@@ -234,7 +234,14 @@ class Line(_PointLike):
 
 
 class LinesXY(_PointLike):
+    """A collection of lines (on line per row) with vertices defined
+    by the lists of columns in ``x`` and ``y``
 
+    Parameters
+    ----------
+    x, y : list
+        Lists of column names for the x and y coordinates
+    """
     def validate(self, in_dshape):
         if not all([isreal(in_dshape.measure[xcol]) for xcol in self.x]):
             raise ValueError('x columns must be real')
@@ -283,10 +290,8 @@ class LinesXY(_PointLike):
         minval, maxval = np.nanmin(x_mins), np.nanmax(x_maxes)
 
         if minval == np.nan and maxval == np.nan:
-            # print("No x values; defaulting to range -1,1")
             minval, maxval = -1, 1
         elif minval == maxval:
-            # print("No x range; defaulting to x-1,x+1")
             minval, maxval = minval - 1, minval + 1
         return minval, maxval
 
@@ -301,10 +306,8 @@ class LinesXY(_PointLike):
         minval, maxval = np.nanmin(y_mins), np.nanmax(y_maxes)
 
         if minval == np.nan and maxval == np.nan:
-            # print("No x values; defaulting to range -1,1")
             minval, maxval = -1, 1
         elif minval == maxval:
-            # print("No x range; defaulting to x-1,x+1")
             minval, maxval = minval - 1, minval + 1
         return minval, maxval
 
--- a/datashader/glyphs.py
+++ b/datashader/glyphs.py
@@ -17,6 +17,10 @@ class _PointLike(Glyph):
         self.x = x
         self.y = y
 
+    def __hash__(self):
+        return hash(
+            tuple([self.__class__, self.x, self.y]))
+
     @property
     def inputs(self):
         return (self.x, self.y)
@@ -135,6 +139,11 @@ class _PolygonLike(_PointLike):
             self.z = z
         self.interpolate = interp
         self.weight_type = weight_type
+
+    def __hash__(self):
+        return hash(
+            tuple([self.__class__, self.x, self.y, tuple(self.z),
+                   self.interpolate, self.weight_type]))
 
     @property
     def inputs(self):
--- a/datashader/glyphs.py
+++ b/datashader/glyphs.py
@@ -17,10 +17,6 @@ class _PointLike(Glyph):
         self.x = x
         self.y = y
 
-    def __hash__(self):
-        return hash(
-            tuple([self.__class__, self.x, self.y]))
-
     @property
     def inputs(self):
         return (self.x, self.y)
@@ -43,10 +39,12 @@ class _PointLike(Glyph):
         return [self.x, self.y]
 
     def compute_x_bounds(self, df):
-        return self._compute_x_bounds(df[self.x].values)
+        bounds = self._compute_x_bounds(df[self.x].values)
+        return self.maybe_expand_bounds(bounds)
 
     def compute_y_bounds(self, df):
-        return self._compute_y_bounds(df[self.y].values)
+        bounds = self._compute_y_bounds(df[self.y].values)
+        return self.maybe_expand_bounds(bounds)
 
     @staticmethod
     @ngjit
@@ -60,12 +58,6 @@ class _PointLike(Glyph):
                 if x > maxval:
                     maxval = x
 
-        if not (np.isfinite(minval) and np.isfinite(maxval)):
-            #print("No x values; defaulting to range -1,1")
-            minval, maxval = -1.0, 1.0
-        elif minval==maxval:
-            #print("No x range; defaulting to x-1,x+1")
-            minval, maxval = minval-1, minval+1
         return minval, maxval
 
     @staticmethod
@@ -80,11 +72,14 @@ class _PointLike(Glyph):
                 if y > maxval:
                     maxval = y
 
+        return minval, maxval
+
+    @staticmethod
+    def maybe_expand_bounds(bounds):
+        minval, maxval = bounds
         if not (np.isfinite(minval) and np.isfinite(maxval)):
-            #print("No y values; defaulting to range -1,1")
-            minval, maxval = -1, 1
-        elif minval==maxval:
-            #print("No y range; defaulting to y-1,y+1")
+            minval, maxval = -1.0, 1.0
+        elif minval == maxval:
             minval, maxval = minval-1, minval+1
         return minval, maxval
 
@@ -95,14 +90,7 @@ class _PointLike(Glyph):
         """
         xs = df[self.x].values
         minval, maxval = np.nanmin(xs), np.nanmax(xs)
-        
-        if minval == np.nan and maxval == np.nan:
-            #print("No x values; defaulting to range -1,1")
-            minval, maxval = -1,1
-        elif minval==maxval:
-            #print("No x range; defaulting to x-1,x+1")
-            minval, maxval = minval-1, minval+1
-        return minval, maxval
+        return self.maybe_expand_bounds((minval, maxval))
 
     @memoize
     def compute_y_bounds_dask(self, df):
@@ -111,14 +99,7 @@ class _PointLike(Glyph):
         """
         ys = df[self.y].values
         minval, maxval = np.nanmin(ys), np.nanmax(ys)
-        
-        if minval == np.nan and maxval == np.nan:
-            #print("No y values; defaulting to range -1,1")
-            minval, maxval = -1,1
-        elif minval==maxval:
-            #print("No y range; defaulting to y-1,y+1")
-            minval, maxval = minval-1, minval+1
-        return minval, maxval
+        return self.maybe_expand_bounds((minval, maxval))
 
 
 
@@ -140,11 +121,6 @@ class _PolygonLike(_PointLike):
         self.interpolate = interp
         self.weight_type = weight_type
 
-    def __hash__(self):
-        return hash(
-            tuple([self.__class__, self.x, self.y, tuple(self.z),
-                   self.interpolate, self.weight_type]))
-
     @property
     def inputs(self):
         return tuple([self.x, self.y] + list(self.z))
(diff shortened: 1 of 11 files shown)


datashader/glyphs.py
```diff
@@ -444,6 +570,33 @@ def extend_line(vt, bounds, xs, ys, plot_start, *aggs_and_cols):
     return extend_line
 
 
+def _build_extend_lines_xy(draw_line, map_onto_pixel):
+    extend_line = _build_extend_line(draw_line, map_onto_pixel)
+
+    @ngjit
+    def extend_lines_xy(vt, bounds, xs, ys, plot_start, *aggs_and_cols):
```
philippjfr: I've had a go at rewriting this by adapting the ``extend_line`` code and that seems somewhat more efficient than unpacking this format and reusing the existing codepath. Probably needs some more fixes, but here's what my 10 minutes of hacking produced:

```

    @ngjit
    def extend_lines_xy(vt, bounds, xs, ys, plot_start, *aggs_and_cols):
        """
        here xs and ys are tuples of arrays and non-empty
        """
        xmin, xmax, ymin, ymax = bounds
        nrows = xs[0].shape[0]
        ncols = len(xs)
        i = 0
        while i < nrows - 1:
            plot_start = True
            j = 0
            while j < ncols - 1:
                x0 = xs[j][i]
                y0 = ys[j][i]
                x1 = xs[j+1][i]
                y1 = ys[j+1][i]

                # Use Liang-Barsky (1992) to clip the segment to a bounding box
                if outside_bounds(x0, y0, x1, y1, xmin, xmax, ymin, ymax):
                    plot_start = True
                    j += 1
                    continue

                t0, t1 = 0, 1
                dx = x1 - x0

                t0, t1, accept = clipt(-dx, x0 - xmin, t0, t1)
                if not accept:
                    j += 1
                    continue

                t0, t1, accept = clipt(dx, xmax - x0, t0, t1)
                if not accept:
                    j += 1
                    continue

                dy = y1 - y0

                t0, t1, accept = clipt(-dy, y0 - ymin, t0, t1)
                if not accept:
                    j += 1
                    continue

                t0, t1, accept = clipt(dy, ymax - y0, t0, t1)
                if not accept:
                    j += 1
                    continue

                if t1 < 1:
                    clipped = True
                    x1 = x0 + t1 * dx
                    y1 = y0 + t1 * dy

                if t0 > 0:
                    # If x0 is clipped, we need to plot the new start
                    clipped = True
                    plot_start = True
                    x0 = x0 + t0 * dx
                    y0 = y0 + t0 * dy

                x0i, y0i = map_onto_pixel(vt, bounds, x0, y0)
                x1i, y1i = map_onto_pixel(vt, bounds, x1, y1)
                draw_line(x0i, y0i, x1i, y1i, i, plot_start, clipped, *aggs_and_cols)
                plot_start = False
                j += 1
            i += 1

    return extend_lines_xy
```

datashader/glyphs.py
```diff
@@ -206,6 +233,103 @@ def extend(aggs, df, vt, bounds, plot_start=True):
...
+        y_bounds_list = [self._compute_y_bounds(ycol.values) for ycol in ys]
+        y_mins, y_maxes = zip(*y_bounds_list)
+
+        y_min = min(y_mins)
+        y_max = max(y_maxes)
+        return y_min, y_max
+
+    @memoize
+    def compute_x_bounds_dask(self, df):
+        """Like ``PointLike._compute_x_bounds``, but memoized because
```
jbednar:
```suggestion
        """Like ``PointLike.compute_x_bounds``, but memoized because
```

datashader/glyphs.py
```diff
@@ -206,6 +233,103 @@ def extend(aggs, df, vt, bounds, plot_start=True):
...
+
+        if minval == np.nan and maxval == np.nan:
+            minval, maxval = -1, 1
+        elif minval == maxval:
+            minval, maxval = minval - 1, minval + 1
+        return minval, maxval
+
+    @memoize
+    def compute_y_bounds_dask(self, df):
+        """Like ``PointLike._compute_x_bounds``, but memoized because
```
jbednar:
```suggestion
        """Like ``PointLike.compute_x_bounds``, but memoized because
```

datashader/glyphs.py
jbednar: Cool!  @jonmmease, can you benchmark that like you did above?

philippjfr: Using this on a 10M vertex trimesh wireframe using dask this renders in 0.5 seconds as opposed to 1.5 seconds with ``cvs.line``.

jonmmease: Awesome, thanks for giving this a try @philippjfr! I'll incorporate this approach and add it to my benchmarks

jonmmease:
@jbednar for documentation I was thinking of adding a section to http://datashader.org/user_guide/4_Trajectories.html for the case where you have a collection of fixed length trajectories (just make up some more random walks).

Until we have (2) from the description above I think the last section of http://datashader.org/user_guide/3_Timeseries.html would still be a bit awkward because the user would need to create columns for the x-data that would repeat for every row.  With (2) it will be a one-liner without any utility functions or work-arounds needed.

What do you think?

jbednar:
Ok, sounds good.  In that case please briefly mention the new approach in 3_Timeseries, for now, and once approach 2 is implemented that section can be rewritten to use the new way.

BTW, can approach 2 be approximated already in the current codebase simply by passing the y column `n` numbers of times as the `y_label`, for `n` x columns?  It might not be quite as fast, but wouldn't require duplicating the actual column, right?

jonmmease:
> BTW, can approach 2 be approximated already in the current codebase simply by passing the y column `n` numbers of times as the `y_label`, for `n` x columns?  It might not be quite as fast, but wouldn't require duplicating the actual column, right?

I'm not sure, I'll give it another look when I incorporate @philippjfr 's suggestion above. And if something looks relatively straightforward, I'll just go ahead and add it.

jbednar:
Note that I'm just suggesting a (untested) way for the user to use the current code, not a change to the code...

jonmmease:
Updated performance results after implementing optimization suggested by @philippjfr in bc7631c:

newplot 15

newplot 14

Performance is now basically equivalent for single threaded-case and is much faster for multi-threaded case.

jonmmease:
I added a new API consideration to the main description:
> Additional consideration: For cases 1 and 2, we should also have a way to specify whether lines are
represented by rows (as I originally had in mind writing the descriptions above) or by columns. Representing lines by rows is useful when the line lengths are manageable, but there are lots and lots of lines. Specifying lines by columns is useful when the number of lines is manageable, but each line has lots and lots of points.

I need to think about it a little more, but if we have an `axis` style argument that determines whether lines are specified by rows or columns, then we might be able to unify this functionality with the current `cvs.line` method, rather than introduce a new `cvs.lines` method.

jbednar:
It looks very flexible and powerful to me!  

I wonder if there is some way to avoid having to introduce `x_constants` and `y_constants` as different names. I see it's not practical to deduce that `[0.0, 1.0]` is a constant array rather a Pandas column names, since Pandas allows numeric column names.   Could use tuples instead of lists, but that seems arbitrary.  Maybe if it's an array, treat it as a constant axis to broadcast, but otherwise as column names?

jonmmease:
> I wonder if there is some way to avoid having to introduce `x_constants` and `y_constants` as different names

Yeah, I was hoping to avoid the need for `x_constants`/`y_constants` as well. But I've struggled to come up with a natural way to distinguish column labels from numeric vectors as well.  Using a numeric numpy array for this might work out alright, but then I'm not sure what to do with a numpy array of strings or a numeric pandas Series.  Do you have any thoughts @philippjfr ?

jbednar:
I guess to refine my proposal then, an array _or_ Pandas series type would be treated as  a constant numeric vector, while if one has a numpy array of strings, you'd have to convert it to a list of strings for it to be treated as  a list of column names.

jonmmease:
> I guess to refine my proposal then, an array or Pandas series type would be treated as a constant numeric vector, while if one has a numpy array of strings, you'd have to convert it to a list of strings for it to be treated as a list of column names.

Works for me!

jonmmease:
Alright, I went ahead and implemented all of the aggregation approaches in the proposed API except for RaggedArray.

### Examples from the new `Canvas.line` docstring
Define a canvas and a pandas DataFrame with 6 rows
```python
import pandas as pd
import numpy as np
from datashader import Canvas
cvs = Canvas(plot_height=400, plot_width=400)
df = pd.DataFrame({
   'A1': [1, 1.5, 2, 2.5, 3, 4],
   'A2': [1.5, 2, 3, 3.2, 4, 5],
   'B1': [10, 12, 11, 14, 13, 15],
   'B2': [11, 9, 10, 7, 8, 12],
}, dtype='float64')
```

Aggregate one line across all rows, with coordinates `df.A1` by `df.B1`
```python
agg = cvs.line(df, x='A1' , y='B1')
tf.shade(agg)
```
download 1

Aggregate two lines across all rows. The first with coordinates `df.A1` by `df.B1` and the second with coordinates `df.A2` by `df.B2`
```python
agg = cvs.line(df, x=['A1', 'A2'], y=['B1', 'B2'], axis=0)
tf.shade(agg)
```
download 2

Aggregate two lines across all rows where the lines share the same x coordinates. The first line will have coordinates `df.A1` by `df.B1` and the second will have coordinates `df.A1` by `df.B2`.

```python
agg = cvs.line(df, x='A1', y=['B1', 'B2'], axis=0)
tf.shade(agg)
```
download 3

Aggregate 6 length-2 lines, one per row, where the ith line has coordinates `[df.A1[i], df.A2[i]]` by `[df.B1[i], df.B2[i]]`
```python
agg = cvs.line(df, x=['A1', 'A2'], y=['B1', 'B2'], axis=1)
tf.shade(agg)
```
download 4

Aggregate 6 length-4 lines, one per row, where the x coordinates of every line are `[0, 1, 2, 3]` and the y coordinates of the ith line are [df.A1[i], df.A2[i], df.B1[i], df.B2[i]].

```python
agg = cvs.line(df, x=np.arange(4), y=['A1', 'A2', 'B1', 'B2'], axis=1)
tf.shade(agg)
```
download 5

### Documentation
I haven't updated the documentation notebooks yet, but with these changes I'll be able to rewrite the last section of http://datashader.org/user_guide/3_Timeseries.html and remove the use of the `dataframe_from_multiple_sequences` utility function all together.

jbednar:
Looks great!!!  Thanks for rounding that out.  Seems like a very reasonable API now.

jonmmease:
I added the Dask tests, and rewrote the last section of the Timeseries documentation to use the new approach.

Ready for review @jbednar

datashader/glyphs.py
```diff
@@ -59,46 +72,34 @@ def _compute_y_bounds(ys):
                 if y > maxval:
                     maxval = y
 
+        return minval, maxval
+
+    @staticmethod
+    def maybe_expand_bounds(bounds):
```
jbednar: Useful thing to refactor, thanks.

jbednar:
I modified the last example in the user-guide notebook so that it was a random walk with a fixed starting point:

image

This now seems like a very reasonable synthesized approximation to something people will use in practice, i.e. overlaying huge numbers of lines with the same starting point.

jonmmease:
Hmm, hitting an AppVeyor permission denied error on Python 2.7.  Is this the same one we've been fighting with?

screen shot 2019-01-30 at 7 01 58 pm

jonmmease:
Merging so I can go ahead and rebase https://github.com/pyviz/datashader/pull/696.  Thanks for the review and doc updates @jbednar!

PR #696, opened 2019-01-24
Initial Dask trimesh support

jonmmease:
## Overview
This PR provides initial support for parallel aggregation of trimesh glyphs using dask

**Note:** This PR is based on https://github.com/pyviz/datashader/pull/694 as it relies on some of the refactoring performed in that PR.

## Usage
To take advantage of dask trimesh support, the `datashader.utils.mesh` utility function should be called with dask `DataFrame`s for the the `vertices` and `simplices` arguments.  In this case, the resulting `mesh` DataFrame will be a dask DataFrame rather than a pandas DataFrame.

When this dask mesh is passed into the `cvs.trimesh` functino, the trimesh aggregations will be performing in parallel.  For example

```python
# Dask mesh
verts_ddf = dd.from_pandas(verts, npartitions=4)
tris_ddf = dd.from_pandas(pd.concat([tris]*copies, axis=0), npartitions=4)
mesh_ddf = du.mesh(verts_ddf, tris_ddf).persist()

cvs = ds.Canvas(plot_height=900, plot_width=900)
agg = cvs.trimesh(verts_ddf, tris_ddf, mesh_ddf)
```

## Implementation Notes
The job of the `du.mesh` function is to return a DataFrame containing the coordinates of every vertex in every triangle in the mesh in the proper winding order.  A triangle is represented in this data structure by three rows, one for each vertex.  If a single vertex is used by more then one triangle, then the coordinates of that vertex will show up in multiple rows in this DataFrame.

One important characteristic of the updated `du.mesh` function when called with a dask DataFrame is that it makes sure that no triangles straddle a partition boundary in the output.  This amounts to making sure that the number of rows in each partition is a multiple of 3.  The function attempts to build the output dask dataframe with the greater of the number of partitions in `vertices` and `simplices`, but the constraint to avoid breaking up triangles takes precedence over the number of partitions.

The speedup here is only in the call to `cvs.trimesh`, the call to `du.mesh` still requires pulling the `vertices` and `simplices` DataFrames into memory. Parallelizing this step in the calculation will take a bit more thought, and may require some spatial ordering of the input DataFrames.

## Bechmarking
I ran some benchmark tests comparing the pandas aggregation with this new dask aggregation.  These were run on a 2015 MBP with a quadcore processor. All of the dask tests were run using 4 partitions.

To scale the number of triangles I used the Chesapeake Bay mesh (https://github.com/pyviz/datashader/blob/master/examples/topics/bay_trimesh.ipynb), and then duplicated the simplices between 1 and 100 times.  This scales from ~1 million to ~100 million triangles.

**pandas vs dask runtime:**
dask_pandas_time

**dask speedup factor (pandas runtime / dask runtime)**
dask_speedup_factor

So the dask implementation is about 1.2 times faster on 1 million triangles and up the 4.3 times faster on 100 million triangles.

### DIFF
--- a/datashader/glyphs.py
+++ b/datashader/glyphs.py
@@ -123,7 +123,8 @@ class _PolygonLike(_PointLike):
 
     @property
     def inputs(self):
-        return tuple([self.x, self.y] + list(self.z))
+        return (tuple([self.x, self.y] + list(self.z)) +
+                (self.weight_type, self.interpolate))
 
     def validate(self, in_dshape):
         for col in [self.x, self.y] + list(self.z):
--- a/datashader/utils.py
+++ b/datashader/utils.py
@@ -450,18 +450,16 @@ def _dd_mesh(vertices, simplices):
     Dask DataFrame objects.
     """
     # Construct mesh by indexing into vertices with simplex indices
-    # TODO: For dask: avoid .compute() calls, and add winding auto-detection
-    vertex_idxs = simplices.values[:, :3].astype(np.int64)
-    vals = vertices.values.compute()[vertex_idxs]
-    vals = vals.reshape(np.prod(vals.shape[:2]), vals.shape[2])
-    res = pd.DataFrame(vals, columns=vertices.columns)
+    # TODO: For dask: avoid .compute() calls
+    res = _pd_mesh(vertices.compute(), simplices.compute())
 
-    # If vertices don't have weights, use simplex weights
-    verts_have_weights = len(vertices.columns) > 2
-    if not verts_have_weights:
-        weight_col = simplices.columns[3]
-        res[weight_col] = simplices.values[:, 3].compute().repeat(3)
+    # Compute a chunksize that will not split the vertices of a single
+    # triangle across partitions
+    approx_npartitions = max(vertices.npartitions, simplices.npartitions)
+    chunksize = int(np.ceil(len(res) / (3*approx_npartitions)) * 3)
 
+    # Create dask dataframe
+    res = dd.from_pandas(res, chunksize=chunksize)
     return res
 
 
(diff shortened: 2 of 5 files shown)


datashader/glyphs.py
```diff
@@ -123,7 +123,8 @@ def __init__(self, x, y, z=None, weight_type=True, interp=True):
 
     @property
     def inputs(self):
-        return tuple([self.x, self.y] + list(self.z))
+        return (tuple([self.x, self.y] + list(self.z)) +
+                (self.weight_type, self.interpolate))
```
jonmmease: This change was needed because the `inputs` tuple is used by the parent class to implement hashing and equality, which are in tern used for memoization.  Without this, I was seeing cases where repeated use of `canvas.trimesh` with different values for `interpolate` was not resulting in updated aggregation behavior.

datashader/utils.py
```diff
@@ -450,18 +450,16 @@ def _dd_mesh(vertices, simplices):
     Dask DataFrame objects.
     """
     # Construct mesh by indexing into vertices with simplex indices
-    # TODO: For dask: avoid .compute() calls, and add winding auto-detection
-    vertex_idxs = simplices.values[:, :3].astype(np.int64)
-    vals = vertices.values.compute()[vertex_idxs]
-    vals = vals.reshape(np.prod(vals.shape[:2]), vals.shape[2])
-    res = pd.DataFrame(vals, columns=vertices.columns)
+    # TODO: For dask: avoid .compute() calls
+    res = _pd_mesh(vertices.compute(), simplices.compute())
```
jonmmease: We were calling `compute` on both `vertices` and `simplices` anyway, so I opted to just call the pandas versions.  In addition making this more concise, the pandas version has winding auto-detection enabled that was not previously enabled here.

jonmmease:
@jbednar Ready for review.

jonmmease:
@jbednar tests passing!

PR #702, opened 2019-02-07
Add SpatialPointsFrame class to support spatial partitioning for points aggregation

jonmmease:
This PR brings the functionality prototyped in https://github.com/jonmmease/hilbert_frame into the datashader project.

The class `HilbertFrame2D` has been renamed `SpatialPointsFrame` as I think this is a more informative and more precise name.  If we work out a way to support spatial partitioning of lines or trimeshes in the future then we can add additional `Spatial*Frame` classes.

The implementation has been substantially cleaned up and the first three items in https://github.com/jonmmease/hilbert_frame/issues/3 have been addressed

1) The metadata is now stored directly in the parquet file as custom metadata, so the pickle file is no more.
2) The Hilbert distance index column is no longer stored in the resulting parquet file.  So now there is negligible storage overhead associated with storing a data set in this spatially partitioned format.  And, the file can still be loaded directly by pandas/dask/fastparquet/etc. for use in non-spatial contexts.
3) Numpy-style doc strings have been added

4) A test suite has been added

The last remaining item is documentation.

I would appreciate thoughts on how this functionality could be integrated into the existing documentation structure.

I have the three notebooks I posed in https://github.com/jonmmease/hilbert_frame that could be a starting point.  We could also consider distributing the OSM or census parquet datasets in this form and then have the corresponding topics take advantage of the spatial partitioning.

cc @jbednar @philippjfr

### DIFF
--- a/datashader/spatial/__init__.py
+++ b/datashader/spatial/__init__.py
@@ -525,7 +525,7 @@ SpatialPointsFrame.partition_and_write static method.""".format(
         -------
         x_bin_edges: np.ndarray
         """
-        return np.linspace(*self.x_range, self._side_length + 1)
+        return np.linspace(*self.x_range, num=self._side_length + 1)
 
     @property
     def y_bin_edges(self):
@@ -537,7 +537,7 @@ SpatialPointsFrame.partition_and_write static method.""".format(
         -------
         y_bin_edges: np.ndarray
         """
-        return np.linspace(*self.y_range, self._side_length + 1)
+        return np.linspace(*self.y_range, num=self._side_length + 1)
 
     @property
     def distance_divisions(self):
--- a/datashader/spatial/__init__.py
+++ b/datashader/spatial/__init__.py
@@ -1,0 +1,621 @@
+from __future__ import absolute_import
+import copy
+import os
+import shutil
+import json
+
+import numpy as np
+import pandas as pd
+import dask.dataframe as dd
+
+from datashader.spatial import hilbert_curve as hc
+try:
+    import fastparquet as fp
+except ImportError:
+    fp = None
+
+# Declare Python2/3 unicode-safe string type
+try:
+    basestring
+except NameError:
+    basestring = str
+
+
+def _data2coord(vals, val_range, side_length):
+    """
+    Convert an array of values from continuous data coordinates to discrete
+    Hilbert coordinates
+
+    Parameters
+    ----------
+    vals: list or tuple or np.ndarray
+        Array of continuous data coordinate to be converted to discrete
+        distance coordinates
+    val_range: tuple
+        Start (val_range[0]) and stop (val_range[1]) range in continuous
+        data coordinates
+    side_length: int
+        The number of discrete distance coordinates
+
+    Returns
+    -------
+    np.ndarray
+        Of discrete Hilbert coordinates
+    """
+    if isinstance(vals, (list, tuple)):
+        vals = np.array(vals)
+
+    x_width = val_range[1] - val_range[0]
+    return ((vals - val_range[0]) * (side_length / x_width)
+            ).astype(np.int64).clip(0, side_length - 1)
+
+
+def _compute_distance(df, x, y, p, x_range, y_range):
+    """
+    Compute an array of Hilbert distances from a pandas dataframe
+
+    Parameters
+    ----------
+    df: pd.DataFrame
+        pandas dataframe containing the coordinate columns
+    x
+        Label of the column containing the x-coordinates
+    y
+        Label of the column containing the y-coordinates
+    p: int
+        Hilbert-curve p value
+    x_range: tuple
+        Start (x_range[0]) and stop (x_range[1]) range of Hilbert x
+        scale in data coordinates
+    y_range: tuple
+        Start (y_range[0]) and stop (y_range[1]) range of Hilbert y
+        scale in data coordinates
+
+    Returns
+    -------
+    np.ndarray
+        Of Hilbert distances
+    """
+    side_length = 2 ** p
+    x_coords = _data2coord(df[x], x_range, side_length)
+    y_coords = _data2coord(df[y], y_range, side_length)
+    return hc.distance_from_coordinates(p, x_coords, y_coords)
+
+
+def _compute_extents(df, x, y):
+    """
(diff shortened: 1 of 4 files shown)


datashader/spatial/__init__.py
```diff
@@ -0,0 +1,621 @@
...
+            The dask.dataframe.to_parquet compression method.
+        """
+
+        _validate_fastparquet()
+
+        # Validate filename
+        if (not isinstance(filename, basestring) or
+                not filename.endswith('.parquet')):
+            raise ValueError(
+                'filename must be a string ending with a .parquet extension')
```
jbednar: We've been using .parq for the extension, so please allow that convention as well.

datashader/spatial/__init__.py
```diff
@@ -0,0 +1,621 @@
...
+                index=s.index))
+
+    @property
+    def x(self):
+        """
+        Column label in frame containing the x coordinate of each row
+        """
+        return self._x
+
+    @property
```
jbednar: Can you explain what these various properties are doing?  I assume there is a reason not to simply make them attributes?

jonmmease:
> How costly is it to rebuild the index on loading?

I'm not actually rebuilding the full distances index column.  Now I'm just keeping track of the partition divisions (the dask dataframe divisions before the the distance index was dropped), and using these to select the necessary partitions.

The time to construct SpatialPointsFrame scales roughly with 4^p (where p is the Hilbert curve order), but is constant with the length of the dataframe.  This is because I precompute a (2^p) x (2^p) grid of the partition number of every discrete coordinate.

I set the defualt p value to 10, which effectively discretizes each of the x and y coordinates into 1024 bins prior to computing distances.  This works well for the census and OSM data sets, and the load time is still < 1 second (after numba acceleration in a32eae)

datashader/spatial/__init__.py
jonmmease: done in ff82fe1

datashader/spatial/__init__.py
jonmmease: This might be going a bit overboard, but there are two reasons I use this approach:

 1) These properties should be read only, and if a user tries to set one they get an error.
 2) Docstrings added to the getters are visible in the notebook. E.g. `sframe.x?`

jonmmease:
I added `fastparquet` to `extras_require['tests']` so that it is installed during test time.  This is needed in order to use `SpatialPointsFrame`, but it's not required for general datashader use so I didn't make it a full dependency.

Does that sound alright?

jonmmease:
Ok, got the test suite squared away. Merging!

datashader/spatial/__init__.py
jbednar: Ok. Maybe overboard, but shouldn't hurt.

jbednar:
That all sounds good.  Can you open an issue outlining how you added the metadata and what it would take to add pyarrow's parquet support alongside the fastparquet support?

jbednar:
As for documentation, I think this topic fits into the currently empty 2_Points.ipynb user guide.  The as-yet-entirely-unrealized goal for 2_Points was to generalize the parts of nyc_taxi, nyc_taxi_nongeo and census notebooks to provide a full overview of functionality available and techniques to use. So I think that file should include a brief discussion of spatial indexing, with a section like:


## Spatial indexing

In most cases, Datashader must iterate through your entire dataset to render any plot, because it cannot assume the datapoints have been stored in any particular order.  Thus the aggregation performance is dependent on the number of datapoints in your entire dataframe, not just those in the current viewport (x and y range).  If you have a large dataset covering a wide area and you want to support fast local operations (e.g. if you hae data at a global level but analysis is typically done in small local regions), Datashader supports optionally storing your data in a spatially indexed format. This format makes it very fast to create a new dataframe with only the points from a restricted local region, without even needing to bring the other data points into main memory.

For instance, if you start with an unordered dataframe `df`:

... show synthetic data like in http://datashader.org/getting_started/2_Pipeline.html but with maybe 1 billion points ...

You can sort it spatially on `x` and `y` using:

```
from datashader.spatial import SpatialPointsFrame
%time SpatialPointsFrame.from_dataframe(df,'sorted.parq', 'x', 'y', shuffle='disk')
```

This process will take some time, e.g. several minutes for a billion-point dataframe on a 4-core 16GB laptop, writing the resulting ordered dataframe to the given Parquet-format filename (requires `fastparquet` and `python-snappy`). You can now quickly create a regular dataframe with data for the region of interest:

```
sf = SpatialPointsFrame('sorted.parq', persist=True)
df = hframe.range_query(x_range, y_range)
```

The resulting dataframe `df` can then be used just like any other dataframe as described in the above documentation, but now containing only the data in the given x and y range, instead of the full dataset.  For example:

... compare speed between the regular and spatial query of the smallest blob of datapoints in this example...

You can also supply the SpatialPointsFrame to [HoloViews](http://holoviews.org), which will generate the spatial subset dynamically for interactive zooming in a [Bokeh](http://bokeh.pydata.org)-based plot:
    
... same example but interactive ...

See the [SpatialPointsFrame](XXXXX/spatial-points.ipynb) notebook for details of how the spatial dataframe is constructed and for performance comparisons on real-world datasets.



[Where spatial-points.ipynb consists of a HoloViews vesion of [HilbertCurvePartitioningOverview](https://nbviewer.jupyter.org/github/jonmmease/hilbert_frame/blob/master/notebooks/1-HilbertCurvePartitioningOverview.ipynb) plus sample timings like those in [HilbertCurvePartitioningCensusExample.ipynb](https://nbviewer.jupyter.org/github/jonmmease/hilbert_frame/blob/master/notebooks/2-HilbertCurvePartitioningCensusExample.ipynb).  I'm not sure where to put spatial-points.ipynb; it can't go in the Datashader user guide, which is meant only to rely on the taxi dataset and otherwise use only synthetic data to make the site easily buildable and domain-independent.  The default would be a topic on pyviz.org, since the datasets will already be there for building the census and osm-1billion notebooks, but it's definitely a notebook more about the operation of software than about the actual domain topic like the rest of the things that go there, so there may be some better place for it.

We should also update the census and osm datasets distributed with Datashader to be spatially ordered, and should update census.ipynb and osm-1billion.ipynb to make use of the spatial ordering (once we have HoloViews support for it).]

PR #703, opened 2019-02-08
Added missing entry to datasets.yml

jlstevens:
Added an entry to fetch `nyc_taxi_wide.parq` needed by the updated dashboard example. Longer term, it would be good to eliminate the redundancy of having this entry along with `nyc_taxi.csv`.

### DIFF
--- a/examples/datasets.yml
+++ b/examples/datasets.yml
@@ -12,6 +12,11 @@ data:
     files:
       - nyc_crime.csv
 
+  - url: https://s3.amazonaws.com/datashader-data/nyc_taxi_wide.parq
+    title: 'NYC Taxi Data Wide'
+    files:
+      - nyc_taxi_wide.parq
+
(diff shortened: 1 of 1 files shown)


jlstevens:
@jbednar I did a grep and I don't think `nyc_taxi.csv` is used anymore. Can you confirm? If so, I'll delete that entry.

jbednar:
This is fine for now, because the datasets.hml and dashboard.yml files should probably be merged to a single file using intake, and the topics need to move to pyviz.org topics, etc.  We also need to make a new Parquet file containing everything important from the .csv file; right now nyc_taxi_wide.parq doesn't include e.g. the `fare` column, and neither file includes enough area to cover JFK which has some interesting patterns, so we just need a whole new file.

jbednar:
nyc_taxi.csv is used in nyc_taxi-nongeo.ipynb, and can't easily be switched because that notebook needs `fare`.

PR #708, opened 2019-02-13
Updated docs and examples for SpatialPointsFrame

jbednar:
Changed census and osm-1billion examples to use SpatialPointsFrame.  Still need to update osm.ipynb and the user guide.


examples/topics/osm-1billion.ipynb
```diff
@@ -40,7 +40,7 @@
    "outputs": [],
    "source": [
     "%%time\n",
-    "df = dd.io.parquet.read_parquet('../data/osm-1billion.snappy.parq')\n",
+    "df = SpatialPointsFrame.from_parquet('../data/osm-1billion.parq').persist()\n",
     "df = df.persist()"
```
philippjfr: Persisting twice here.

jbednar: Oops; thanks!

jbednar: Fixed, and also hadn't uploaded the updated census.ipynb, also fixed.

jbednar:
@jonmmease, can we set it up to have a more similar interface to dask?  Dask supports `import dask.dataframe as dd; df = dd.read_parquet(....)`, which is what I'd usually suggest using:

```
def read_parquet(path, columns=None, filters=None, categories=None, index=None,
                 storage_options=None, engine='auto', infer_divisions=None):
```

Because a SpatialPointsFrame is a drop-in replacement for any other Dask DataFrame, it would be nice if people could use the same interface for loading it, so that it's simple to switch a notebook between either approach.  E.g. if we had a similar `read_parquet` in ds.spatial, we could do `import datashader.spatial as dss; df = dss.read_parquet(....)`, and tell people to use it just like Dask's function.

jbednar:
And maybe make the creation of a SpatialPointsFrame follow `dd.to_parquet`?

jonmmease:
Great idea @jbednar, unifying the API makes sense.  Now that we have the `.spatial` accessor, it will be pretty easy for people to check whether the spatial metadata was found so I'm thinking we should always return a `SpatialPointsFrame`, that may or may not contain the spatial metadata.

I can take a crack at this this afternoon.

jonmmease:
> And maybe make the creation of a SpatialPointsFrame follow `dd.to_parquet`?

Yeah, we can name it `to_parquet` and give the supported options the same names, but most of the `dd.to_parquet` options won't be supported.

For unsupported options, do you think it's worth the effort to accept them and then error if they have unsupported values?

jbednar:
Great.  Please make those changes in this PR, updating the notebooks involved so that we can see how the API is affected in practice.  (The PR can be renamed to match, of course!) 

For `to_parquet`, I think it's fine to accept only the options that make sense for a SpatialPointsFrame.

jbednar:
I've also now updated osm.ipynb to use a SpatialPointsFrame, which gives massive speedups for small areas (44 seconds for the full dataset, 15 seconds for France and most of Germany, and a half second for Paris):

image
image
image

jbednar:
Interactive performance for the 3-billion point version is still pretty bad, because at least when starting with the whole dataset, it's not practical to call .persist(), making it all be disk-based rather than memory based.  But one could use the SpatialPointsFrame to quickly select a subset of the data explicitly before launching an interactive exploration, e.g. to pull out only France and launch that as an interactive plot of a .persisted subset of the data.  It would be nice to show an example of doing so.

jonmmease:
Ok, I updated the API to match dask more closely (as discussed above) and I updated the API usage in the notebooks. I haven't reviewed the discussion in the notebooks yet, just the code.  I'll work more on the prose this evening.

examples/topics/census.ipynb
```diff
@@ -615,23 +611,35 @@
...
+    "The dashboard.ipynb example shows this same census data in the context of an interactive dashboard, including color keys and hover information that help reveal the magnitudes at every location even while the plot faithfully reveals the structure."
    ]
   }
  ],
  "metadata": {
+  "kernelspec": {
+   "display_name": "Python 3",
+   "language": "python",
+   "name": "python3"
+  },
```
jbednar: Would be good to clear this metadata before merging this.  I use strip_kernel from Jean-Luc:

```
#!/bin/bash
cp "$1" /tmp/
/usr/local/bin/jq --indent 1 \
    '
    (.cells[] | select(has("outputs")) | .outputs) = []
    | (.cells[] | select(has("execution_count")) | .execution_count) = null
    | .metadata = {"language_info": {"name":"python", "pygments_lexer": "ipython3"}}
    | .cells[].metadata = {}
    ' "$1" > /tmp/stripped.ipynb
cat /tmp/stripped.ipynb > "$1"
```

examples/topics/osm-1billion.ipynb
```diff
@@ -157,7 +157,7 @@
    "source": [
     "### Interactive plotting\n",
     "\n",
-    "As long as the data fits into RAM on your machine, performance should be high enough to explore the data interactively, with only a brief pause after zooming or panning to a new region of the data space:"
+    "As long as the data fits into RAM on your machine, performance should be high enough to explore the data interactively, with only a brief pause after zooming or panning to a new region of the data space.  Performance should be much higher when you are zoomed in, thanks to the spatially sorted data structure used here; if you use a regular Pandas or Dask dataframe if will need to iterate through the entire dataset on each zoom rather than only considering a subset as used here."
```
jbednar:
```suggestion
    "As long as the data fits into RAM on your machine, performance should be high enough to explore the data interactively, with only a brief pause after zooming or panning to a new region of the data space.  Performance should be much higher when you are zoomed in, thanks to the spatially sorted data structure used here; if you use a regular Pandas or Dask dataframe it will need to iterate through the entire dataset on each zoom rather than only considering a subset as used here."
```

examples/topics/census.ipynb
jonmmease: Thanks for the tip, will do.

jonmmease:
> Hopefully you'll be able to add some docs, particularly for how to create the files, how to set the parameters on creation, etc.

Yup, that's what I'm working on.  I just finished the background/research notebook that will live on anaconda cloud (https://anaconda.org/jonmmease/spatial-partitioning-for-datashader-points-rendering/notebook).  Next I'm going to look at creating the "Points" notebook section with some information on the `SpatialPointsFrame` API, and a reference to this notebook as background material.

> Also please double-check the text in the example notebooks that's changed to make sure it matches the final API that we settle on.

Will do, I'll also add references to the new Points section for people to learn more.

> I guess you moved the points support from `spatial/__init__.py` to `spatial/points.py` in anticipation of adding additional spatial datatypes later?

Yes, apologies for not calling that out.  I expect that the required metadata for a future `SpatialTrimeshFrame` will be substantively different from what we have now for `SpatialPointsFrame`.  But that honestly not very well thought out at this point, so it's hard to settle on an API in advance.

For `read_parquet`, I think it definitely makes sense to import it as `spatial.read_parquet` because down the road this function could be generic and return whichever spatial data structure is encoded in the parquet file.  I'm a bit reluctant about importing `to_parquet` as `spatial.to_parquet`, because I expect the preprocessing logic, and API, for a spatially partitioned trimesh to be somewhat different than what we have now for points.

So I'd lean towards supporting and encouraging `datashader.spatial.read_parquet`, but keeping `datashader.spatial.points.to_parquet` as-is.  How does that sound?

jbednar:
Sounds perfect, thanks!

jonmmease:
Updated documentation is ready for review!

In addition, I added a new `nrows` metadata property to the `SpatialPointsFrame` parquet metadata.  This is now created automatically by `spatial.points.to_parquet`, but here is a code snippet to add `nrows` to a  `SpatialPointsFrame` parquet file that was generated before this change

```python
# Path to SpatialPointsFrame file that doesn't contain nrows metadata
path = '/path/to/census2010.parq'

# Imports
import fastparquet as fp
import json
import dask.dataframe as dd
import copy
import os

# Read as dask dataframe
ddf = dd.read_parquet(path)

# Open parquet file
pf = fp.ParquetFile(path)

# Load spatial points props
props = json.loads(pf.key_value_metadata['SpatialPointsFrame'])

# Compute nrows
nrows = len(ddf)

# Add nrows to properties
props['nrows'] = nrows

# Update file metadata
fmd = copy.copy(pf.fmd)
spatial_key_val = [kv for kv in fmd.key_value_metadata if kv.key=='SpatialPointsFrame'][0]
spatial_key_val.value = json.dumps(props)

# Overwrite file metadata
fn = os.path.join(path, '_metadata')
fp.writer.write_common_metadata(fn, fmd, no_row_groups=False)

fn = os.path.join(path, '_common_metadata')
fp.writer.write_common_metadata(fn, fmd)

# Make sure load works
from datashader import spatial
sframe = spatial.read_parquet(path)
print(sframe.spatial.nrows)
```

examples/topics/census.ipynb
jonmmease: Done in 2f6b7ff

jbednar:
Thanks!  I've used the script to update census2010.parq and osm-1billion.parq to include nrows, and uploaded zips of those files to S3.  I also streamlined the 2_Points.ipynb file a bit, but the story should still be the same.  It should be ready to merge when the tests pass.

PR #717, opened 2019-02-23
Fix memory explosion when auto-calculating canvas range extents with dask

### LINKED ISSUE #668: Memory errors on distributed dask cluster
## Description

I have a persisted dask dataframe which is larger than the amount of memory on my notebook server and any of the individual workers. The structure is `x, y, z` lidar data.

When trying to plot with datashader it seems to attempt to transfer the whole dataframe to the notebook when aggregating before plotting.

```
ddf = client.persist(dd.read_parquet('Some 20GB dataset'))
cvs = ds.Canvas(900, 525)
agg = cvs.points(ddf, 'x', 'y', agg=ds.mean('z'))
```

This results in 20GB of data being transferred to my notebook (and it gets killed by the OOM killer as I only have 16GB of RAM).

## Your environment

Datashader version: 0.6.8
Dask version: 0.20.0
Distributed version: 1.24.0

jonmmease:
This PR improves the Canvas autorange logic to avoid bringing each full x/y array into memory in order to compute the min/max values.

This fixes  #668 for me.  Here is the test case I I've been using to diagnose and test this PR.  I started a Dask distributed instance on a large workstation and persisted the ~3-billion point OSM dataset into memory. After this persist, the Dask dashboard reports ~28GB memory used.  Then I set up a VM with 8GB of RAM and connected it as a client of the distributed scheduler. I then performed a `cvs.points` aggregation without specifying `x_range`/`y_range`, causing the autorange logic to be invoked. 

```
import dask.dataframe as dd
import datashader as ds
osm = dd.read_parquet('/path/to/osm-3billion.parq/').persist()
cvs = ds.Canvas()
agg = cvs.points(osm, x='x', y='y')
```

Before these changes, the memory usage of the client would climb steadily until the kernel died.  With these changes, the aggregation completes successfully with no noticeable increase in memory usage on the client.

cc @jacobtomlinson

### DIFF
--- a/datashader/dask.py
+++ b/datashader/dask.py
@@ -31,8 +31,13 @@ def dask_pipeline(df, schema, canvas, glyph, summary):
 
 
 def shape_bounds_st_and_axis(df, canvas, glyph):
-    x_range = canvas.x_range or glyph.compute_x_bounds_dask(df)
-    y_range = canvas.y_range or glyph.compute_y_bounds_dask(df)
+    if not canvas.x_range or not canvas.y_range:
+        x_extents, y_extents = glyph.compute_bounds_dask(df)
+    else:
+        x_extents, y_extents = None, None
+
+    x_range = canvas.x_range or x_extents
+    y_range = canvas.y_range or y_extents
     x_min, x_max, y_min, y_max = bounds = compute(*(x_range + y_range))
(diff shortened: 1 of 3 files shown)


jbednar:
Looks great, thanks!

PR #718, opened 2019-02-28
Fixed topics index

jbednar:
The topics index was not listing Strange Attractors.  Also cleaned up the formatting, and clarified wording in one notebook.


jbednar:
Fixes applied by @jsignell separately.

PR #723, opened 2019-03-09
Updated dashboard for newer param syntax

jbednar:
Following https://github.com/pyviz/datashader/issues/698 , updates dashboard.ipynb to track new features of the Param API that allow simpler parameter specification and usage.  Probably requires branch bokeh_layout_refactor of HoloViews and latest dev release of Bokeh, plus latest master of panel and param.

Mostly works, but is not updating on zooms.  Also, I don't see how to use the improvements in https://github.com/pyviz/holoviews/pull/3474 in this case; can they e.g. be used for applying opacity?


jbednar:
Should work for zooming now, but I had hoped the explicit RangeXY wasn't going to be needed anymore...

examples/dashboard.ipynb
```diff
@@ -117,16 +112,13 @@
...
-    "        c_stream   = hv.streams.Params(self, ['cmap', 'normalization'])\n",
-    "        s_stream   = hv.streams.Params(self, ['spreading'], rename={'spreading': 'px'})\n",
-    "        shaded     = spread(shade(rasterized, streams=[c_stream]), streams=[s_stream], how=\"add\")\n",
-    "        shaded     = hv.util.Dynamic(shaded, operation=self.apply_opacity)\n",
-    "        \n",
-    "        return hv.DynamicMap(self.tiles) * shaded * hv.DynamicMap(self.labels)\n",
+    "        rasterized = rasterize(hv.DynamicMap(self.elem), aggregator=self.aggregator, width=800, height=400)\n",
+    "        shaded     = shade(rasterized, cmap=self.param.cmap)\n",
+    "        spreaded   = spread(shaded, px=self.param.spreading, how=\"add\")\n",
+    "        dataplot   = spreaded.map(self.apply_opacity)\n",
```
philippjfr:
```suggestion
    "        dataplot   = spreaded.apply(self.apply_opacity)\n",
```

philippjfr:
I'm not following your last comment, it looks like you got rid of RangeXY...

examples/dashboard.ipynb
jbednar: Odd; I had already done that but it seems to have been overwritten somehow.

jbednar:
> I'm not following your last comment, it looks like you got rid of RangeXY...

I'm not sure what was going on there; somewhere between me, git, and jupyter things weren't updating properly, losing the change from map to apply and not including RangeXY.  It looks like the PR is now up to date properly, and you should be able to see that RangeXY is still needed if we want zooming to work.  I would have also hoped to avoid some of the extra methods that we have now; can any of them be replaced with simply providing an instance parameter?

philippjfr:
Definitely shouldn't be needed, so if this is real it's likely a bug or maybe I didn't rebase my hv PR a second time.

jbednar:
I'll pull again and test.

philippjfr:
You could also try rebasing the PR yourself.

philippjfr:
Yeah definitely needs rebasing, I'll do it in a bit if you haven't already.

jbednar:
I pulled current master on panel and param and then pulled into holoviews' branch bokeh_layout_refactor, but they were already up to date, and I verified that it only works for zooming if I include the RangeXY stream explicitly.  Rebasing again might help; I can try later, but will be on a road trip until this evening now.

jbednar:
Ok, I rebased the hv bokeh_layout_refactor branch locally, which did remove the need for RangeXY.  Thanks for whatever you merged to hv to support that! I won't push the branch to github, though, as there was some error indicating that the PR was set up in an unusual way, so it's better to let you rebase the hv PR.

I also noticed that the normalization parameter hadn't been wired up to have any effect, which I fixed. 

The last thing I can see to address here is how to make it easier to set options based on widget values.  Can `.ops()` be extended so that we don't need `apply_opacity` and its dependency declaration, and instead simply do `dataplot   = spreaded.opts(alpha=self.param.data_opacity, show_legend=False)` in the `viewable()` method?  To do this .opts would need to understand Parameter objects the way operations do, and watch their dependencies appropriately.

Alternatively, seems like we could provide a generic and reusable operation that applies options, but it would be more obvious and more convenient if `.opts()` can do it. @jlstevens and @philippjfr, can that be done?

jbednar:
Ok, I guess there's one more thing we could consider.  Should we support arithmetic operations on Parameter instance objects?  E.g. in this particular case, once .opts() allows parameter instances, we could eliminate the `labels` method as well, by doing something like `gts.StamenLabels.options(level='annotation', alpha=int(self.param.show_labels)-1.0)`, which isn't a great example but hopefully conveys what I mean (that we could support expressions that transform a parameter value via some expression (as for dim()) before it gets used).

philippjfr:
I've considered that in the past but I know Jean-Luc has reservations and I do too. Personally I think doing it for just ``.opts`` would be inconsistent and we should find a way of supporting it everywhere, at the same time I think doing that would cause horrible complexity all over the place. Therefore I'd suggest something else, namely that the ``apply`` method accepts the names of methods to apply dynamically, e.g. to dynamically apply some option you would do ``.apply('opts', color=c.param.value)``. This takes about three extra lines of code in ``.apply``:

color_picker

jlstevens:
I think a string that points to a method for `apply` seems like a reasonable extension and more general than hacking about in `.opts`.

jbednar:
If it ever comes down to it, .opts (or applying options in general) is an extremely important specific case when making dashboards, so it's worth it working well.  That said, the proposed extension to `apply` sounds very useful and essentially subsumes such specific support.

jbednar:
Just to double check, with the proposed syntax we'd be able to do `dataplot   = spreaded.apply('opts', alpha=self.param.data_opacity, show_legend=False)`, and delete `apply_opacity`?

philippjfr:
Yes, that's exactly right.

jbednar:
I thought of one strong reason we might still want to consider special-purpose support for parameter instances in .opts: if we had that, wouldn't we be able to support having live widgets even on static export?  I don't see how that would be possible from `.apply('opts')`, but it seems straightforward to support static export if .opts accepts parameter objects and that mapping can then be set up statically.

philippjfr:
>I thought of one strong reason we might still want to consider special-purpose support for parameter instances in .opts: if we had that, wouldn't we be able to support having live widgets even on static export?

Don't see how that makes any real difference, currently holoviews knows nothing about static links and if we ever add that it makes no difference whether that lives in ``apply`` or ``opts``.

jbednar:
But `apply` just calls any Python method, which seems like it has no hope of ever returning a static link, while .opts has a very clear semantics that doesn't depend on calling any Python code; it simply sets the value of each option.  So I can see hope for `.opts`, but not for `apply`.

philippjfr:
IMO both are equally problematic, and at best they can leave hints for panel to replace the dynamic behavior with a static link at rendering time.

philippjfr:
To expand on this, neither ``.apply`` or ``.opts`` will know that a static link is even possible if all it's given is a param instance. It would also be weird for it to change the return type depending on whether it thinks a static link is possible. This means that it will always return a DynamicMap, it's only during render time that panel can inspect both the widget and the DynamicMap and replace dynamic behavior with a static link. In either case ``.apply`` or ``.opts`` will therefore likely have to rely on a special ``Callable`` subclass which provides a hint to panel that the DynamicMap callback can be substituted for a static link if rendered alongside the widget or parameterized that controls the value.

jbednar:
Still seems simpler to do that for .opts than for .apply, but as you'd be the one doing it, that's up to you! :-)

I've extensively rewritten the "how it works" section to reflect the newly supported syntax, which has gotten much simpler and easier to explain.  It assumes we'll implement the `apply('opts'` approach, but if we don't, we can revert my last commit; the previous commit doesn't make that assumption.

My explanation is still quite wordy, and probably difficult for people to follow, but at least now I'm happy with the concepts involved, and I don't see any major simplifications that could be made to the approach other than eliminating some methods by supporting dynamic arithmetic and logical operators on Parameter instances.  And the resulting dashboard is now only 50 lines, with everything working once apply is updated!

jlstevens:
I'm in favor of the `apply` approach - it isn't too clunky but a lot more explicit. When using `apply`, you know you are working with dynamic content (i.e `DynamicMaps`) due to the application of the function/method. This is unlike `.opts` which applies to static elements (that can't update on parameter changes unless it also promoted to dynamic, which it definitely shouldn't).

This is on top of the fact that I don't ever want to touch `.opts` again :-)

philippjfr:
I am highly perplexed by this:

```
source.to_dask().persist();
```

I understand that sources now implement a persist method, but that's no using that, here you are converting to a dask dataframe and then persisting but not actually assigning the persisted dataframe to anything. So this loads the data into memory and then immediately discards it

jbednar:
That looks like a copy-paste error from a while back; can't check right now...

jbednar:
I benchmarked it, and that line does speed up the generation of a plot of the data by about 1 second for nyc_taxi, but as you would expect it doesn't speed up later plots, so it makes for a net slowdown. For now I've removed it, and will merge, but I can't see any way that actually does control how an Intake source is persisted internally, which I'll raise at the Intake site.  (Intake's `persist` is about saving to local disk, not persisting in memory, so it's not related.)  Thanks for finding this!

jbednar:
BTW, the net result of all the changes in this PR is that the entire dashboard now fits into 48 lines of code, many of which are simpler than before this PR (both in character count and in the concepts involved).  The Bokeh original was over 500 lines, with almost identical functionality, and the version before this PR was 59 lines.  

Further reductions in length or complexity are always possible, but they aren't likely to be dramatic unless functionality is removed.  Of the remaining lines, most are to do with directly declaring things for the user to manipulate:

```
- 16 Declaring parameters and their ranges
-  7 Plotting and declaring plot options
-  6 Python class boilerplate (`class`, `def`)
-  5 Applying user-selected parameters
-  4 Python imports
-  4 Expressing Parameter dependencies
-  3 Opening Intake catalog and dataset
-  3 Building the dashboard with Panel
  == 
  48 lines total
```

PR #730, opened 2019-04-02
Adding persist to make it use dask

jsignell:


examples/dashboard.yml
```diff
@@ -36,11 +37,11 @@ sources:
           label: Drop-off Hour
         pickup_hour:
           label: Pick-up Hour
-      plot:
```
philippjfr: I'm confused why was this removed?

jsignell: I was using the github UI. Just a mistake

jsignell:
I'm going to merge this since it's passing. Sound good @jbednar?

PR #733, opened 2019-04-05
change tiles.py to avoid rendering empty tiles.

abayuen:
All tiles within the full_data_extent were rendered. However, not all tiles contained data and many of them can be empty.
In this PR, we added check to avoid rendering empty tiles.

### DIFF
--- a/datashader/tiles.py
+++ b/datashader/tiles.py
@@ -4,6 +4,8 @@ from io import BytesIO
 import math
 import os
 
+import numpy as np
+
 import dask
 import dask.bag as db
 
--- a/datashader/tiles.py
+++ b/datashader/tiles.py
@@ -92,7 +92,11 @@ def render_super_tile(tile_info, output_path, load_data_func,
     agg = rasterize_func(df, x_range=tile_info['x_range'],
                          y_range=tile_info['y_range'], height=tile_size,
                          width=tile_size)
-    ds_img = shader_func(agg, span=tile_info['span'])
+    # if the rasterized data is empty, do not create subtiles.
+    if np.sum(agg) == 0:
+        pass
+    else:
+        ds_img = shader_func(agg, span=tile_info['span'])
     return create_sub_tiles(ds_img, level, tile_info, output_path, post_render_func)
 
 def create_sub_tiles(data_array, level, tile_info, output_path, post_render_func=None):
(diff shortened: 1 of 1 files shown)


jbednar:
Do all tile-server clients reliably handle missing tiles, e.g. by scaling the tile from the previous zoom level?  If not it doesn't seem safe to skip the empty tiles in general, though it would make sense to add an option for it.

abayuen:
Hello @jbednar,
Thanks for your contribution to datashader!

All time-server client should reliable handle 'missing' tiles. Datashader is already introducing 'missing tiles' by passing full_extent as argument in the render_tile function, in a sense that the tiles outside of full_extend are not generated.

abayuen:
I am not sure what the rules are for merging to master.. so I will wait for your reply

jbednar:
> Datashader is already introducing 'missing tiles' by passing full_extent as argument in the render_tile function, in a sense that the tiles outside of full_extend are not generated.

I think that's a different issue.  Sure, all sets of tiles are finite, and so unless they cover the world, it's possible to come to the end of them.  But that's different than saying that at some zoom level, a particular tile in the middle that has no data should not be rendered at all; in some cases it should, and others not.  So I think there are valid reasons for wanting a tile to be generated even when there is no data, so I'm happy to merge this once there is an argument to turn this behavior on and off.

brendancol:
@abayuen I'm not sure if the `sum == 0` is a good-enough heuristic for empty tiles...but maybe it would...I'm looking at this soon and will ping back

jbednar:
@brendancol, did you ever have a check about whether these changes are ok?  They seem fine to me if they are controlled by an option, but I'm less sure they should be the default.

datashader/tiles.py
```diff
@@ -92,7 +94,11 @@ def render_super_tile(tile_info, output_path, load_data_func,
     agg = rasterize_func(df, x_range=tile_info['x_range'],
                          y_range=tile_info['y_range'], height=tile_size,
                          width=tile_size)
-    ds_img = shader_func(agg, span=tile_info['span'])
+    # if the rasterized data is empty, do not create subtiles.
+    if np.sum(agg) == 0:
+        pass
```
philippjfr: Won't this just error because `ds_img` won't be defined below?

jbednar:
```suggestion
        return
```
?

brendancol: @jbednar I would vote for closing this PR and not using the `np.sum(agg) == 0` heuristic because it feels brittle to me.

jbednar: Ok, will do.  Anyone who feels otherwise can reopen while addressing the above issue(s).

Issue #749, opened 2019-05-21
Import datashader in Jupter notebook with Python 3.5 kernel hangs cell.


akum036:
## Description

If install a fresh Python 3.5.2 environment with just Jupyter and Datashader (with dependencies), then try and import Datashader in the notebook, the notebook hangs in any cells run after the datashader import.

This does not happen when running from an ipython terminal or if using Python 3.7

## How to reproduce

From the CLI:
`python 3.5 -m venv venv`
`. venv/bin/activate`
`pip install jupyter datashader`
`jupyter notebook`

In the notebook:
`import datashader` - Cell 1

`print('hello')` - Cell 2

The `print('hello')` command will not complete after Cell 1 has finished running. No warnings are given.
## Errors and warnings

Looking at the debug output of the terminal when running the notebook, the pattern of status, execute_input and stream is broken, the execute_inpu part doesn't run again.

## Your environment
Ubuntu 16.04
Python 3.5.2
Firefox 66.0.5

attrs==19.1.0
backcall==0.1.0
bleach==3.1.0
bokeh==1.1.0
certifi==2019.3.9
chardet==3.0.4
Click==7.0
cloudpickle==1.1.1
colorcet==2.0.1
cycler==0.10.0
dask==1.2.2
datashader==0.7.0
datashape==0.5.2
decorator==4.4.0
defusedxml==0.6.0
distributed==1.28.1
entrypoints==0.3
HeapDict==1.0.0
idna==2.8
imageio==2.5.0
ipykernel==5.1.1
ipython==7.5.0
ipython-genutils==0.2.0
ipywidgets==7.4.2
jedi==0.13.3
Jinja2==2.10.1
jsonschema==3.0.1
jupyter==1.0.0
jupyter-client==5.2.4
jupyter-console==6.0.0
jupyter-core==4.4.0
kiwisolver==1.1.0
llvmlite==0.28.0
locket==0.2.0
MarkupSafe==1.1.1
matplotlib==3.0.3
mistune==0.8.4
msgpack==0.6.1
multipledispatch==0.6.0
nbconvert==5.5.0
nbformat==4.4.0
networkx==2.3
notebook==5.7.8
numba==0.43.1
numpy==1.16.3
packaging==19.0
pandas==0.24.2
pandocfilters==1.4.2
param==1.9.0
parso==0.4.0
partd==0.3.10
pexpect==4.7.0
pickleshare==0.7.5
Pillow==6.0.0
pkg-resources==0.0.0
[...]

PR #750, opened 2019-05-21
Updating travis to test with python 3.7 - removing 3.5

jsignell:


jsignell:
Removing references to supporting 3.5 and adding tests for 3.7 in response to #749

PR #751, opened 2019-05-22
Removal of InteractiveImage

jlstevens:
This PR aims to 1) remove `InteractiveImage` from all the docs 2) remove the `InteractiveImage` code 3) remove direct use of Bokeh in the remaining spots (HoloViews can be used instead).

Currently all mentions of `InteractiveImage` have been removed from `datashader/examples` which means that the corresponding code has been replaced with a HoloViews equivalent. 

I've made minimal changes to the text around the code (except for the Interactivity guide that needed to be completely reworked). In short, more edits will probably be required and @jbednar may want to have a pass over the textual content.

@jsignell Pinging you as the changes in the topic notebooks will need to propagate to `pyviz-topics/examples`.


jbednar:
Merging as the PR build is passing; the broken push build shouldn't matter as master has been updated. The wording in the notebooks will need some additional updating on master, once pyviz-topics/examples are ready.

PR #752, opened 2019-05-22
Removing topics

jsignell:
Topics are moving to pyviz-topics/examples so they no longer need to be here.


jbednar:
Also need to move nyc_taxi-nongeo.ipynb (never categorized when topics/ were created, but can now be part of the nyc_taxi project).  And need to update docs to point to pyviz_topics/examples/ instead, but as that's still in flux, presumably can't complete that?

jsignell:
Right this is not quite ready yet, need to build the static website version of pyviz-topics/examples

jbednar:
It's ok to merge this to unblock other PRs, as long as we'll deal with that before the next release.

jsignell:
Ok then I'll merge.

PR #754, opened 2019-06-04
Implementation of compare aggregator

jonmmease:
As a follow-on to the filled area support that was added in https://github.com/pyviz/datashader/pull/734, this PR adds a `compare` aggregator as discussed in https://github.com/pyviz/datashader/pull/734#issuecomment-485019246.

Here is the docstring
```
    Compare the value of two columns per bin.

    Returns a value of 1 if column1 >= column2 in a particular bin more often
    column1 < column2.  Returns -1 otherwise.

    Parameters
    ----------
    column1, column2: str
        The names of the two columns to compare
```

Example with line glyph:
```python
import pandas as pd
import numpy as np
import datashader as ds
from datashader import Canvas
cvs = Canvas()
xs = np.array([1, 2, 3, 4, 5, 6])
ys1 = np.array([3, 2, 3.1, 3, 3, 4])
ys2 = np.array([0, 0.5, 2, 4, 4, 5])
df = pd.DataFrame({'xs': xs, 'ys1': ys1, 'ys2': ys2})
agg = cvs.line(df, x='xs', y=['ys1', 'ys2'],  agg=ds.compare('ys1', 'ys2'))
tf.spread(tf.shade(agg))
```
lines_compare

Example with area glyph:
```python
import pandas as pd
import numpy as np
import datashader as ds
from datashader import Canvas
cvs = Canvas()
xs = np.array([1, 2, 3, 4, 5, 6])
ys1 = np.array([3, 2, 3.1, 3, 3, 4])
ys2 = np.array([0, 0.5, 2, 4, 4, 5])
df = pd.DataFrame({'xs': xs, 'ys1': ys1, 'ys2': ys2})
agg = cvs.area(df, x='xs', y='ys1', y_stack='ys2',  agg=ds.compare('ys1', 'ys2'))
tf.spread(tf.shade(agg))
```
area_compare

Note: For the line/area case, the behavior of the aggregation at the crossing is a little unintuitive because the single aggregation result is used for the entire line/area segment. I don't think there's really a way around this within the current aggregation architecture, but let me know what you think.

### DIFF
--- a/datashader/core.py
+++ b/datashader/core.py
@@ -642,7 +642,7 @@ The axis argument to Canvas.line must be 0 or 1
 
         if agg is None:
             agg = mean_rdn(weight_col)
-        elif agg.column is None:
+        elif hasattr(agg, 'column') and agg.column is None:
             agg.column = weight_col
 
         cols = source.columns
@@ -717,7 +717,7 @@ The axis argument to Canvas.line must be 0 or 1
                              % type(source).__name__)
 
         column = None
-        if isinstance(agg, rd.Reduction):
+        if isinstance(agg, rd.ReductionSingleColumn):
             agg, column = type(agg), agg.column
             if (isinstance(source, DataArray) and column is not None
                 and source.name != column):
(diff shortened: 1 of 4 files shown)


jbednar:
Jon, are you planning to support a scalar baseline?  If so I'll wait on that; if not I'm tempted to close this PR as generating results that are too confusing to interpret, and too difficult to implement properly (by finding the intersection when it's located in between data points).

jonmmease:
I think we should close for now. The baseline idea doesn't really fit the aggregator paradigm as it's currently implemented, so that would take some more thought.

PR #761, opened 2019-07-02
added tile_previewer utility function

brendancol:
The `tile_previewer` utility is a helper for creating a Bokeh figure with a tile layer. It works in notebooks, but can also save the figure out to be stored with a tileset.  A future PR should include generating an `index.html` file in the root of the tileset directory so tiles can be easily looked at.

Example Usage
---------------

```python
from datashader.tiles import tile_previewer
tileset_url = 'https://my-s3-bucket.s3.amazonaws.com/9615-forced-test/{Z}/{X}/{Y}.png'
fig = tile_previewer(full_extent=full_extent_of_data, height=800, width=800, tileset_url=tileset_url)
```

### DIFF
--- a/datashader/tiles.py
+++ b/datashader/tiles.py
@@ -316,6 +316,62 @@ class TileRenderer(object):
             yield (img, x, y, z)
 
 
+def tile_previewer(full_extent, tileset_url,
+                   output_dir=None,
+                   filename='index.html',
+                   title='Datashader Tileset',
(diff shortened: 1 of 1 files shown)


jbednar:
Looks good. Please document that wherever the tile generation is documented. Actually, I can't see any docs for that, so maybe tile generation needs documenting altogether?

brendancol:
@jbednar I'm promising to provide documentation for tiles on a follow up PR.

PR #762, opened 2019-07-05
Implemented distributed regridding

philippjfr:
Implements regridding of ``dask.array.Array`` backed ``xarray.DataArray`` objects without loading the entire array into memory.

The basic idea here is that the operation will no longer load dask arrays into memory, instead it will resample the original data in chunks and then stack the resampled chunks along the two axes. By default it will try to keep the chunksize consistent, however it now also exposes a memory constraint which can be used to limit the amount of data in memory at any one time. 

If enabled it is possible to end up with situations where resampling cannot be performed given the memory constraints, e.g. let us say you have a 1 GB array which is roughly 12k x 12k in shape, and we want to resample it into a 10x10 grid. If we we give it a memory limit of 100 KB, even if each output chunk is only 1x1 it still has to load a 1200x1200 grid, i.e. about 11.5 MB, into memory . This could be solved by splitting the task and creating temporary chunks which are then aggregated but I considered that out of scope. Instead you will get an error that you have to relax the memory constraint.

Here is the profiling output when regridding a 10k x 10k grid into a 100x100 grid:

bokeh_plot - 2019-07-15T131812 114

Here is the same, but this time the array isn't already in memory but loaded from disk with zarr:

bokeh_plot - 2019-07-15T140555 076

Here we are zooming on a 50k x 50k (20 GB) array loaded with zarr (note how it slows down when we zoom out), while trying to minimize memory load:

dask_zarr

### Known issues

- Upsampling with 'linear' interpolation does not yet use correct padding so boundaries are wrong

- [x] Add tests

### DIFF
--- a/datashader/core.py
+++ b/datashader/core.py
@@ -675,7 +675,7 @@ The axis argument to Canvas.line must be 0 or 1
         Also supports resampling out-of-core DataArrays backed by dask
         Arrays. By default it will try to maintain the same chunksize
         in the output array but a custom chunksize may be provided.
-        If there are memory constrains they may be defined using the
+        If there are memory constraints they may be defined using the
         max_mem parameter, which determines how large the chunks in
         memory may be.
 
@@ -696,7 +696,7 @@ The axis argument to Canvas.line must be 0 or 1
         agg : Reduction, optional default=mean()
             Resampling mode when downsampling raster.
             options include: first, last, mean, mode, var, std, min, max
-            Also accepts string names, for backwards compatibility.
+            Accepts an executable function, function object, or string name.
         interpolate : str, optional  default=linear
             Resampling mode when upsampling raster.
             options include: nearest, linear.
--- a/datashader/core.py
+++ b/datashader/core.py
@@ -5,7 +5,7 @@ from numbers import Number
 import numpy as np
 import pandas as pd
 import dask.dataframe as dd
-from dask.array import Array
+import dask.array as da
 from six import string_types
 from xarray import DataArray, Dataset
 from collections import OrderedDict
@@ -14,7 +14,7 @@ from datashader.spatial.points import SpatialPointsFrame
 from .utils import Dispatcher, ngjit, calc_res, calc_bbox, orient_array, compute_coords
 from .utils import get_indices, dshape_from_pandas, dshape_from_dask
 from .utils import Expr # noqa (API import)
-from .resampling import resample_2d
+from .resampling import resample_2d, resample_2d_distributed
 from . import reductions as rd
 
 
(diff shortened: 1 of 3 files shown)


datashader/core.py
```diff
@@ -670,28 +672,44 @@ def raster(self,
         "nodata" attribute of the raster) are replaced with `NaN` if
         floats, and 0 if int.
 
+        Also supports resampling out-of-core DataArrays backed by dask
+        Arrays. By default it will try to maintain the same chunksize
+        in the output array but a custom chunksize may be provided.
+        If there are memory constrains they may be defined using the
```
jbednar:
```suggestion
        If there are memory constraints they may be defined using the
```

datashader/core.py
```diff
@@ -670,28 +672,44 @@ def raster(self,
...
+            Grid cell aggregation method for a possible downsampling.
+        us_method : str (optional)
+            Grid cell interpolation method for a possible upsampling.
+        nan_value : int or float, optional
+            Optional nan_value which will be masked out when applying
+            the resampling.
         agg : Reduction, optional default=mean()
             Resampling mode when downsampling raster.
             options include: first, last, mean, mode, var, std, min, max
             Also accepts string names, for backwards compatibility.
```
jbednar: This is confusing; isn't "first" a string?  Maybe should be clarified:
```suggestion
            Accepts an executable function, function object, or string name.
```

datashader/resampling.py
```diff
@@ -79,30 +84,241 @@
...
+    Parameters
+    ----------
+    src : dask.array.Array
+        The source array to resample
+    w : int
+        New grid width
+    h : int
+        New grid height
+    chunksize : tuple(int, int) (optional)
+        Size of the output chunks. By default this the chunk size is
```
jbednar:
```suggestion
        Size of the output chunks. By default the chunk size is
```

datashader/resampling.py
```diff
@@ -79,30 +84,241 @@
...
+            ch -= 1
+        dim = not dim
+    if ch == 0 or cw == 0:
+        min_mem = height_fraction * width_fraction * nbytes
+        raise ValueError(
+            "Given the memory constraints the resampling operation "
+            "could not find a chunksize that avoids loading too much "
+            "data into memory. Either relax the memory constraint to "
+            "a minimum of %d bytes or resample to a larger grid size. "
+            "Note: A future implementation may handle this condition "
```
jbednar:
```suggestion
            "Note: A future implementation could handle this condition "
```

(Changed to a message that will prompt someone to write a PR if they need this, rather than promising that we might do this ourselves. :-)

datashader/resampling.py
```diff
@@ -79,30 +84,241 @@
...
+        Grid cell aggregation method for a possible downsampling
+        (one of the *DS_* constants).
+    us_method : str (optional)
+        Grid cell interpolation method for a possible upsampling
+        (one of the *US_* constants, optional).
+    fill_value : scalar (optional)
+        If None, numpy's default value is used.
+    mode_rank : scalar (optional)
+        The rank of the frequency determined by the *ds_method*
+        ``DS_MODE``. One (the default) means most frequent value, zwo
```
jbednar:
```suggestion
        ``DS_MODE``. One (the default) means most frequent value, two
```

datashader/resampling.py
```diff
@@ -147,31 +371,43 @@ def upsample_2d(src, w, h, method=US_LINEAR, fill_value=None, out=None):
...
-    :param mode_rank: *scalar*, optional
-        The rank of the frequency determined by the *method* ``DS_MODE``. One (the default) means
-        most frequent value, zwo means second most frequent value, and so forth.
-    :param out: 2-D *ndarray*, optional
-        Alternate output array in which to place the result. The default is *None*; if provided, it must have the same
-        shape as the expected output.
-    :return: A downsampled version of the *src* array.
+    mode_rank : scalar (optional)
+        The rank of the frequency determined by the *ds_method*
+        ``DS_MODE``. One (the default) means most frequent value, zwo
```
jbednar:
```suggestion
        ``DS_MODE``. One (the default) means most frequent value, two
```

PR #763, opened 2019-07-09
Fix warnings with latest Numba

philippjfr:
Various fixes to datashader numba code to avoid the following warnings:

* Use correct JIT decorator for function which has no parallel target, avoiding a warning with latest numba versions.

* Avoid reflected list deprecation warnings, mostly by passing in tuples instead

* Declare forceobj mode for functions which don't support nopython and looplifting.

### DIFF
--- a/datashader/bundling.py
+++ b/datashader/bundling.py
@@ -33,9 +33,9 @@ def distance_between(a, b):
     return (((a[0] - b[0]) ** 2) + ((a[1] - b[1]) ** 2))**(0.5)
 
 
-@nb.jit
+@nb.jit(forceobj=True)
 def resample_segment(segments, new_segments, min_segment_length, max_segment_length, ndims):
-    next_point = np.array([0.0] * ndims)
+    next_point = np.zeros(ndims)
     current_point = segments[0]
     pos = 0
     index = 1
@@ -183,9 +183,8 @@ def get_gradients(img):
 
 class BaseSegment(object):
     @classmethod
-    @nb.jit
     def create_delimiter(cls):
-        return np.array([[np.nan] * cls.ndims])
+        return np.full((1, cls.ndims), np.nan)
 
 
 class UnweightedSegment(BaseSegment):
@@ -393,12 +392,11 @@ class connect_edges(param.ParameterizedFunction):
 
 directly_connect_edges = connect_edges # For bockwards compatibility; deprecated
 
-@nb.jit
+
 def minmax_normalize(X, lower, upper):
     return (X - lower) / (upper - lower)
 
 
-@nb.jit
 def minmax_denormalize(X, lower, upper):
     return X * (upper - lower) + lower
 
--- a/datashader/layout.py
+++ b/datashader/layout.py
@@ -171,7 +171,7 @@ def _merge_points_with_nodes(nodes, points, params):
     return n
 
 
-@nb.jit(nogil=True)
+@nb.jit(forceobj=True)
 def cooling(matrix, points, temperature, params):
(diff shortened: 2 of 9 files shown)


philippjfr:
There's a whole range of compilation issues: https://travis-ci.org/pyviz/datashader/jobs/556465042#L2446

Will try to fix those in this PR.

datashader/bundling.py
```diff
@@ -33,9 +33,9 @@ def distance_between(a, b):
     return (((a[0] - b[0]) ** 2) + ((a[1] - b[1]) ** 2))**(0.5)
 
 
-@nb.jit
+@nb.jit(forceobj=True)
 def resample_segment(segments, new_segments, min_segment_length, max_segment_length, ndims):
```
philippjfr: This function uses numpy array creation and can therefore not be compiled in nopython mode. Maybe could be rewritten?

datashader/bundling.py
```diff
@@ -183,9 +183,8 @@ def get_gradients(img):
 
 class BaseSegment(object):
     @classmethod
-    @nb.jit
     def create_delimiter(cls):
```
philippjfr: No point at all to jit this.

datashader/bundling.py
```diff
@@ -393,12 +392,11 @@ def __call__(self, nodes, edges, **params):
 
 directly_connect_edges = connect_edges # For bockwards compatibility; deprecated
 
-@nb.jit
+
 def minmax_normalize(X, lower, upper):
```
philippjfr: No point jitting these, they already operate on NumPy arrays.

datashader/layout.py
```diff
@@ -171,7 +171,7 @@ def _merge_points_with_nodes(nodes, points, params):
     return n
 
 
-@nb.jit(nogil=True)
+@nb.jit(forceobj=True)
```
philippjfr: Uses numpy array creation and cannot therefore be compiled in nopython mode, could maybe be rewritten?

datashader/resampling.py
```diff
@@ -231,7 +231,7 @@ def _get_fill_value(fill_value, src, out):
     return fill_value
 
 
-@ngjit_parallel
+@ngjit
 def _get_dimensions(src, out):
```
philippjfr: No point jitting this.

datashader/layout.py
philippjfr: To expand a little bit here, the numpy array creation isn't really the problem. The problem is that looplifting also isn't possible due to some type inference issues and presumably the ``.toarray()`` call inside the loop.

datashader/bundling.py
philippjfr: The real problem here is that looplifting is also disabled due to some type inference issues.

philippjfr:
Going to merge this so I can continue working on other PRs. Happy to revert stuff if needed before a new release.

datashader/bundling.py
jbednar: Turns out the underlying issue was that the dtype for `next_point` was int64 in one case and int32 (I think) in the other, which is fixed in https://github.com/pyviz/datashader/pull/796. With this PR alone, the performance was terrible, but that's been fixed in https://github.com/pyviz/datashader/pull/796 .

datashader/layout.py
jbednar: According to Siu, the problem is that `matrix` is a scipy.sparse.csr.csr_matrix, and Numba can't yet speed up sparse matrices.

jbednar: Fixed in https://github.com/pyviz/datashader/pull/763 .

PR #764, opened 2019-07-09
Fix CI by upgrading to Python 3.7

philippjfr:
Upgrades system Python to 3.7 to make latest conda work.


philippjfr:
Looks like the windows errors has been reported before: https://github.com/conda-forge/pytest-benchmark-feedstock/issues/6

Issue #326, opened 2017-04-27
Implement spread and dynspread for aggregates, not just images


FelixLouis:
Both datashader.transfer_functions.spread and datashader.transfer_functions.dynspread have a "how" argument, set by default to "over".

The problem is that there is no explanation in the documentation (http://datashader.readthedocs.io/en/latest/api.html#datashader.transfer_functions.spread) about what values this argument can take and the effects of these values. After looking through the source code I finally found 4 possible values for this argument (https://github.com/bokeh/datashader/blob/master/datashader/composite.py):

 - "over", the default one, which seems to work so that each pixel go over the other ones with a priority set in function of their coordinates.
 - "source", which seems to work the same way (I couldn't determine the difference by testing)
 - "add", which I expected to add the values of the pixels but seems to works differently when a pixel already at the maximum value collides with another one.
 - "saturate", which I expected to take the maximum of the two pixels colliding but has the same weird behavior as "add".

jbednar:
Sorry about that! Clearly an oversight.  The various possible operators are illustrated nicely at https://www.cairographics.org/operators/ .  We have only implemented a subset, as you list above.  Unfortunately, for RGB data there is a fixed upper bound on intensity for each channel, which makes many of the operators behave in unexpected ways.  As a result, none of them are particularly great; you should definitely try to avoid spreading that overlaps in the first place.  If you have any ideas for a better approach, adding a new operator should be trivial...

Actually, based on that link, maybe there is a problem with "over"?  It is meant to be different from "source"...

FelixLouis:
Here is an example of what "over" can yield: If I understand correctly this is supposed to be "source" behavior (here it's a dynspread with a threshold of 1).
![over_behavior](https://cloud.githubusercontent.com/assets/25052300/25527132/9325d946-2c17-11e7-8c8c-98c595cf4470.png)

Thanks to your link I realize that what I would need is to overlap not by adding the colors but the values of the matrix just before .shade convert it to an image. For example I would like a 1px square spread to transform this:
[
[0,0,0,1],
[0,0,0,1],
[0,0,0,0],
[0,1,0,0]
]
into this:
[
[0,0,2,2],
[0,0,2,2],
[1,1,2,1],
[1,1,1,0]
]
so basically I would need to modify the canvas and not the picture. I didn't notice such function, though, is there one I could have missed? If not, is there a more efficient way to do that than the naive approach I can think about which would consist in iterating over the whole canvas with datashader.glyphs.Point?

[...]

PR #771, opened 2019-07-24
WIP: Spread support for aggregate arrays

jbednar:
Previously, spreading has been supported only for Image objects, working directly on RGBA pixels.  As a result, it has not been available for use with the HoloViews rasterizing operation, which returns aggregate arrays rather than Images.

As proposed in https://github.com/holoviz/datashader/issues/326, this PR adds a set of array spreading operators and allows spread to accept either an image or an aggregate array.

Remaining issues:

- [x] Special handling will be needed for categorical aggregates, which are stacks of regular aggregates (e.g. `aggc` in 2_Pipeline.ipynb)
- [x] dynspread is not yet supported for aggregate arrays; changes will be needed to `_density()` in transfer_functions.py
- [x] Spreading for aggregates is currently only supported when Numba is disabled, `composite.arr_operator(f)` doesn't currently have the type-specific Numba compilation code from `composite.operator(f)`, and it will need to handle whatever datatypes are supported for aggregations (i32, i64, f32, f64? Not sure.). Spreading is remarkably slow without Numba!
- [x] How to interpret the various spreading operators for the aggregate case is open for discussion. There are four different operators defined for images, but right now all those operators are defined for aggregates as only one of two different behaviors, with source_arr returning src if truthy, and else dest, and the rest all returning src+dst.  Is this correct/useful/the only possibility?
- [ ] Need to update examples/getting_started/2_Pipeline.ipynb to explain agg spreading. Explaining it will be tricky, because spreading is currently explained at the image level, which is in a different section of the pipeline from aggregates; spreading discussion either needs to move to the aggregate stage and then be expanded later when discussed for images (which have different allowable operations), or the other way around (pointing forward to the image section from the aggregates section).
- [x] Presumably need to update the `spread()` operation in HoloViews to allow it to be used with `rasterize()` output (hv.Image) instead of just `spread()` output (hv.RGB).

### DIFF
--- a/datashader/composite.py
+++ b/datashader/composite.py
@@ -10,7 +10,8 @@ import numba as nb
 import numpy as np
 import os
 
-__all__ = ('composite_op_lookup', 'over', 'add', 'saturate', 'source')
+__all__ = ('composite_op_lookup', 'over', 'add', 'saturate', 'source',
+           'over_arr', 'add_arr', 'saturate_arr', 'source_arr')
 
 
 @nb.jit('(uint32,)', nopython=True, nogil=True, cache=True)
@@ -46,7 +47,7 @@ composite_op_lookup = {}
 
 
 def operator(f):
-    """Define and register a new composite operator"""
+    """Define and register a new image composite operator"""
 
     if jit_enabled:
         f2 = nb.vectorize(f)
@@ -109,3 +110,33 @@ def saturate(src, dst):
     g = (factor * sg + dg * da)/a
     b = (factor * sb + db * da)/a
     return combine_scaled(r, g, b, a)
+
+
+
+def arr_operator(f):
+    """Define and register a new array composite operator"""
+    ## Should be using jit as for operator(f), but will need
+    ## to deal with different agg types (uint32, float32 at least)
+    f2 = np.vectorize(f)
+
+    composite_op_lookup[f.__name__] = f2
+    return f2
+
+@arr_operator
+def source_arr(src, dst):
+    if src:
+        return src
+    else:
+        return dst
+
+@arr_operator
+def over_arr(src, dst):
+    return src + dst
+
+@arr_operator
+def add_arr(src, dst):
+    return src + dst
+
+@arr_operator
+def saturate_arr(src, dst):
+    return src + dst
--- a/datashader/composite.py
+++ b/datashader/composite.py
@@ -114,13 +114,25 @@ def saturate(src, dst):
 
 
 def arr_operator(f):
-    """Define and register a new array composite operator"""
-    ## Should be using jit as for operator(f), but will need
-    ## to deal with different agg types (uint32, float32 at least)
-    f2 = np.vectorize(f)
+    """Define and register a new image composite operator"""
+
+    if jit_enabled:
+        f2 = nb.vectorize(f)
+        f2._compile_for_argtys(
+           (nb.types.int32, nb.types.int32))
+        f2._compile_for_argtys(
+           (nb.types.int64, nb.types.int64))
+        f2._compile_for_argtys(
+            (nb.types.float32, nb.types.float32))
+        f2._compile_for_argtys(
+            (nb.types.float64, nb.types.float64))
+        f2._frozen = True
+    else:
+        f2 = np.vectorize(f)
 
     composite_op_lookup[f.__name__] = f2
     return f2
+
 
 @arr_operator
 def source_arr(src, dst):
--- a/datashader/composite.py
+++ b/datashader/composite.py
@@ -137,36 +137,40 @@ def arr_operator(f):
 
 @arr_operator
 def source_arr(src, dst):
-    if src:
-        return src
-    else:
-        return dst
+    if np.isnan(src): return dst
+    if np.isnan(dst): return src
 
-@arr_operator
-def over_arr(src, dst):
-    return src + dst
+    if src:  return src
+    else:    return dst
 
 @arr_operator
 def add_arr(src, dst):
+    if np.isnan(src): return dst
+    if np.isnan(dst): return src
     return src + dst
 
 @arr_operator
 def saturate_arr(src, dst):
+    if np.isnan(src): return dst
+    if np.isnan(dst): return src
     return src + dst
 
 @arr_operator
 def max_arr(src, dst):
+    if np.isnan(src): return dst
+    if np.isnan(dst): return src
     return max([src,  dst])
 
 @arr_operator
-def maxabs_arr(src, dst):
-    if abs(src) > abs(dst):
-        return src
-    else:
-        return dst
+def over_arr(src, dst):
+    if np.isnan(src): return dst
+    if np.isnan(dst): return src
+    return src + dst
 
 @arr_operator
 def min_arr(src, dst):
+    if np.isnan(src): return dst
+    if np.isnan(dst): return src
     return min([src,  dst])
 
 
--- a/datashader/composite.py
+++ b/datashader/composite.py
@@ -11,7 +11,8 @@ import numpy as np
 import os
 
 __all__ = ('composite_op_lookup', 'over', 'add', 'saturate', 'source',
-           'over_arr', 'add_arr', 'saturate_arr', 'source_arr')
+           'over_arr', 'add_arr', 'saturate_arr', 'source_arr',
+           'max_arr', 'min_arr', 'maxabs_arr')
 
 
 @nb.jit('(uint32,)', nopython=True, nogil=True, cache=True)
@@ -152,3 +153,20 @@ def add_arr(src, dst):
 @arr_operator
 def saturate_arr(src, dst):
     return src + dst
+
+@arr_operator
+def max_arr(src, dst):
+    return max([src,  dst])
+
+@arr_operator
+def maxabs_arr(src, dst):
+    if abs(src) > abs(dst):
+        return src
+    else:
+        return dst
+
+@arr_operator
+def min_arr(src, dst):
+    return min([src,  dst])
+
+
--- a/datashader/composite.py
+++ b/datashader/composite.py
@@ -137,40 +137,27 @@ def arr_operator(f):
 
 @arr_operator
 def source_arr(src, dst):
-    if np.isnan(src): return dst
-    if np.isnan(dst): return src
-
     if src:  return src
     else:    return dst
 
 @arr_operator
 def add_arr(src, dst):
-    if np.isnan(src): return dst
-    if np.isnan(dst): return src
     return src + dst
 
 @arr_operator
 def saturate_arr(src, dst):
-    if np.isnan(src): return dst
-    if np.isnan(dst): return src
     return src + dst
 
 @arr_operator
 def max_arr(src, dst):
-    if np.isnan(src): return dst
-    if np.isnan(dst): return src
     return max([src,  dst])
 
 @arr_operator
 def over_arr(src, dst):
-    if np.isnan(src): return dst
-    if np.isnan(dst): return src
     return src + dst
 
 @arr_operator
 def min_arr(src, dst):
-    if np.isnan(src): return dst
-    if np.isnan(dst): return src
     return min([src,  dst])
 
 
--- a/datashader/composite.py
+++ b/datashader/composite.py
@@ -11,8 +11,7 @@ import numpy as np
 import os
 
 __all__ = ('composite_op_lookup', 'over', 'add', 'saturate', 'source',
-           'over_arr', 'add_arr', 'saturate_arr', 'source_arr',
-           'max_arr', 'min_arr', 'maxabs_arr')
+           'over_arr', 'add_arr', 'saturate_arr', 'source_arr', 'max_arr', 'min_arr')
 
 
 @nb.jit('(uint32,)', nopython=True, nogil=True, cache=True)
--- a/datashader/composite.py
+++ b/datashader/composite.py
@@ -114,7 +114,7 @@ def saturate(src, dst):
 
 
 def arr_operator(f):
-    """Define and register a new image composite operator"""
+    """Define and register a new array composite operator"""
 
     if jit_enabled:
         f2 = nb.vectorize(f)
@@ -158,5 +158,4 @@ def over_arr(src, dst):
 @arr_operator
 def min_arr(src, dst):
     return min([src,  dst])
-
 
--- a/datashader/composite.py
+++ b/datashader/composite.py
@@ -10,8 +10,20 @@ import numba as nb
 import numpy as np
 import os
 
-__all__ = ('composite_op_lookup', 'over', 'add', 'saturate', 'source',
-           'over_arr', 'add_arr', 'saturate_arr', 'source_arr', 'max_arr', 'min_arr')
+image_operators = ('over', 'add', 'saturate', 'source')
+array_operators = ('add_arr', 'max_arr', 'min_arr', 'source_arr')
+__all__ = ('composite_op_lookup', 'validate_operator') + image_operators + array_operators
+
+
+def validate_operator(how, is_image):
+    name = how if is_image else how + '_arr'
+    if is_image:
+        if name not in image_operators:
+            raise ValueError('Operator %r not one of the supported image operators: %s'
+                            % (how, ', '.join(repr(el) for el in image_operators)))
+    elif name not in array_operators:
+        raise ValueError('Operator %r not one of the supported array operators: %s'
+                        % (how, ', '.join(repr(el[:-4]) for el in array_operators)))
 
 
 @nb.jit('(uint32,)', nopython=True, nogil=True, cache=True)
@@ -144,16 +156,8 @@ def add_arr(src, dst):
     return src + dst
 
 @arr_operator
-def saturate_arr(src, dst):
-    return src + dst
-
-@arr_operator
 def max_arr(src, dst):
     return max([src,  dst])
-
-@arr_operator
-def over_arr(src, dst):
-    return src + dst
 
 @arr_operator
 def min_arr(src, dst):
--- a/datashader/transfer_functions/__init__.py
+++ b/datashader/transfer_functions/__init__.py
@@ -730,19 +730,17 @@ def dynspread(img, threshold=0.5, max_px=3, shape='circle', how='over', name=Non
         out = spread(img, px, shape=shape, how=how, name=name)
         if is_image:
             density = _rgb_density(out.data)
-        elif out.data.dtype in [np.int32, np.int64]:
-            cast = out.data.astype(np.float64)
-            cast[out.data == 0] = np.nan
-            density = _array_density(cast)
         else:
-            density = _array_density(out.data)
+            float_type = out.dtype in [np.float32, np.float64]
+            density = _array_density(out.data, float_type)
         if density >= threshold:
             break
+
     return out
 
 
 @nb.jit(nopython=True, nogil=True, cache=True)
-def _array_density(arr):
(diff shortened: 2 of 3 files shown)


jbednar:
See examples at https://anaconda.org/jbednar/2_pipeline-aggspread .

jbednar:
@jlstevens , please also consider more fundamental changes instead or in addition; e.g. can we take a different approach to making isolated points visible that has better properties than spreading (e.g. would a simple convolution with a box filter work?).

jbednar:
Also consider:

- [ ] Should the operators be normalized, to make e.g. `add` into `mean`?  Seems more useful in general than having to hit overflow limits.

ltalirz:
This would be really useful, it would potentially also make https://github.com/holoviz/holoviews/issues/2102 less of an issue for people who need a colorbar together with spread

jlstevens:
It is hard for me to know what the best tradeoffs are, but what I can do is list the pros and cons of different approaches. I'll list the following proposals: convolution (normalized) and convolution (not normalized):

## Convolution (normalized)

*Pros*

 - Mathematically defined, preserves the total weight across the input.
 - Flexible as you can use different kernel shapes.

*Cons*

 - By spreading the weight around, you change the absolute values. E.g the hover tool hovering over a value of 1.0 before spreading will show a lower value at that point after spreading.

    *Note*: *Visually* the increase in spatial extent should make things easier to see and auto-ranging for e.g colormaps/colorbars can compensate for the change in absolute values.

## Convolution (not normalized)

*Pros*

  - Allowing non normalized kernels is a superset of normalized kernels so you have extra flexibility.   In other words, you can accept any kernel which may or may not be normalized.
  - You can use kernels that keep the center pixel the same and simply increase the weight of surrounding pixels accordingly (e.g to the same absolute value in the case of a box filter). This means that you can hover over a larger area (the area spread out by the kernel) and still see the original value with a hover tool (for instance).
  
*Cons*

  - If not normalized, the result is not as well mathematically defined. The overall weight may go up or down.

Thinking about the two cases above, the best choice (in my opinion!) is to offer spread as a convolution with 1) a spatial box kernel by default that can be overridden with any custom kernel 2) a boolean to normalize the kernel that is off by default. By a spatial box kernel, I mean a kernel that copies the value of the central pixel to surrounding pixels in a spatially uniform way (e.g as an approximation to a circle or as a square).

One operation could then offer the user with the available tradeoffs. The default aims to keep hover information correct in value (but available over a larger space), making things easier to see. Enabling normalization is then optional and allows the user to preserve the overall weight in the input if they wish. Using custom kernels (e.g gaussians) with normalization, the user can then use this operation to do mathematically correct image blurring (for instance).

No matter how this operation is used, the user needs to understand how it is distorting the original data for the sake of visual clarity. The suggested defaults should result in a purely spatial distortion (i.e giving tiny points a spatial extent) which I think is probably what most people want in most cases.

jlstevens:
Numba is now enabled for spreading on aggregates, namely `int32`, `int64`, `float32` and `float64`. Support for RGBs is preserved and spreading on the integer types matches the previous results (after colormapping, obviously!). Spreading on floats gives different enough results that some investigation will be needed there.

> How to interpret the various spreading operators for the aggregate case is open for discussion. There are four different operators defined for images, but right now all those operators are defined for aggregates as only one of two different behaviors, with source_arr returning src if truthy, and else dest, and the rest all returning src+dst. Is this correct/useful/the only possibility?

I've been thinking about this and I don't think the spreading operators (e.g `over` and `saturate` etc) make much sense outside an RGB context. I think your suggestion is the best we will be able to do with this model and that `add` is the most intuitive default option when working with aggregate arrays.

My next step is to get this working with `dynspread`.

datashader/composite.py
```diff
@@ -109,3 +110,45 @@ def saturate(src, dst):
...
+def over_arr(src, dst):
+    return src + dst
+
+@arr_operator
+def add_arr(src, dst):
+    return src + dst
+
+@arr_operator
+def saturate_arr(src, dst):
+    return src + dst
```
jbednar: Seems like for arrays we also need max and min operators.  Maybe source, over, or saturate could be most appropriately interpreted as one or both of those?  If not, maybe need to add separate max and min operators, presumably both for images and for arrays.

datashader/transfer_functions.py
```diff
@@ -412,8 +412,9 @@ def spread(img, px=1, shape='circle', how='over', mask=None, name=None):
         Optional string name to give to the Image object to return, 
         to label results for display.
     """
-    if not isinstance(img, Image):
-        raise TypeError("Expected `Image`, got: `{0}`".format(type(img)))
+    if not isinstance(img, xr.DataArray):
+        raise TypeError("Expected `xr.DataArray`, got: `{0}`".format(type(img)))
+    is_image = isinstance(img, Image)
```
jbednar: Be sure to test with categorical arrays; those will have is_image be False but need a loop to iterate over the Z dimension, presumably spreading in each one separately?  Or maybe they should spread analogously to RGB, where instead of 3 layers R,G,B they have arbitrarily many layers, all spread at once?  Seems tricky.

datashader/composite.py
jlstevens: This is leading me to believe that maybe we do some validation about the operators used. We can default to `add` which makes sense for RGB and aggregate arrays and then complain if an unsuitable operator is chosen for the supplied data type. 

Then we could offer `min` and `max` operators for use with aggregate arrays only instead of trying to map these two concepts back to the set of operators that make sense with RGB.

jlstevens:
@jbednar @philippjfr I have pushed an update to how density is computed for dynspread to allow it to support float and int aggregates. 

I don't think the results aren't quite right *yet* but I think that is down to fixes needed in `spread` and not due to this last set of updates to `dynspread`. I'll investigate shortly but for now this should be enough to allow Philipp to check that these dynspread updates will work with the new HoloViews release.

philippjfr:
Could you rebase the PR or do you mind if I push up my rebase? It's not usable with HoloViews in its current state.

philippjfr:
@jbednar asked me to do it, so make sure to `git reset origin/arr_spreading --hard` before continuing.

jlstevens:
I just notices that this PR doesn't have an images to show the effect of the new aggregate based spreading. Here is the result of spread (input on the left and spread output on the right) where `agg` is an integer aggregate array:

```python
int_spread = tf.spread(agg, name="spread 1px")
hv.Image(agg).opts(cmap='Blues', clim=(0, 20)) + hv.Image(int_spread).opts(cmap='Blues', clim=(0, 20))
```
</img>

And here is the result of spread (input on the left and spread output on the right) where `aggf` is an float aggregate array:

```python
float_spread = tf.spread(aggf, name="spread 1px")
hv.Image(agg).opts(cmap='Blues', clim=(0, 20)) + hv.Image(float_spread).opts(cmap='Blues', clim=(0, 20))
```

</img>

Being able to apply spread on aggregate arrays will allow colormapping with Bokeh which isn't possible when working in RGB space.

jlstevens:
Although [numba stencils](https://numba.pydata.org/numba-doc/latest/user/stencil.html) looked like an ideal fit for implementing the spread operation, the spreading implementation [earlier in this PR](https://github.com/holoviz/datashader/blob/ba1e1053346acc43fd30b343234fcc96e5ce1a53/datashader/transfer_functions/__init__.py#L542) which did work correctly had major performance issues (~10x slower than using manual loops).

I have now gone back to explicit loops, defining a separate float and integer kernel which restores the performance. Now, spreading on uint32 is as fast as the original RGB spreading implementation (very slightly faster in fact) and the operation remains fast for the newly supported dtypes.

This PR will be ready once I've added some tests and I'll file an issue on the numba repo to report the performance issue I've encountered while using stencils.

jlstevens:
Looks like I'll have to rebase this PR once https://github.com/holoviz/datashader/pull/948 is ready to run the tests properly.

jlstevens:
@jbednar Ready for review/merge.

jlstevens:
@jbednar Here is an example of categorical spreading from the census example (showing Chinatown):

`px=0`:
image

`px=1`:
image

`px=2`:
image

datashader/composite.py
```diff
@@ -46,7 +47,7 @@ def combine_scaled(r, g, b, a):
 
 
 def operator(f):
```
jbednar, line 49:
```suggestion
def img_operator(f):
```
?

datashader/composite.py
```diff
@@ -109,3 +110,53 @@ def saturate(src, dst):
     g = (factor * sg + dg * da)/a
     b = (factor * sb + db * da)/a
     return combine_scaled(r, g, b, a)
+
+
+
+def arr_operator(f):
+    """Define and register a new image composite operator"""
```
jbednar, line 117:
```suggestion
    """Define and register a new array composite operator"""
```

datashader/transfer_functions/__init__.py
```diff
@@ -544,7 +544,7 @@ def spread(img, px=1, shape='circle', how='over', mask=None, name=None):
 
     Parameters
     ----------
-    img : Image
+    img : Image or DataArray
```
jbednar, line 547:
```suggestion
    img : Image or other DataArray
```

datashader/composite.py
jbednar: Is this resolved, then?

datashader/transfer_functions/__init__.py
```diff
@@ -574,20 +575,106 @@ def spread(img, px=1, shape='circle', how='over', mask=None, name=None):
...
+        kernel = _build_spread_kernel(how, is_image)
+    elif float_type:
+        kernel = _build_float_kernel(how, w)
+    else:
+        kernel = _build_int_kernel(how, w, img.dtype == np.uint32)
+
+    if is_image:
+        M, N = img.shape
+        buf = np.zeros((M + 2*extra, N + 2*extra),
+                       dtype='uint32' if is_image else img.dtype)
```
jbednar, line 593: `is_image` is always true in this branch, so it will never use `else img.dtype`. But is there some reason to hardcode it here to `uint32`?  Won't `img.dtype` already be `uint32`?  I think it's better to use `img.dtype` consistently instead of hardcoding it, but maybe you had some reason.

datashader/transfer_functions/__init__.py
```diff
@@ -574,20 +575,106 @@ def spread(img, px=1, shape='circle', how='over', mask=None, name=None):
...
+        kernel(img.data, mask, buf)
+        out = buf[extra:-extra, extra:-extra].copy()
+    elif float_type:
+        M, N = img.shape
+        buf = np.full((M + 2*extra, N + 2*extra), np.nan, dtype=img.dtype)
```
jbednar, line 598: Please pull out the shape calculation to before this `if` statement to avoid repeating it 5 times:

```
shape = (M + 2*extra, N + 2*extra)
```

Also consider whether it's useful to eliminate the other repeated code in these branches to make it more readable. Maybe not, but probably?

datashader/transfer_functions/__init__.py
```diff
@@ -574,20 +575,106 @@ def spread(img, px=1, shape='circle', how='over', mask=None, name=None):
...
+            else:
+                buf = np.zeros((M + 2*extra, N + 2*extra), dtype=layer.dtype)
+            kernel(layer.data, mask, buf)
+            layers.append(buf[extra:-extra, extra:-extra].copy())
+            out = np.dstack(layers)
+
+    if is_image:
+        return Image(out, dims=img.dims, coords=img.coords, name=name)
+    else:
+        return xr.DataArray(out, dims=img.dims, coords=img.coords, name=name)
```
jbednar, line 622:
```suggestion
   return img.__class__(out, dims=img.dims, coords=img.coords, name=name)
```
?

Or at least:

```suggestion
    class_ = Image if is_image else xr.DataArray
    return class_(out, dims=img.dims, coords=img.coords, name=name)
```

datashader/transfer_functions/__init__.py
```diff
@@ -574,20 +575,106 @@ def spread(img, px=1, shape='circle', how='over', mask=None, name=None):
...
+        M, N = arr.shape
+        for y in range(M):
+            for x in range(N):
+                el = arr[y, x]
+                for i in range(mask_size):
+                    for j in range(mask_size):
+                        if mask[i, j]:
+                            if ignore_zeros and el==0:
+                                result = out[i + y, j + x]
+                            elif ignore_zeros  and out[i + y, j + x]==0:
```
jbednar, line 641:
```suggestion
                            elif ignore_zeros and out[i + y, j + x]==0:
```

datashader/composite.py
jlstevens: I think the idea of 'saturate' being `src+dst` might be a bit iffy but the 'min' and 'max' operators are now offered and I think *those* semantics are clear at least. I suppose the issue is whether we want to support the same operator names as RGB? Having to support some operators for RGB only and a different set for arrays would be a pain so this kind of behavior isn't totally unreasonable imho as long as it is described in the docs.

datashader/transfer_functions.py
jlstevens: This is now resolved: categorical aggregates are supported by spreading each category layer by the same amount.

datashader/composite.py
jbednar: If we leave `saturate` and `over` defined for arrays, then we'd presumably not mention them in the docs, but I'm inclined to delete them if they don't make sense for arrays.  That would mean changing the default `how` to None, with code selecting `add` if it's an array and `over` if it's an Image, I guess?

datashader/transfer_functions/__init__.py
jlstevens: I think I agree: that code was left over from an earlier codepath that is no longer in used. I'll using `img.dtype`.

datashader/transfer_functions/__init__.py
jlstevens: I agree on pulling out `padded_shape` but I don't think it is clearer to try and pull out much more than that. I find that tackling each dtype separately makes the code easier to think about.

jlstevens: Addressed in 9facd5c

jbednar: Sure, just look at it once you've got the shape out and see if anything else jumps out as being better to centralize.  I considered whether the first `if` should be about all those with `len(img.shape)==2`, then only making minor adjustments for type that would then jump out at the reader, but I don't know if that's really going to be easier to read or not without writing it.

datashader/transfer_functions/__init__.py
jlstevens: Addressed in 9facd5c

datashader/composite.py
jlstevens: I'm not sure about this one. If anything, I would like to remove the RGB spreading support entirely so `operator` would stay but only apply to arrays, not RGB.

datashader/composite.py
jlstevens: I don't think `how=None` is particularly clear as there *is* a default 'how' which (in my opinion) ought to be explicit. Tricky! Maybe we could do the following?

1) As we want to encourage array spreading instead of the questionable RGB spreading, I think the default should be driven by that choice. So the default should be `how='add'` but there is already an image operator named 'add'...
2) ... which means I could rename the array `add` operator to `'sum'`? I think 'source' is the only operator that really makes sense for both images and array cases.
3) So the default could be `how='sum'` in which case image inputs could warn the user that they are switching to `how='over'` for backwards compatibility?
4) Given a default that works for both images and arrays, I think we should then do some validation. Images can then only use `source`, `over`, `saturate` and `add` which arrays can only use `source`, `sum`, `min` and `max`

Would this make sense?

jbednar: I think that will be confusing, as there is no semantic difference between `add` and `sum`.  Warning that we'll switch over to `over` for backwards compatibility seems like a pain for users; there's nothing wrong with their current code, and they should expect it to keep working with no warnings unless we wished to warn about using RGB spreading at all (which even if we did wouldn't be for another 6 months at the earliest given how long RGB spreading has been the only spreading available!).  I do think we should do validation, but I find the issues with changing the default to None much less than the issues with these other alternatives.

jlstevens: Ok...we can do that but we will need to state that the default 'how' is 'over' and 'add' for images and arrays respectively both in the docstrings and the online docs.

datashader/composite.py
jbednar: Changing the semantics but keeping the same name seems more confusing than changing the name and keeping the same semantics.  I guess just leave it as-is and someday there will only be arr_operator.

datashader/composite.py
jbednar: Sure, but online docs might not even mention any details about RGB spreading, so it might not come up there at all...

datashader/transfer_functions/__init__.py
```diff
@@ -574,20 +577,104 @@ def spread(img, px=1, shape='circle', how='over', mask=None, name=None):
...
+        kernel = _build_float_kernel(how, w)
+    else:
+        kernel = _build_int_kernel(how, w, img.dtype == np.uint32)
+
+    if is_image:
+        buf = np.zeros(padded_shape, dtype=img.dtype)
+        kernel(img.data, mask, buf)
+        out = buf[extra:-extra, extra:-extra].copy()
+    elif float_type:
+        M, N = img.shape
```
jbednar, line 601:
```suggestion
```

datashader/transfer_functions/__init__.py
```diff
@@ -574,20 +577,104 @@ def spread(img, px=1, shape='circle', how='over', mask=None, name=None):
...
+        buf = np.full(padded_shape, np.nan, dtype=img.dtype)
+        kernel(img.data, mask, buf)
+        out = buf[extra:-extra, extra:-extra].copy()
+    elif len(img.shape)==2:
+        M, N = img.shape
```
jbednar, line 606:
```suggestion
```

datashader/transfer_functions/__init__.py
```diff
@@ -574,20 +577,102 @@ def spread(img, px=1, shape='circle', how='over', mask=None, name=None):
...
+        layers = []
+        for category in range(img.shape[2]):
+            layer = img[:,:,category]
+            if float_type:
+                buf = np.full(padded_shape, np.nan, dtype=img.dtype)
+            else:
+                buf = np.zeros(padded_shape, dtype=layer.dtype)
+            kernel(layer.data, mask, buf)
+            layers.append(buf[extra:-extra, extra:-extra].copy())
+            out = np.dstack(layers)
```
jbednar, line 618:
```suggestion
        out = np.dstack(layers)
```

Seems like it would have worked, but with the previous indentation it would have overwritten `out` every time a layer is processed, which probably was not intended?)

jbednar: This is totally untested pseudocode, but I think this whole section of the code would be much more readable and maintainable if it's boiled down to what it actually does:

```
    fill_value = np.nan if float_type else 0

    def apply_kernel(layer):
        buf = np.full(padded_shape, fill_value, dtype=layer.dtype)
        kernel(layer.data, mask, buf)
        return buf[extra:-extra, extra:-extra].copy()
        
    if len(img.shape)==2:
        out = apply_kernel(img)
    else:
        out = np.dstack([apply_kernel(img[:,:,category])
                        for category in range(img.shape[2])]
```

Otherwise we can't see that what's happening in all these cases is quite similar, with only trivial differences that are hard to spot otherwise.

jbednar:
Thanks, @jlstevens, for hitting this home!

PR #776, opened 2019-08-10
Split glyph type definitions into separate files

jonmmease:
This is a maintenance PR that splits the prior `datashader/glyphs.py` file, which was getting rather large, into one file per glyph type. The PR contains no logic changes beyond moving classes/functions and updating imports.

I'd wanted to get this done in its own PR before adding additional glyph types.

cc @jbednar @philippjfr


philippjfr:
Very much in favor of this. Would love to hear your thoughts on whether you think gridded types (i.e. non-tabular) types could be folded into the glyph based code, in their current form both ``raster`` and ``quadmesh`` are handled completely separately.

jonmmease:
> Would love to hear your thoughts on whether you think gridded types (i.e. non-tabular) types could be folded into the glyph based code

I think so, I think what I'll do next is try to port the logic from your quad mesh PR (https://github.com/pyviz/datashader/pull/769) into a glyph structure and see how well it will fit.  Sound good?

jonmmease:
Hmm, I think there are some test issues cropping up from the pandas 0.25 update.

philippjfr:
I'll merge for now, compatibility with the pandas 0.25 should happen in another PR.

jonmmease:
Thanks @philippjfr, I'll take a look at the pandas 0.25 situation tomorrow

jbednar:
Sounds good, thanks!

PR #777, opened 2019-08-12
Pandas 0.25 test suite compatibility

jonmmease:
Removed pandas<0.25.0 pin and update datatype tests

Compatibility with Pandas 0.25.0 only required test suite updates, so the past releases of datashader should be fine with pandas 0.25.0.


philippjfr:
Looks good to me.

jonmmease:
Thanks!

PR #778, opened 2019-08-12
Refactor compiler's make_append to handle more than one dimension

jonmmease:
This PR is a small refactor to the `make_append` function in the compiler module.  Previously, this function assumed that the input data structure for all glyphs was 1-dimensions (e.g. a DataFrame).  Because `trimesh` doesn't fit this assumption, there was special-casing logic in `make_append` to check if the glyph was a `Triangle`.

With this PR, each glyph has an `ndims` property that indicates the number of dimensions required to index into the data structure that the glyph is built from.  Currently, this value is 1 for all of the glyph except Triangle, in which case it is `None`.

When quadmesh is added, it will have an `ndims` value of 2. If the raster functionality is ported to the glyph framework, then it would have an `ndims` value of 2 as well.

### DIFF
--- a/datashader/compiler.py
+++ b/datashader/compiler.py
@@ -7,7 +7,6 @@ import numpy as np
 import xarray as xr
 
 from .compatibility import _exec
-from .glyphs import Triangles
 from .reductions import summary
 from .utils import ngjit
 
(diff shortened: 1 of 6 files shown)


jbednar:
Looks good, thanks!

PR #779, opened 2019-08-13
[WIP] Add quadmesh glyph

jonmmease:
This PR is an alternative implementation of quadmesh rasterization, roughly based on the logic from https://github.com/pyviz/datashader/pull/769.  So far only the rectilinear case is implemented.

Unlike that PR, this PR adds quadmesh glyph classes, and supports the standard datashader aggregation framework.

Overall, the architecture fits well with that of the dataframe-based glyphs (points, line, and area). And relying on the datashader aggregation framework results in a lot less code duplication compared to https://github.com/pyviz/datashader/pull/769.

**But**, it's an order of magnitude slower than the implementations in https://github.com/pyviz/datashader/pull/769.  Here is a notebook showing some timing results: https://anaconda.org/jonmmease/rectquadmesh_examples/notebook.

Roughly speaking, this PR is ~13x faster than representing a rectilinear quadmesh with a trimesh. But the specialized implementation from https://github.com/pyviz/datashader/pull/769 is ~13 faster than this PR.  Note that I disabled numba parallelization for these tests for consistency

I did some performance debugging and I found that nearly all of the extra overhead in this PR, compared to the specialized implementation, is coming from the use of the aggregation framework.  If in the `_extend` function, I don't call `append` but instead implement a single aggregation then the performance is comparable to the specialized implementations.

So the bad news is that right now we need to choose between performance and consistency/maintainability for the quadmesh implementation.  The good news is that there may be an order of magnitude speedup to be had across `points`, `line`, and `area` glyphs as well if we can work out how to optimize the aggregation framework.

@jbednar @philippjfr

### DIFF
--- a/datashader/glyphs/quadmesh.py
+++ b/datashader/glyphs/quadmesh.py
@@ -1,0 +1,123 @@
+from toolz import memoize
+
+from datashader.glyphs.glyph import Glyph
+from datashader.resampling import infer_interval_breaks
+from datashader.utils import isreal, ngjit
+
+
+class _QuadMeshLike(Glyph):
+    def __init__(self, x, y, name):
+        self.x = x
+        self.y = y
+        self.name = name
+
+    @property
+    def ndims(self):
+        return 2
+
+    @property
+    def inputs(self):
+        return (self.x, self.y, self.name)
+
+    def validate(self, in_dshape):
+        if not isreal(in_dshape.measure[str(self.x)]):
+            raise ValueError('x must be real')
+        elif not isreal(in_dshape.measure[str(self.y)]):
+            raise ValueError('y must be real')
+        elif not isreal(in_dshape.measure[str(self.name)]):
+            raise ValueError('aggregate value must be real')
+
+    @property
+    def x_label(self):
+        return self.x
+
+    @property
+    def y_label(self):
+        return self.y
+
+
+class QuadMeshRectilinear(_QuadMeshLike):
+    def _compute_bounds_from_1d_centers(self, xr_ds, dim):
+        vals = xr_ds[dim].values
+
+        # Assume dimension is sorted in ascending or descending order
+        v0, v1, v_nm1, v_n = [
+            vals[i] for i in [0, 1, -2, -1]
+        ]
+
+        # Check if we should swap order
+        if v_n < v0:
+            v0, v1, v_nm1, v_n = v_n, v_nm1, v1, v0
+
+        bounds = (v0 - 0.5 * (v1 - v0), v_n + 0.5 * (v_n - v_nm1))
+        return self.maybe_expand_bounds(bounds)
+
+    def compute_x_bounds(self, xr_ds):
+        return self._compute_bounds_from_1d_centers(xr_ds, self.x)
+
+    def compute_y_bounds(self, xr_ds):
+        return self._compute_bounds_from_1d_centers(xr_ds, self.y)
+
+    @memoize
+    def _build_extend(self, x_mapper, y_mapper, info, append):
+        x_name = self.x
+        y_name = self.y
+
+        @ngjit
+        def _extend(vt, bounds, xs, ys, *aggs_and_cols):
+            sx, tx, sy, ty = vt
+            xmin, xmax, ymin, ymax = bounds
+
+            # Compute max valid x and y index
+            xmaxi = int(round(x_mapper(xmax) * sx + tx))
+            ymaxi = int(round(y_mapper(ymax) * sy + ty))
+
+            for i in range(len(xs) - 1):
+                for j in range(len(ys) - 1):
+                    x0, x1 = max(xs[i], xmin), min(xs[i + 1], xmax)
+                    y0, y1 = max(ys[j], ymin), min(ys[j + 1], ymax)
+
+                    # Makes sure x0 <= x1 and y0 <= y1
+                    if x0 > x1:
+                        x0, x1 = x1, x0
+                    if y0 > y1:
+                        y0, y1 = y1, y0
+
+                    # check whether we can skip quad. To avoid overlapping
+                    # quads, skip if upper bound equals viewport lower bound.
+                    if x1 <= xmin or x0 > xmax or y1 <= ymin or y0 > ymax:
+                        continue
+
+                    # Map onto pixels and clip to viewport
+                    x0i = max(int(x_mapper(x0) * sx + tx), 0)
+                    x1i = min(int(x_mapper(x1) * sx + tx), xmaxi)
+                    y0i = max(int(y_mapper(y0) * sy + ty), 0)
+                    y1i = min(int(y_mapper(y1) * sy + ty), ymaxi)
+
+                    # Make sure single pixel quads are represented
+                    if x0i == x1i and x1i < ymaxi:
+                        x1i += 1
+
+                    if y0i == y1i and y1i < ymaxi:
+                        y1i += 1
+
+                    # x1i and y1i are not included in the iteration. this
+                    # serves to avoid overlapping quads and it avoids the need
+                    # for special handling of quads that end on exactly on the
+                    # upper bound.
+                    for xi in range(x0i, x1i):
+                        for yi in range(y0i, y1i):
+                            append(j, i, xi, yi, *aggs_and_cols)
+
+        def extend(aggs, xr_ds, vt, bounds):
+            # Convert from bin centers to interval edges
+            xs = infer_interval_breaks(xr_ds[x_name].values)
+            ys = infer_interval_breaks(xr_ds[y_name].values)
+            cols = aggs + info(xr_ds.transpose(y_name, x_name))
+            _extend(vt, bounds, xs, ys, *cols)
+
+        return extend
+
+
+class QuadMeshCurvialinear(_QuadMeshLike):
+    pass
--- a/datashader/tests/test_quadmesh.py
+++ b/datashader/tests/test_quadmesh.py
@@ -1,0 +1,127 @@
+import numpy as np
+from numpy import nan
+import xarray as xr
+import datashader as ds
+
+
+def test_rect_quadmesh_autorange():
+    c = ds.Canvas(plot_width=8, plot_height=4)
+    da = xr.DataArray(
+        [[1, 2, 3, 4],
+         [5, 6, 7, 8]],
+        coords=[('b', [1, 2]),
+                ('a', [1, 2, 3, 4])],
+        name='Z')
+
+    y_coords = np.linspace(0.75, 2.25, 4)
+    x_coords = np.linspace(0.75, 4.25, 8)
+    out = xr.DataArray(np.array(
+        [[1., 1., 2., 2., 3., 3., 4., 4.],
+         [1., 1., 2., 2., 3., 3., 4., 4.],
+         [5., 5., 6., 6., 7., 7., 8., 8.],
+         [5., 5., 6., 6., 7., 7., 8., 8.]],
+        dtype='i4'),
+        coords=[('b', y_coords),
+                ('a', x_coords)]
+    )
+
+    res = c.quadmesh(da, x='a', y='b', agg=ds.sum('Z'))
+    assert res.equals(out)
+
+    # Check transpose gives same answer
+    res = c.quadmesh(da.transpose('a', 'b'), x='a', y='b', agg=ds.sum('Z'))
+    assert res.equals(out)
+
+
+def test_rect_quadmesh_autorange_reversed():
+    c = ds.Canvas(plot_width=8, plot_height=4)
+    da = xr.DataArray(
+        [[1, 2, 3, 4],
+         [5, 6, 7, 8]],
+        coords=[('b', [-1, -2]),
+                ('a', [-1, -2, -3, -4])],
+        name='Z')
+
+    y_coords = np.linspace(-2.25, -0.75, 4)
+    x_coords = np.linspace(-4.25, -0.75, 8)
+    out = xr.DataArray(np.array(
+        [[8., 8., 7., 7., 6., 6., 5., 5.],
+         [8., 8., 7., 7., 6., 6., 5., 5.],
+         [4., 4., 3., 3., 2., 2., 1., 1.],
+         [4., 4., 3., 3., 2., 2., 1., 1.]],
+        dtype='i4'),
+        coords=[('b', y_coords),
+                ('a', x_coords)]
+    )
+
+    res = c.quadmesh(da, x='a', y='b', agg=ds.sum('Z'))
+    print(res)
+    assert res.equals(out)
+
+    # Check transpose gives same answer
+    res = c.quadmesh(da.transpose('a', 'b'), x='a', y='b', agg=ds.sum('Z'))
+    assert res.equals(out)
+
+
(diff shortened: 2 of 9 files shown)


jbednar:
That's both exciting and alarming!  Can you expand on "disabled numba parallelization for these tests for consistency"?  Given that the final implementation will use Numba, are comparisons without Numba meaningful here?

jonmmease:
> Can you expand on "disabled numba parallelization for these tests for consistency"?

@philippjfr's implementation uses the numba parallel `prange` loops (https://numba.pydata.org/numba-doc/latest/user/parallel.html#explicit-parallel-loops), but the current datashader glyphs do not.  I tried turning it on for this PR, but got a numba error that I didn't spend time diagnosing.

philippjfr:
I think unifying the dispatch/aggregation pipeline and internal APIs is really important and I'm really excited about this but going from a 600-700x to a 13x speedup would indeed be pretty sad :(

jbednar:
Well, it wouldn't be _that_ drastic; presumably the Numba issue can be worked around, which would leave it off by a single factor of 13, right?

datashader/tests/test_quadmesh.py
```diff
@@ -0,0 +1,128 @@
...
+         [8., 8., 7., 7., 6., 6., 5., 5.],
+         [4., 4., 3., 3., 2., 2., 1., 1.],
+         [4., 4., 3., 3., 2., 2., 1., 1.]],
+        dtype='i4'),
+        coords=[('b', y_coords),
+                ('a', x_coords)]
+    )
+
+    res = c.quadmesh(da, x='a', y='b', agg=ds.sum('Z'))
+    print(res)
```
philippjfr: Stray print.

jonmmease:
I was only seeing a factor of 2-3x speedup with parallelization in @philippjfr 's code (on a machine with many more cores than that). But, yeah, we should figure out where we can use this for the glyph generation code.

philippjfr:
I'm very confused why there is such a huge performance difference. Numba should automatically inline the append function. I'll play around with it and maybe ask Val about it when he returns next week.

jonmmease:
Yeah, I think it's largely about inlining. And I don't have intuition of when things get inlined.  The numba docs say that numba itself doesn't do it (https://numba.pydata.org/numba-doc/dev/user/faq.html#does-numba-inline-functions), but that LLVM does in some cases.

jonmmease:
Also, the `append` function is constructed from evaluating a string, so perhaps that's a factor.

philippjfr:
>Also, the append function is constructed from evaluating a string, so perhaps that's a factor.

I wouldn't have thought so since once it's JIT compiled there should be no difference.

philippjfr:
As an aside, I can't get the implementation in this PR to work with an x_range or y_range provided to the Canvas.

jonmmease:
Hmm, should be tested by

https://github.com/pyviz/datashader/blob/711c40d8846d40bc8241f11a6977cd1b0f75ee81/datashader/tests/test_quadmesh.py#L67-L96

philippjfr:
Yeah, nevermind that's not the actual issue. The problem seems to be the order the dimensions are parsed in and returned as. I'd expect the DataArray dimensions to be ordered ``y, x`` on input and output, something appears to be inverting that.

philippjfr:
Don't really understand why this happens because this code is correct:

```
if x is None and y is None:
    y, x = source.dims
```

but in this example things are really weird:
 
```python
xs = np.logspace(0, 3, 50)
ys = np.arange(50)

zs = xs * ys[:, np.newaxis]
ds = hv.QuadMesh((xs, ys, zs), datatype=['xarray']).data
da = da.z

canvas = Canvas(x_range=(200, 1000), y_range=(0, 50))
​
print(da.dims)
​
agg1 = canvas.quadmesh(da, agg=ds.reductions.mean('z'))
​
agg2 = canvas.quadmesh(da, 'x', 'y', agg=ds.reductions.mean('z'))
​
print(agg1.dims, agg2.dims)
```

```
('y', 'x')
('x', 'y') ('y', 'x')
```

philippjfr:
To summarize the issues:

1. On the input the code inferring the dimension order appears wrong
2. On the output the coord order matches the dim ordering, which is usually not desirable. The coordinate order is meaningful by NetCDF convention and should in the 2D case usually either be the inverse of the dim ordering OR maybe more appropriately simply inherit the ordering of the input DataArray.

philippjfr:
Back to the inlining issue. I'm guessing the problem is that ``append`` is always two-levels deep, i.e. it will itself call another function to do the various base aggregations. I think we may be able to get it to inline properly if, instead of creating one append function, we pass in separate append function for each base aggregation, i.e. for a mean aggregation the inner loop would look something like:

```python
for xi in range(x0i, x1i):
    for yi in range(y0i, y1i):
        append_count(j, i, xi, yi, count_agg)
        append_sum(j, i, xi, yi, sum_agg)
```

The actual implementation would be something like:

```python
for xi in range(x0i, x1i):
    for yi in range(y0i, y1i):
        for append, array in zip(aggs, arrays):
            append(j, i, xi, yi, array)       
```

jonmmease:
The dimension flip seems to happen when converting the `DataArray` to a `Dataset`...
```python
print(da.dims)
print(list(da.to_dataset().dims))
```
```
['y', 'x']
['x', 'y']
```

So I'll just be sure to grab it from the `DataArray` itself.

jonmmease:
The dimension and coord ordering logic is being carried over from the default behavior of the glyph aggregators.  Here's what you get from `points`
```python
canvas = Canvas(x_range=(200, 1000), y_range=(0, 50))
canvas.points(pd.DataFrame({'x': [300], 'y': [20]}), 'x', 'y')
```
```
<xarray.DataArray (y: 600, x: 600)>
array([[0, 0, 0, ..., 0, 0, 0],
       [0, 0, 0, ..., 0, 0, 0],
       [0, 0, 0, ..., 0, 0, 0],
       ...,
       [0, 0, 0, ..., 0, 0, 0],
       [0, 0, 0, ..., 0, 0, 0],
       [0, 0, 0, ..., 0, 0, 0]], dtype=int32)
Coordinates:
  * y        (y) float64 0.04167 0.125 0.2083 0.2917 ... 49.71 49.79 49.88 49.96
  * x        (x) float64 200.7 202.0 203.3 204.7 ... 995.3 996.7 998.0 999.3
```

Do you think we should flip the coord order for what Datashader returns across all of the glyph types? Or we could treat quadmesh individually because it's the only one (so far) where the input data structure is an xarray.

philippjfr:
Actually the other issue with the current ``append`` is that it does the array lookup inside the inner loop which also seems pretty inefficient.

philippjfr:
>Or we could treat quadmesh individually because it's the only one (so far) where the input data structure is an xarray.

Yeah, I think we should have special logic for xarrays.

jonmmease:
> Actually the other issue with the current append is that it does the array lookup inside the inner loop which also seems pretty inefficient.

I'm going to manually inline the `mean` aggregate case so we can see what it looks like and what the performance difference is. I'll paste the inlined version here when I have it.

jbednar:
Isn't the raster input an xarray too?  But that's a different code path right now, I suppose.

jonmmease:
Ok, I've done a bunch of experiments and I think I finally have a handle on the performance differences.

## Primary performance factors

**Factor 1**:  There is a slight algorithmic difference between the "specialized" implementation and the implementation in this PR.  The specialized implementation will skip over quads that are less than one pixel wide or high, whereas the implementation in this PR will process these quads as if they take up the full pixel that they reside in.  This is what the trimesh renderer does as well and I think this is the behavior we want, but let's discuss if there's not agreement on that.

Here's the implementation of the mean aggregation that skips the subpixel quads. In the benchmarking below I call this implementation `spec_skip_parallel`.

```python
def rectilinear_mean(x, y, z, out_sum, out_count):
    for i in prange(len(x)-1):
        for j in range(len(y)-1):
            v = z[j, i]
            if not np.isfinite(v):
                continue

            x0, x1 = x[i], x[i+1]
            y0, y1 = y[j], y[j+1]
                
            for xi in range(x0, x1):
                for yi in range(y0, y1):
                    out_sum[yi, xi] += v
                    out_count[yi, xi] += 1
```

Here's a small modification that adds processing of subpixel quads. In the benchmarking below I call this implementation `spec_parallel`.

```python
def rectilinear_mean(x, y, z, out_sum, out_count):
    for i in prange(len(x)-1):
        for j in range(len(y)-1):
            v = z[j, i]
            if not np.isfinite(v):
                continue

            x0, x1 = x[i], x[i+1]
            y0, y1 = y[j], y[j+1]
            
            # Don't skip subpixel quads
            if x0 == x1:
                x1 += 1
                
            if y0 == y1:
                y1 += 1
                
            for xi in range(x0, x1):
                for yi in range(y0, y1):
                    out_sum[yi, xi] += v
                    out_count[yi, xi] += 1
```

**Factor 2**: It turns out that inlining wasn't the primary cause of the slowdown described above.  The primary cause seems to be that the presence of variadic arguments (i.e. *args in the signature) significantly impedes numba's optimization ability in this case.

The current implementation in this PR looks something this.
```python
def _extend(vt, bounds, xs, ys, *aggs_and_cols):
    for i in range(len(xs) - 1):
        ...
        for j in range(len(ys) - 1):
           ...
            for xi in range(x0i, x1i):
                for yi in range(y0i, y1i):
                     append(j, i, xi, yi, *aggs_and_cols)
```
In the benchmarking below, I call this implementation `append_args`

Numba is much more successful in optimizing this if the variadic args are replaced with a fixed number of arguments. The reason this is variadic in the first place is that different aggregation operations require a different number, but for benchmarking I hardcoded it to the number required for the `mean` aggregation operation.

```python
def _extend(vt, bounds, xs, ys, _0, _1, _2):
    for i in range(len(xs) - 1):
        ...
        for j in range(len(ys) - 1):
           ...
            for xi in range(x0i, x1i):
                for yi in range(y0i, y1i):
                     append(j, i, xi, yi, _0, _1, _2)
```
In the benchmarking below, I call this implementation `append_expanded`

**Factor 3**: No only does the presence of variadic args slow down numba's single-threaded performance, in this case it also prevents the use of parallelization.  At least, I couldn't figure out how to parallelize `append_args` without errors.  I *think* what was happeneing was that the use of  the `*aggs_and_cols` tuple inside the `prange` loops was being treated as a datastructure to be parallelized over, rather than as a constant that all loop iterations need access to.

In the benchmarking below, I call the implementation of `append_expanded` with parallelization `append_expanded_parallel`

```python
def _extend(vt, bounds, xs, ys, _0, _1, _2):
    for i in prange(len(xs) - 1):
        ...
        for j in range(len(ys) - 1):
           ...
            for xi in range(x0i, x1i):
                for yi in range(y0i, y1i):
                     append(j, i, xi, yi, _0, _1, _2)
```

## Benchmarking
For these benchmarks I used the same rectilinear quadmesh from the example notebook, but I varied the edge size of the quadmesh dataset from 500 to 2500 (before I only looked at 1000).

Here is a plot of the speedup factor (compared to the trimesh implementation) for the five algorithms discussed above.  Note the that the y-axis in is log scale.

bokeh_plot-1

### Observations
 1. For the size 2500 case, I see an almost 500x speedup from `trimesh` to `spec_skip_parallel`.  I think this is roughly the speedup that @philippjfr saw originally.

 2. If this implementation is updated to process subpixel quads, then the speedup drops to ~278x.

 3. For the PRs current implementation, `append_args`, there is a ~15x speedup in the 2500 case.

 4. Removing the variadic argument, `append_expanded`, increases the single-threaded speedup to ~50x. 

 5. Adding parallelization to this implementation, `append_expanded_parallel`, increases the speedup to ~285x. This is basically equivalent to the specialized implementation that processes subpixel quads.

## Where to go from here
At this point, I'm pretty satisfied that I understand the primary performance differences.  To remove the first disparity, we just need to decide how to handle subpixel quads.

The performance benefit of removing variadic args is significant, especially considering that this also makes parallelization possible.  But realizing these benefits across aggregations with variable argument length requirements will take some design work.  We don't want to have to write N versions of `_extend` for every glyph type. e.g.

```python
def _extend0(vt, bounds, xs, ys, _0):
def _extend1(vt, bounds, xs, ys, _0, _1):
def _extend2(vt, bounds, xs, ys, _0, _1, _2):
def _extend3(vt, bounds, xs, ys, _0, _1, _2, _3):
...
```

So this design is what I'm going to think on next...

philippjfr:
Wow, thanks for the detailed investigation. I agree we should process sub-pixel quads and the factor of ~<2 difference seems a reason tradeoff for that. We should probably talk to someone on the numba team about the variadic args issue, but assuming that can't be fixed easily we should definitely look into that. Any idea what kind of differences we can expect for other glyphs?

jbednar:
Nice analysis!  Ideally the subpixel rendering would be a user-level configuration option, defaulting to the safe version but able to be set by the user to the faster one for interactive use.  For variable arguments, what I've done before (e.g. in https://examples.pyviz.org/attractors/attractors_panel.html) is simply to accept the largest number of arguments that I might need, i.e. in this case:

```
def _extend3(vt, bounds, xs, ys, _0, _1, _2, _3):
```

and ignore them when not needed.  I don't know if that will work in this situation.

jonmmease:
I think I've worked out an approach to automating this variable-arity optimization that works across glyphs.  I'll open a separate PR with that when it's ready...

jonmmease:
For now, I updated the construction of the returned `DataArray` to specify a coordinate order of [`x`, `y`].  This ordering was the reason that passing an aggregate result into a HoloViews `Image` resulted in an inversion of the x and y axes (which was true for points and lines as well).  Done in commit https://github.com/pyviz/datashader/pull/779/commits/87cf7e3f333744ea21690816da497bec1ab93da7.

jonmmease:
@philippjfr I think I'm satisfied (in terms of correctness, API, and performance) with the rectilinear quadmesh implementation in this PR now.  So feel free to kick the tires again and let me know if anything else sticks out.

I'm going to move on to the curvilinear case now.

jbednar:
So, what's the final word on the performance?

jonmmease:
> So, what's the final word on the performance?

Equivalent to the "append_expanded" line above (~40x speedup over trimesh for rectilinear case).  Turns out that all of the `*_parallel` implementations suffer from a race condition that would require something like https://github.com/numba/numba/issues/3681 to address.  So `prange` parallelization is not enabled at this point.

jonmmease:
@jbednar @philippjfr I've added the curvilinear quadmesh support and updated the PR description to reflect the current contents of the PR.  Ready for review!

datashader/glyphs/quadmesh.py
```diff
@@ -120,16 +120,132 @@ def extend(aggs, xr_ds, vt, bounds):
...
+            yscaled = (y_mapper(y_breaks) - y0) / yspan
+
+            plot_height, plot_width = aggs[0].shape[:2]
+
+            xs = (xscaled * plot_width).astype(int)
+            ys = (yscaled * plot_height).astype(int)
+
+            # Question: Should we try to compute a slice of xs and ys that
+            # eliminates rows and columns of quads that are all outside the
+            # viewport?
```
philippjfr: I tried this but for many real world meshes this wasn't all that effective to reduce the array size. May still be worth it though.

PR #780, opened 2019-08-15
Glyph rendering optimization using variable length argument expansion

jonmmease:
## Background
This PR is motivated by the performance characteristics discovered while developing the quadmesh glyph in https://github.com/pyviz/datashader/pull/779. In particular, see the discussion of fixed vs variable length arguments in https://github.com/pyviz/datashader/pull/779#issuecomment-521248206.

The key insight here is that during glyph rendering, the numba optimization of the rendering functions can result in substantially faster code if the input functions contain a fixed number of arguments rather than a variable length argument (e.g. *args).

In the glyph rendering code, this variable length argument is usually called `*aggs_and_cols` and it contains a list of the aggregate arrays that are being populated and the columns that are used as input to the reduction calculations.  The length of this argument varies depending on the chosen reduction operation, but the length will remain the same for every render call for a given operation.

### expand_varargs
This PR adds a `datashader.macros` module that provides the `expand_varargs` decorator builder.  This decorator builder inputs the desired number of arguments to expand to, and returns a function decorator.  This function decorator will transform the [AST](https://docs.python.org/3/library/ast.html) of the wrapped function to replace variable length arguments with a fixed number of arguments/variables.

This only makes sense for cases where the only thing the function does with the variable length argument is to pass it along to other functions in star-form.  For example, calling...

```python
@expand_varargs(2)
def example_fn(a, b, *args):
    print(a, b)
    other_fn(a, b, *args)
```

would transform the function AST into a function equivalent to 
```python
def example_fn(a, b, _0, _1):
    print(a, b)
    other_fn(a, b, _0, _1)
```

If the variable length argument is used in any other context then an error is raised. For example, an error will be raised if the `example_fn` looks inside `args`.

```python
@expand_varargs(2)
def example_fn(a, b, *args):
    print(a, b, args[0])
    other_fn(a, b, *args)
```
:arrow_up: would raise a `ValueError`.

The whole point of doing this is to transform the input function before it is passed to numba's jit compilation decorator.

### Glyph.expand_aggs_and_cols
A new `expand_aggs_and_cols` method has been added to the `Glyph` baseclass. This method inputs the `append` function that will perform the reduction operation, and returns an `expand_varargs` decorator configured to expand the `*aggs_and_cols` argument to the correct number of fixed arguments.

### Glyph updates
The new decorator has been added to all applicable rendering functions in the `points`, `line`, and `area` glyphs.  I think Trimesh could also benefit from this technique, but it doesn't quite follow the same pattern of passing around the `*aggs_and_cols` argument, so that will take some refactoring.  `raster` doesn't currently use the same glyph/aggregation framework.

## Performance results
Here are some benchmark comparisons of this branch with `master`.

Notebooks:
 - **Before**: https://anaconda.org/jonmmease/expand_varargs_before/notebook
 - **After**: https://anaconda.org/jonmmease/expand_varargs_after/notebook

I tested `points`, `line`, and `area` glyphs using the `count`, `sum`, `mean`, and `std` reduction operations.  Here is a plot of the results: 

newplot

Note that the y-axis is time in seconds.

Notice how the improvements grow more significant as the reduction operation get more complex. This corresponds to the `*aggs_and_cols` argument getting longer.

The speedups for the coming quadmesh glyph should be even more significant than the gains for the area glyph above.

@jbednar @philippjfr

### DIFF
--- a/datashader/geo.py
+++ b/datashader/geo.py
@@ -3,7 +3,7 @@ This module contains geoscience-related transfer functions whose use is complete
 
 """
 
-from __future__ import division
+from __future__ import division, absolute_import
 
 import numpy as np
 import datashader.transfer_functions as tf
--- a/datashader/glyphs/area.py
+++ b/datashader/glyphs/area.py
@@ -89,10 +89,14 @@ class AreaToZeroAxis0(_PointLike):
 
     @memoize
     def _build_extend(self, x_mapper, y_mapper, info, append):
-        draw_trapezoid_y = _build_draw_trapezoid_y(append)
+        expand_aggs_and_cols = self.expand_aggs_and_cols(append)
+        draw_trapezoid_y = _build_draw_trapezoid_y(
+            append, expand_aggs_and_cols
+        )
         map_onto_pixel = _build_map_onto_pixel_for_line(x_mapper, y_mapper)
         extend_area = _build_extend_area_to_zero_axis0(
-            draw_trapezoid_y, map_onto_pixel)
+            draw_trapezoid_y, map_onto_pixel, expand_aggs_and_cols
+        )
         x_name = self.x
         y_name = self.y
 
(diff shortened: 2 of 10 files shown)


jonmmease:
Update: now the `Glyph.expand_aggs_and_cols` method will return an identity decorator if the `NUMBA_DISABLE_JIT` environnmet variable is set.

This environment variable is used by numba to disable JIT compilation.  With this change, functions that are wrapped with both `@jit` and `@expand_varargs` will remain unchanged when `NUMBA_DISABLE_JIT` is set.  This makes it possible to debug these functions with a graphical debugger.

jonmmease:
@philippjfr this is ready to go apart from that one Travis test flaking out again.  Could you restart it?

jonmmease:
all green. merging

PR #791, opened 2019-09-27
Avoid Pandas and Xarray warnings

jbednar:
Fixes https://github.com/pyviz/datashader/issues/789

### DIFF
--- a/datashader/bundling.py
+++ b/datashader/bundling.py
@@ -322,7 +322,7 @@ def _convert_graph_to_edge_segments(nodes, edges, params):
     df = df.filter(items=segment_class.get_merged_columns(params))
 
     edge_segments = []
-    for edge in df.get_values():
+    for edge in df.values:
         edge_segments.append(segment_class.create_segment(edge))
     return edge_segments, segment_class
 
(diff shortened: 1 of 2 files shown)


jbednar:
However, I still have a problem with current master and xarray 0.11.3 or 0.13 and pandas 0.25.1, which is that the http://datashader.org/user_guide/7_Networks.html#Graphs-with-categories section of the user guide is failing with the error:

```
TypeError: ufunc 'over' not supported for the input types, and the inputs could not 
be safely coerced to any supported types according to the casting rule ''safe''
```

Presumably that code used to work.  It's complaining about overlaying the node and edge plots in this case, even though overlaying those same types of plots in other cells above and below works fine using the same code. Plus if I hack `graphplot` to overlay just the nodes on themselves or just the edges on themselves, the overlay works, suggesting that the edges and nodes in this case end up as incompatible types of object that can't be overlaid with `over`.  Yet the two plots appear to have the same type (<class 'datashader.transfer_functions.Image'>) and dtype (uint32). I'll try to see when this problem started, but it has the same behavior on xarray 0.11.3 and 0.13, so if it's due to a change in xarray it would presumably be from an older version.

jbednar:
I had the same `over` problem with xarray 0.9.6 and pandas 0.23, so I'll merge this for now but still need to figure out the problem here.

Issue #181, opened 2016-06-22
Add polygon support


brendancol:
Add support for Canvas.polygon.  Something like:

``` python
endangered_agg = cvs.polygon(df, 'my_polygon_field', how='sum', field='endangered_species_list_score')
```

jonmmease:
I think I have a pretty clear understanding of what this might look like now that I've finished the following PRs:
 - `RaggedArray` pandas extensions array (https://github.com/pyviz/datashader/pull/687)
 - Support for row-oriented coordinates in `Canvas.line` (https://github.com/pyviz/datashader/pull/694)
 - Add quadmesh glyph with rectilinear and curvilinear support (https://github.com/pyviz/datashader/pull/779)

Here's what I'm thinking...

### Polygon representation
I think the easiest thing to do here is to represent a polygon set as a pair of arrays, one for each the x and y coordinate. `nan` values are then used to indicate the end of one polygon and the start of the next.

### Polygon fill rendering
In the quadmesh PR (https://github.com/pyviz/datashader/pull/779) I implemented filling using the "crossing number" point-in-polygon test (http://geomalgorithms.com/a03-_inclusion.html).  If we use this same algorithm for these sets of polygons then we naturally get the following behavior.
  - The first polygon is filled in if it's the only one
  - The second (and third, forth, ...) polygon acts as a hole if it is enclosed in the first polygon, or it is filled in as a separate polygon if it is outside of the first polygon.

### Polygon API
I propose that the `Canvas.polygon` method have the same API as `Canvas.line`. We don't need to cover the full API right away, but in principal every combination of the `axis` and column arguments supported by `Canvas.line` could also be supported by `Canvas.poly`. Here are the two API uses that I expect to be most common

Variable sized polygons, one per row.  In this case, the `x` and `y` coordinates would be stored as `RaggedArray` columns in the DataFrame.

```python
canvas.polygon(df, 'x', 'y', axis=1)
```

A collection of fixed-size polygons.  This would be useful for displaying a large collection of polygons that all have the same number of vertices.  In this case, the data frame would have a column for each x and y coordinate.  For example, rendering a collection of pentagons would look something like this

```python
canvas.polygon(df, ['x0', 'x1', 'x2', 'x3', 'x4'], ['y0', 'y1', 'y2', 'y3', 'y4'], axis=1)
```

### Polygon outlines
[...]

PR #793, opened 2019-10-01
Initial GPU support

jonmmease:
## Overview
This PR adds initial GPU support to Datashader :tada:. This is implemented using a combination of the `cudf` and `cupy` libraries and numba's cuda support.

This work was inspired by the cuDataShader project: https://github.com/rapidsai/cuDataShader.

cc: @jbednar, @philippjfr, @exactlyallan

## Supported Features
The following Datashader features can now be accelerated by an NVIDIA GPU supported by recent versions of `cudf`/`cupy`/`numba`.

 - `Canvas.points` rasterization
 - `Canvas.line` and `Canvas.area` rasterization
 - All reduction operations *except* `var` and `std`. The current algorithm for these is a single pass serial algorithm that doesn't extend to fine-grained parallelization. For GPU parallelization, I think we would want to use a two-pass algorithm (compute the mean during the first pass, then compute the sum of squared difference from the mean in the second pass), but this would require a bit more refactoring to support.
 - `transfer_functions.shade` (both 2D and 3D) inputs
 
 For the `points`/`line`/`area` methods, GPU acceleration is enabled automatically when the input data frame is a `cudf.DataFrame` instance.  In this case, the aggregation results are returned in an xarray `DataArray` that is backed by a `cupy.ndarray` instance (rather than a `numpy.ndarray` instance). This way the aggregation results remain in GPU memory.
 
 The `transfer_functions.shade` function will be GPU accelerated if it is passed an xarray `DataArray` that is backed by a `cupy.ndarray` instance.
 
## Performance
I created the following benchmark notebooks:
 - Benchmark `Canvas.points`: https://anaconda.org/jonmmease/gpu_datashader_points_pr/notebook
 - Benchmark `Canvas.line`: https://anaconda.org/jonmmease/gpu_datashader_lines_pr/notebook
 
For each of these notebooks, I compared the performance of passing a pandas DataFrame (single threaded CPU), a dask DataFrame with 12 partitions (12 threaded CPU on a 14 core workstation), and a cudf DataFrame (GeForce RTX 2080).

### Points with `count` aggregate
Rendering ~100 million points

 - pandas: 1280 ms
 - dask: 284 ms (~4.5x speedup from pandas)
 - cudf: 31.2 ms (~40x speedup from pandas)
 
### Points with `count_cat` aggregate
Rendering ~100 million points

 - pandas: 1690 ms
 - dask: 681 ms (~2.5x speedup from pandas)
 - cudf: 83.6 ms (~20x speedup from pandas)
 
### Line
Rendering 1 million length-10 lines

 - pandas: 2060 ms
 - dask: 317 ms (~6.5x speedup from pandas)
 - cudf: 69.3 ms (~30x speedup from pandas)

## Testing
The test suite is set up to run the GPU tests only if `cupy` and `cudf` are installed. We should talk about how we want to handle CI testing of the GPU code going forward.

### DIFF
--- a/datashader/compiler.py
+++ b/datashader/compiler.py
@@ -15,7 +15,7 @@ __all__ = ['compile_components']
 
 
 @memoize
-def compile_components(agg, schema, glyph):
+def compile_components(agg, schema, glyph, cuda=False):
     """Given a ``Aggregation`` object and a schema, return 5 sub-functions.
 
     Parameters
@@ -51,20 +51,20 @@ def compile_components(agg, schema, glyph):
     reds = list(traverse_aggregation(agg))
 
     # List of base reductions (actually computed)
-    bases = list(unique(concat(r._build_bases() for r in reds)))
+    bases = list(unique(concat(r._build_bases(cuda) for r in reds)))
     dshapes = [b.out_dshape(schema) for b in bases]
     # List of tuples of (append, base, input columns, temps)
-    calls = [_get_call_tuples(b, d, schema) for (b, d) in zip(bases, dshapes)]
+    calls = [_get_call_tuples(b, d, schema, cuda) for (b, d) in zip(bases, dshapes)]
     # List of unique column names needed
     cols = list(unique(concat(pluck(2, calls))))
     # List of temps needed
     temps = list(pluck(3, calls))
 
-    create = make_create(bases, dshapes)
+    create = make_create(bases, dshapes, cuda)
     info = make_info(cols)
     append = make_append(bases, cols, calls, glyph)
     combine = make_combine(bases, dshapes, temps)
-    finalize = make_finalize(bases, agg, schema)
+    finalize = make_finalize(bases, agg, schema, cuda)
 
     return create, info, append, combine, finalize
 
@@ -79,13 +79,18 @@ def traverse_aggregation(agg):
         yield agg
 
 
-def _get_call_tuples(base, dshape, schema):
-    return base._build_append(dshape, schema), (base,), base.inputs, base._build_temps()
+def _get_call_tuples(base, dshape, schema, cuda):
+    return (base._build_append(dshape, schema, cuda),
+            (base,), base.inputs, base._build_temps(cuda))
 
 
-def make_create(bases, dshapes):
+def make_create(bases, dshapes, cuda):
     creators = [b._build_create(d) for (b, d) in zip(bases, dshapes)]
-    array_module = np
+    if cuda:
+        import cupy
+        array_module = cupy
+    else:
+        array_module = np
     return lambda shape: tuple(c(shape, array_module) for c in creators)
 
 
@@ -145,22 +150,22 @@ def make_combine(bases, dshapes, temps):
     return combine
 
 
-def make_finalize(bases, agg, schema):
+def make_finalize(bases, agg, schema, cuda):
     arg_lk = dict((k, v) for (v, k) in enumerate(bases))
     if isinstance(agg, summary):
         calls = []
         for key, val in zip(agg.keys, agg.values):
-            f = make_finalize(bases, val, schema)
+            f = make_finalize(bases, val, schema, cuda)
             try:
                 # Override bases if possible
-                bases = val._build_bases()
+                bases = val._build_bases(cuda)
             except AttributeError:
                 pass
             inds = [arg_lk[b] for b in bases]
             calls.append((key, f, inds))
 
-        def finalize(bases, **kwargs):
-            data = {key: finalizer(get(inds, bases), **kwargs)
+        def finalize(bases, cuda=False, **kwargs):
+            data = {key: finalizer(get(inds, bases), cuda, **kwargs)
                     for (key, finalizer, inds) in calls}
             return xr.Dataset(data)
         return finalize
--- a/datashader/core.py
+++ b/datashader/core.py
@@ -19,6 +19,15 @@ from .utils import Expr # noqa (API import)
 from .resampling import resample_2d, resample_2d_distributed
 from . import reductions as rd
 
+try:
+    import cudf
+except ImportError:
+    cudf = None
+
+try:
+    import dask_cudf
+except ImportError:
+    dask_cudf = None
 
 class Axis(object):
     """Interface for implementing axis transformations.
@@ -107,7 +116,7 @@ class LogAxis(Axis):
     @staticmethod
     @ngjit
     def mapper(val):
-        return log10(val)
+        return log10(float(val))
 
     @staticmethod
     @ngjit
@@ -993,7 +1002,8 @@ def bypixel(source, canvas, glyph, agg):
         source = source.drop([col for col in columns if col not in cols_to_keep])
         source = source.to_dask_dataframe()
 
-    if isinstance(source, pd.DataFrame):
+    if (isinstance(source, pd.DataFrame) or
+            (cudf and isinstance(source, cudf.DataFrame))):
         # Avoid datashape.Categorical instantiation bottleneck
         # by only retaining the necessary columns:
         # https://github.com/bokeh/datashader/issues/396
--- a/datashader/data_libraries/__init__.py
+++ b/datashader/data_libraries/__init__.py
@@ -5,3 +5,14 @@ try:
     from . import dask    # noqa (API import)
 except ImportError:
     pass
+
+try:
+    import cudf as _cudf  # noqa (Test cudf installed)
+    import cupy as _cupy  # noqa (Test cupy installed)
+    from . import cudf    # noqa (API import)
+
+    import dask_cudf as _dask_cudf  # noqa (Test dask_cudf installed)
+    from . import dask_cudf         # noqa (API import)
+
+except ImportError:
+    pass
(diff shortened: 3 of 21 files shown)


jbednar:
Fabulous! How does the performance compare to  https://github.com/rapidsai/cuDataShader ?

jbednar:
For var and std, can we supply separate two-pass implementations and select those when needed for cu-backed operations?

exactlyallan:
Great work! Any plans to add other GPU accelerated features? 
@AjayThrove @wxbn

jonmmease:
> Fabulous! How does the performance compare 

Good question. For the 100 million point test (with count reduction) this implementation is a bit faster.
 - pandas: 1280 ms
 - cudf (this PR): 31.2 ms (~40x speedup from pandas)
 - cuDataShader: 96 ms (13x speedup from pandas)

For 10 million, they are about the same:
 - pandas 176 ms
 - cudf (this PR): 13.6 ms (~13x speedup from pandas)
 - cuDataShader: 13.2 ms (~13x speedup from pandas)

> For var and std, can we supply separate two-pass implementations and select those when needed for cu-backed operations?

Yes, I think we could. It's just that this paradigm can't be represented by the current reduction pipeline code, so it would take a bit of refactoring.

jonmmease:
Thanks @exactlyallan!

> Any plans to add other GPU accelerated features?

What features do you have in mind?

jbednar:
That's great that the performance is the same or better than cuDatashader; the code changes involved generally seem quite reasonable and something we can maintain. I'm greatly relieved to find that overall things fit quite well into the current architecture, without sacrificing performance!

exactlyallan:
We never got to GPU-ing trimesh, which we think we could work as a means to choropleth maps. 

Also, I'm curious if GPU edge bundling makes sense too, like a version for hammer_bundle or something like [cuDataShader edge Bundling notebook](https://github.com/rapidsai/cuDataShader/blob/master/cudatashader-notebooks/cuDatashader%20Edge%20Bundling.ipynb)

jonmmease:
@jbednar, thanks for taking a look.  Yeah, I can definitely split this up more.  As you suggest, I'll create a first PR with as much of the refactoring as possible and a more atomic commit log.

philippjfr:
>We never got to GPU-ing trimesh, which we think we could work as a means to choropleth maps.

I think we were planning on adding a general polygon implementation, which should be a lot more efficient than triangulating the polygon ahead of time. I'd really love to see both being GPU accelerated though.

jbednar:
Thanks! You can make a more atomic commit log if you want, and it certainly doesn't hurt, but for my purposes a single commit for all the non-cuda work would be fine, as long as there is a checklist in the PR text that explains what the non-trivial changes are.  If you want to do that by commit, sure; whatever's easier for you to think about!

jbednar:
Right; I think a full polygon implementation will be much more efficient than a trimesh-based implementation, and we are already working on parts of that.

jonmmease:
> I think we were planning on adding a general polygon implementation, which should be a lot more efficient than triangulating the polygon ahead of time. I'd really love to see both being GPU accelerated though.

One wrinkle here is that cudf doesn't directly support pandas extension array types, so the `RaggedArray` type that is used in Datashader right now for variable length lines and that I proposed using for polygons in https://github.com/pyviz/datashader/issues/181#issuecomment-523590110 won't work out of the box right now.

exactlyallan:
> Right; I think a full polygon implementation will be much more efficient than a trimesh-based implementation, and we are already working on parts of that.

Fantastic, rough ETA on that?

jbednar:
I'm just now trying to coordinate that task between 5 or 6 people, so I don't know yet, but should soon know!

jorisvandenbossche:
> One wrinkle here is that cudf doesn't directly support pandas extension array types, so the RaggedArray type that is used in Datashader right now for variable length lines and that I proposed using for polygons in #181 (comment) won't work out of the box right now.

@jonmmease cudf is based on Arrow memory, right? (but I don't know the details how you interact with it on the python level) And I would also think that the RaggedArray you have for pandas should more or less directly map to Arrow's ListArray. Could that be a way to support this on the GPU?

jonmmease:
@jorisvandenbossche, yes a `ListArray` sounds like the right DataStructure. I haven't looked into the internals of cudf, but this does sound like a good direction. Thanks!

jonmmease:
Good news, it was pretty easy to add `dask_cudf` support on top of what I already had here.  So multi-GPU support should also now work!  I say *should* because I've only tested the `dask_cudf` support on a single GPU, in which case there is predictably no performance benefit.  I have a second GPU on the way so I'll soon have a chance to test it out with two GPUs on the same machine.

jonmmease:
I've confirmed that this PR operates correctly with a `dask_cudf` DataFrame distributed across multiple GPUs.  Unfortunately, I haven't found a case yet where there's a performance benefit for doing so (compared to using a normal CPU dask DataFrame).

jonmmease:
I've updated this PR based on master and added support for `line`/`area` with `axis=0` (in addition to `axis=1`).

@philippjfr, when you have a chance. Could you try this out on your NVIDIA hardware?  It would also be nice to see if your https://github.com/pyviz/holoviews/pull/3982 PR works alright with these changes.

pzwang:
I dunno why, but I somehow missed this huge news.  Distributed GPU DataShader has been a dream of mine since the beginning... this is outstanding!

@jonmmease For the distributed GPU dataframe to show wins, maybe you just need a larger dataset... :-)

jbednar:
@pzwang , you were probably just distracted a little bit by becoming CEO or some such. :-)  Yes, surely there would be a benefit for a large enough dataset, but we'll see...

jonmmease:
Thanks for the kind words @pzwang!  I'm very excited about this as well.

When I push my 2 RTX 2080s to the limit, I can get almost 200 million rows of a 3 column DataFrame persisted.  And this does indeed render in about 1/4 the time of a two worker CPU Dask LocalCluster (~1000 ms vs. ~260 ms).

Even so, it looks like there's larger constant overhead for the `LocalCudaCluster` compared to the normal `LocalCluster`.  If I render only the first 1000 points of this DataFrame with the `LocalCluster` time is ~100ms, but with the `LocalCudaCluster` it's ~220ms.  So perhaps there's some optimization that could be done in `dask-cudf` to lower this overhead to be more in line with `dask.dataframe` when used with the distributed scheduler.

kkraus14:
Hey all, cuDF maintainer here, I'd just like to mirror @pzwang's comment that this is extremely exciting!

> When I push my 2 RTX 2080s to the limit, I can get almost 200 million rows of a 3 column DataFrame persisted. And this does indeed render in about 1/4 the time of a two worker CPU Dask LocalCluster (~1000 ms vs. ~260 ms).
>
> Even so, it looks like there's larger constant overhead for the LocalCudaCluster compared to the normal LocalCluster. If I render only the first 1000 points of this DataFrame with the LocalCluster time is ~100ms, but with the LocalCudaCluster it's ~220ms. So perhaps there's some optimization that could be done in dask-cudf to lower this overhead to be more in line with dask.dataframe when used with the distributed scheduler.

Is there any way you can run the computation twice to see if the second run is much faster? You may be hitting an operation that we currently JIT compile using Numba and the JIT compilation overhead is typically ~200ms. If it's not JIT compilation then would it be possible to dump a cProfile that I can get some eyes on to see where we're currently spending our time? There's some big optimizations coming to cuDF with respect to avoiding a lot of slow Python control flow that I'm curious if you're hitting.

jonmmease:
Hi @kkraus14, thanks for chiming in!  The profiling I'm doing is running everything multiple times to give numba a chance to JIT everything since Datashader itself already uses numba extensively.

And to be clear, I'm seeing really impressive/exciting speedup when operating directly on a single `cudf.DataFrame` (See benchmarks in the top issue post). The place where I'm seeing some overhead that looks a bit larger than I would expect is only when operating on `dask_cudf.DataFrame` instances spread across multiple GPUs using a `LocalCUDACluster`.  

We don't need to fully investigate this area before merging this PR, but I'll see if I can create a MWE that doesn't depend on Datashader that shows what I'm talking about.

jbednar:
I'll go ahead and merge this, as we don't have an imminent release; hopefully other people will be able to test it soon!

jbednar:
>> Right; I think a full polygon implementation will be much more efficient than a trimesh-based implementation, and we are already working on parts of that.
> Fantastic, rough ETA on that?

The polygon implementation is ready! It's in https://github.com/holoviz/datashader/pull/826 and should be merged very soon now that the GPU support has been released.

jbednar:
For anyone wondering, I think this is the current status of support for the various data backends in Datashader for drawing the various glyph types available:

image

I'll try to find somewhere to put that in the docs.

Supporting GPU raster, trimesh, and quadmesh types should already be feasible, but those glyphs each use different rendering code that has to be implemented for GPUs separately, so it will take us a good bit of effort.  #826 has now been merged, adding choropleth (polygon) and outline (multiline) rendering to Datashader, but further extending that to work with the GPU has to wait on support for Pandas ExtensionArrays in cuDF.  

@kkraus14 and @exactlyallan, I don't know if NVIDIA is planning to support ExtensionArrays in cuDF, but it would open up some very cool applications! E.g. we've currently got a million-polygon choropleth dataset that takes 49 minutes to render one frame on the CPU with Datashader, which would surely get a big boost from the GPU.

We'd also love any help you can give with the LocalCudaCluster overhead issue, which I guess is waiting on an MRE from @jonmmease.

Issue #792, opened 2019-09-27
ufunc 'over' not supported for the input types in 7_Networks.ipynb


jbednar:
With current master and xarray 0.11.3 or 0.13 and pandas 0.25.1, the http://datashader.org/user_guide/7_Networks.html#Graphs-with-categories section of the user guide is failing with the error:

```
TypeError: ufunc 'over' not supported for the input types, and the inputs could not 
be safely coerced to any supported types according to the casting rule ''safe''
```

Presumably that code used to work, since the page looks fine on the website. Python's complaining about overlaying the node and edge plots in this case, even though overlaying those same types of plots in other cells above and below works fine using the same code. Plus if I hack graphplot to overlay just the nodes on themselves or just the edges on themselves, the overlay works, suggesting that the edges and nodes in this case are individually fine but end up as incompatible types of object that can't be overlaid with over. Yet the two plots appear to have the same type (<class 'datashader.transfer_functions.Image'>) and dtype (uint32). 

I'll try to see when this problem started, but it has the same behavior on xarray 0.11.3 and 0.13, so if it's due to a change in xarray it would presumably be from an older version.

jbednar:
I had the same over problem with xarray 0.9.6 and pandas 0.23, which are quite old by now, so it's a mystery.

jbednar:
Reproducer:

```
import math
import numpy as np
import pandas as pd

import datashader as ds
import datashader.transfer_functions as tf
from datashader.layout import random_layout, circular_layout, forceatlas2_layout
from datashader.bundling import connect_edges, hammer_bundle

from itertools import chain

np.random.seed(1)
cats,n,m = 4,10,20

cnodes = pd.concat([
           pd.DataFrame.from_records([("node"+str(i+100*c),"c"+str(c)) for i in range(n)], 
                        columns=['name','cat']) 
             for c in range(cats)], ignore_index=True)
cnodes.cat=cnodes.cat.astype('category')

cedges = pd.concat([
           pd.DataFrame(np.random.randint(n*c,n*(c+1), size=(m, 2)), 
                        columns=['source', 'target'])
         for c in range(cats)], ignore_index=True)

cvsopts = dict(plot_height=400, plot_width=400)

def nodesplot(nodes, name=None, canvas=None, cat=None):
    canvas = ds.Canvas(**cvsopts) if canvas is None else canvas
    aggregator=None if cat is None else ds.count_cat(cat)
    agg=canvas.points(nodes,'x','y',aggregator)
[...]

PR #795, opened 2019-10-03
Fix coordinate ordering of Image returned by _colorize

jonmmease:
Fix for #792.  Could you confirm that this fixes the issue for you @jbednar?


jbednar:
Yes, it does, thanks so much!

PR #796, opened 2019-10-03
Fixed edge bundling performance regression

jbednar:
There was a serious performance regression in edge bundling caused by https://github.com/pyviz/datashader/pull/763.  On the left is what's on the website from 0.7.0,  on the right is a new build on my recent MacBook Pro.  49 seconds became 582 seconds (9:42):

image

On investigation, it turns out that that PR was effectively disabling Numba acceleration for a key step in bundling, which did prevent a bunch of warnings from being shown, but resulted in unacceptably slow performance.  The Numba team helped me figure out that the real problem was that the dtype was changing between a default value (used when the loop was not executed) and the typical value (used when the loop was), which made Numba complain a lot.  After fixing that, the performance is now faster than on the website, with no warnings:

image

### DIFF
--- a/datashader/bundling.py
+++ b/datashader/bundling.py
@@ -33,9 +33,9 @@ def distance_between(a, b):
     return (((a[0] - b[0]) ** 2) + ((a[1] - b[1]) ** 2))**(0.5)
 
 
-@nb.jit(forceobj=True)
+@nb.jit
 def resample_segment(segments, new_segments, min_segment_length, max_segment_length, ndims):
-    next_point = np.zeros(ndims)
+    next_point = np.zeros(ndims, dtype=segments.dtype)
     current_point = segments[0]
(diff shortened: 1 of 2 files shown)


jbednar:
@philippjfr , it looks like similar issues also apply to https://github.com/pyviz/datashader/pull/763/files#diff-f5958d030b3961fac0730f0e72760f43L174 ?

jbednar:
For  https://github.com/pyviz/datashader/pull/763/files#diff-f5958d030b3961fac0730f0e72760f43L174, the warnings were indicating that Numba doesn't currently support that function ("cooling") at all, because of its use of a sparse matrix.  Support for speeding up cooling will have to wait on Numba sparse matrix support someday...

philippjfr:
Looks good, will merge.

PR #798, opened 2019-10-04
Optimize dshape_from_dask

### LINKED ISSUE #653: Handle categorical columns correctly on dask dataframes
In https://github.com/pyviz/datashader/pull/634 we discovered that when loading data using dask, categorical columns are not necessarily handled correctly. Specifically the code currently calls ``.head()`` to load the categories into memory, but this only works correctly for parquet files written with fastparquet, for other data sources the categories can only reliably be computed by loading the whole column into memory.

The appropriate solution is to ensure that ``categorize`` or ``as_known`` are called on the categorical columns before aggregation. However this can cause significant slowdowns in many scenarios so we have to carefully consider how best to go about implementing this, i.e. consider some caching strategy or strongly suggest that users run this beforehand to ensure that the categories do no have to be inferred on every aggregation.

### LINKED ISSUE #797: Memoize or otherwise optimize dshape_from_dask
Currently whenever you do an aggregation datashader has to figure out the columns and types of the data by fetching the .head of the dask dataframe. When the data is persisted locally this is not a big deal but on a remote Cluster this can introduce serious overhead due to latency. 

As an example, currently I'm playing with a remote dask gateway cluster and aggregating about 7 million rows. This takes about 1 second, but about 300 ms of that time is spent fetching the df.head() to determine its datashape. I'm not quite sure how to address this yet since dask dataframes can be modified inplace and a simple memoization scheme may not be sufficient.

philippjfr:
- Fixes #797 
- Closes #653

### DIFF
--- a/datashader/utils.py
+++ b/datashader/utils.py
@@ -361,6 +361,8 @@ def dshape_from_pandas_helper(col):
     dataframe.
     """
     if isinstance(col.dtype, type(pd.Categorical.dtype)) or isinstance(col.dtype, pd.api.types.CategoricalDtype):
+        if not getattr(col.cat, 'known', True):
+            col = col.cat.as_known()
         cat_dshape = datashape.dshape('{} * {}'.format(
             len(col.cat.categories),
             col.cat.categories.dtype,
--- a/datashader/utils.py
+++ b/datashader/utils.py
@@ -391,7 +391,8 @@ def dshape_from_pandas(df):
 
 def dshape_from_dask(df):
     """Return a datashape.DataShape object given a dask dataframe."""
-    return datashape.var * dshape_from_pandas(df.head()).measure
+    return datashape.var * datashape.Record([(k, dshape_from_pandas_helper(df[k]))
+                                             for k in df.columns])
 
 
 def dshape_from_xarray_dataset(xr_ds):
--- a/datashader/utils.py
+++ b/datashader/utils.py
@@ -393,6 +393,12 @@ def dshape_from_pandas(df):
 
 def dshape_from_dask(df):
     """Return a datashape.DataShape object given a dask dataframe."""
+    cat_columns = [col for col in df.columns if isinstance(df[col].dtype, type(pd.Categorical.dtype))
+                   or isinstance(df[col].dtype, pd.api.types.CategoricalDtype)]
+    if len(cat_columns) > 1:
+        # If there is more than one categorical column it is faster
+        # to compute the df.head() which will contain all categories
+        return datashape.var * dshape_from_pandas(df.head()).measure
     return datashape.var * datashape.Record([(k, dshape_from_pandas_helper(df[k]))
                                              for k in df.columns])
 


jonmmease:
Yeah, I think this is a good approach. I expect it will, but can you double check that it works for categorical columns?

philippjfr:
Will do, although I don't think we handle those correctly right now either, see https://github.com/pyviz/datashader/issues/653

philippjfr:
Okay, I've pushed a fix to handle the case where the categories aren't known. Just wondering whether maybe we should modify the columns on the dataframe inplace so that the categories only need to be inferred once.

philippjfr:
It seems like the original guidance I got from dask folks wasn't quite correct, calling .head() does seem to load all the categories and not just the categories that are actually present in the head. So #653 was not actually an issue as far as I can tell. Explicitly calling `as_known` is therefore only needed for the new approach. Calling that is faster than calling .head() if there is 1-2 categorical columns but ends up being slower when there are more. It is however much faster if there are no categorical columns. So maybe to get the best of both worlds we should use the old approach for more than one categorical columns and use the new approach for zero to one categorical columns. Does that sound reasonable?

jonmmease:
The 1-2 seems a bit arbitrary and I would think it wouldn't always be the optimal cutoff.

If you call `head` on the subset of columns that are categorical do you get comparable performance for the 1-2 category case? e.g. `df[[cat_cols]].head()`.  If so, then I think I'd prefer the approach of always calling `head` on this subset (but don't call it at all if there are no categoricals), and then handle the rest of the columns as in your original PR here.

philippjfr:
>If you call head on the subset of columns that are categorical do you get comparable performance for the 1-2 category case?

No, the overhead of .head doesn't seem to reduce much when selecting a subset of columns (in fact it increases in some cases). So this approach would be much worse than even the current performance.

jonmmease:
Ok, then your approach seems the most reasonable :slightly_smiling_face:

philippjfr:
I've made the cut-off zero to one categorical columns, because those are the cases where it was consistently faster.

jbednar:
That all sounds good to me.

datashader/utils.py
```diff
@@ -391,7 +393,14 @@ def dshape_from_pandas(df):
 
 def dshape_from_dask(df):
     """Return a datashape.DataShape object given a dask dataframe."""
-    return datashape.var * dshape_from_pandas(df.head()).measure
+    cat_columns = [col for col in df.columns if isinstance(df[col].dtype, type(pd.Categorical.dtype))
+                   or isinstance(df[col].dtype, pd.api.types.CategoricalDtype)]
+    if len(cat_columns) > 1:
+        # If there is more than one categorical column it is faster
+        # to compute the df.head() which will contain all categories
+        return datashape.var * dshape_from_pandas(df.head()).measure
```
jbednar: I don't suppose it will get any faster if you specify `df.head(1)`, but maybe it's more explicit?  No strong opinion.

philippjfr: I tried, it's slower.

jbednar: Hmm; that's very strange.  Is `df.head(5)` (or whatever the default is) _also_ slower?  Seems fishy!

philippjfr: Not too surprising, it still has load the first partition and the network overhead difference between 1 and 5 rows is tiny. So the only difference is that it adds another task to the graph which runs the iloc.

datashader/utils.py
```diff
@@ -361,6 +361,8 @@ def dshape_from_pandas_helper(col):
     dataframe.
     """
     if isinstance(col.dtype, type(pd.Categorical.dtype)) or isinstance(col.dtype, pd.api.types.CategoricalDtype):
+        if not getattr(col.cat, 'known', True):
+            col = col.cat.as_known()
```
jcrist: This is potentially expensive, and since you're dropping the reference this is work that is lost. I'm not sure if there's a better way, users interested in performance should provide dataframes with known categories in beforehand.

philippjfr: >This is potentially expensive, and since you're dropping the reference this is work that is lost.

Right I considered assigning back to the original dataframe but that would be a bit too magical. I think if we can get memoization to work I won't feel so bad about this. 

While I've got you here, I remember you telling me that calling ``.head()`` was not usually sufficient for inferring the categories and that only fastparquet loads and stores the categories in metadata. Having played around with loading a variety of different datasets from csv, parquet loaded via pyarrow and simply converting a pandas dataframe to a dask dataframe, the `.head()` seems to have consistently returned all available categories even if they are not present in `.head()`. Is that just a quirk of the examples I've tried or can this be relied on afterall?

jcrist: That's just  quirk of the examples you've tried.

```
In [38]: df = pd.DataFrame({'a': ['a', 'b', 'b']})

In [39]: df2 = pd.DataFrame({'a': ['b', 'b', 'c']})

In [40]: ddf = dd.concat([df, df2])

In [41]: ddf2 = ddf.astype({'a': 'category'})

In [42]: ddf2
Out[42]:
Dask DataFrame Structure:
                               a
npartitions=2
               category[unknown]
                             ...
                             ...
Dask Name: astype, 6 tasks

In [43]: ddf2.head(1).a.cat.categories
Out[43]: Index(['a', 'b'], dtype='object')

In [44]: ddf2.a.cat.as_known().cat.categories
Out[44]: Index(['a', 'b', 'c'], dtype='object')
```

philippjfr: Thanks!

PR #801, opened 2019-10-04
Memoize dask datashape

philippjfr:
As suggested in https://github.com/pyviz/datashader/issues/797

### DIFF
--- a/datashader/utils.py
+++ b/datashader/utils.py
@@ -394,8 +394,9 @@ def dshape_from_pandas(df):
 @memoize(key=lambda args, kwargs: tuple(args[0].__dask_keys__()))
 def dshape_from_dask(df):
     """Return a datashape.DataShape object given a dask dataframe."""
-    cat_columns = [col for col in df.columns if isinstance(df[col].dtype, type(pd.Categorical.dtype))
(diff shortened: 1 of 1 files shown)


philippjfr:
Okay, the previous approach still wasn't saved. This new approach loads all the categories into memory in a way that is probably less efficient but guaranteed to be correct, but then adds memoization so you only pay the cost once.

jbednar:
Does that mean that we have to forego the speedups when reading fastparquet-written files with fastparquet, in order to make the other cases safe?  Or can we have both the speedup in the best case and the correctness in the other cases?

philippjfr:
>Does that mean that we have to forego the speedups when reading fastparquet-written files with fastparquet, in order to make the other cases safe?

I would assume that fastparquet categorical columns are known at the time of load so we would get the benefit but I'll have to test that assumption.

philippjfr:
>I would assume that fastparquet categorical columns are known at the time of load so we would get the benefit but I'll have to test that assumption.

This assumption was wrong but loading the categories using fastparquet is significantly faster than with pyarrow (although pyarrow seems to be faster than fastparquet in general). Therefore I do think it must be using the metadata in some form.

PR #807, opened 2019-10-09
Brendancol/viewshed testing

brendancol:
- [x] small fixes to viewshed numba dtypes
- [x] adds better tests for viewshed accuracy

### DIFF
--- a/datashader/spatial/viewshed.py
+++ b/datashader/spatial/viewshed.py
@@ -1,7 +1,10 @@
 import xarray
 import numpy as np
 import math
+
 from math import atan, sqrt, fabs
+from math import pi as PI
+
 import numba as nb
 from numba import jit
 
@@ -36,9 +39,9 @@ SMALLEST_GRAD = float('-inf')
 PROJ_LL = 0
 PROJ_NONE = -1
 
-PI = math.pi
 
-#TODO: refactor NAN value...
+# Future refactor?:
+# This value is a placeholder for np.nan in integer arrays
 NAN = -9999999999999999
 
 
--- a/datashader/spatial/viewshed.py
+++ b/datashader/spatial/viewshed.py
@@ -31,7 +31,7 @@ CENTER_EVENT = 0
 
 # this value is returned by findMaxValueWithinDist() if there is no key within
 # that distance
-SMALLEST_GRAD = -9999999999999999999999.0
+SMALLEST_GRAD = -9999999999999999.0
 
 PROJ_LL = 0
 PROJ_NONE = -1
@@ -95,6 +95,10 @@ class TreeValue:
 
         return self.grad[0]
 
+    def _print_tv(self):
+        print('key=', self.key, 'grad=', self.grad, 'ang=', self.ang,
+              'max_grad=', self.max_grad)
+
 
 class TreeNode:
 
@@ -122,7 +126,7 @@ class TreeNode:
             for l in self.left.__iter__():
                 yield l
 
-        yield self.val.key
+        yield self.val
 
         if self.right != NIL:
             for r in self.right.__iter__():
@@ -395,7 +399,6 @@ class Tree:
 
         cur_node = key_node
         max = SMALLEST_GRAD
-
         while cur_node.parent != NIL:
             if cur_node == cur_node.parent.right:
                 # its the right node of its parent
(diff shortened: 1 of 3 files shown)


datashader/spatial/viewshed.py
```diff
@@ -31,7 +31,7 @@
 
 # this value is returned by findMaxValueWithinDist() if there is no key within
 # that distance
-SMALLEST_GRAD = -9999999999999999999999.0
+SMALLEST_GRAD = -9999999999999999.0
```
jbednar: -Inf doesn't work?

brendancol:
@jbednar This is currently not passing the nbsmoke appveyor tests, but there were no changes to notebooks.  It looks like appveyor is missing a holoviews dependency.  is this a known issue?

jbednar:
@jsignell ?

datashader/spatial/viewshed.py
```diff
@@ -31,13 +31,14 @@
...
-SMALLEST_GRAD = -9999999999999999999999.0
+SMALLEST_GRAD = float('-inf')
 
 PROJ_LL = 0
 PROJ_NONE = -1
 
 PI = math.pi
 
+#TODO: refactor NAN value...
 NAN = -9999999999999999
```
jbednar: Is there a good reason not to simply use NaN and test for it with isnan?  I don't see anything special about the NAN in the code there that would require a separate implementation like this.

datashader/spatial/viewshed.py
```diff
@@ -31,13 +31,14 @@
 
 # this value is returned by findMaxValueWithinDist() if there is no key within
 # that distance
-SMALLEST_GRAD = -9999999999999999999999.0
+SMALLEST_GRAD = float('-inf')
 
 PROJ_LL = 0
 PROJ_NONE = -1
 
 PI = math.pi
```
jbednar: Why not just `from math import pi` and then use `pi`; why `PI`?

datashader/spatial/viewshed.py
brendancol: @jbednar Yes...all hold overs from C++ code.  I marked them for refactor and didn't hit it yet, but let's hold off on merging this and I'll refactor today.

datashader/spatial/viewshed.py
brendancol: @jbednar Yes...all hold overs from C++ code.  I marked them for refactor and didn't hit it yet, but let's hold off on merging this and I'll refactor today.

datashader/spatial/viewshed.py
jbednar: Ok, thanks.

datashader/spatial/viewshed.py
brendancol: @jbednar on second look, I like having the PI alias because its used everywhere, but it doesn't need to be so explicit...I've refactored accordingly and use an import alias instead of a variable

datashader/spatial/viewshed.py
brendancol: On second look, this NAN value is import as a token NAN value for integer dtype arrays.

datashader/spatial/viewshed.py
jbednar: Ok.

datashader/spatial/viewshed.py
jbednar: Damned integers with no NaN support!

jsignell:
> This is currently not passing the nbsmoke appveyor tests, but there were no changes to notebooks. It looks like appveyor is missing a holoviews dependency. is this a known issue?

This is super strange. The deps are the same and don't include holoviews, but on travis holoviews gets pulled in and on appveyor holoviews doesn't. The only difference is the channel. I guess we can try getting rid of conda-forge on appveyor and see what happens.

brendancol:
> > This is currently not passing the nbsmoke appveyor tests, but there were no changes to notebooks. It looks like appveyor is missing a holoviews dependency. is this a known issue?
> 
> This is super strange. The deps are the same and don't include holoviews, but on travis holoviews gets pulled in and on appveyor holoviews doesn't. The only difference is the channel. I guess we can try getting rid of conda-forge on appveyor and see what happens.

@jsignell cool cool. thanks for checking this out. I'll see if 

> > This is currently not passing the nbsmoke appveyor tests, but there were no changes to notebooks. It looks like appveyor is missing a holoviews dependency. is this a known issue?
> 
> This is super strange. The deps are the same and don't include holoviews, but on travis holoviews gets pulled in and on appveyor holoviews doesn't. The only difference is the channel. I guess we can try getting rid of conda-forge on appveyor and see what happens.

Cool I will try this and revert if it doesnt work

jsignell:
Seems like the python 2.7 failures are legit. 

```
UnsupportedParforsError: Failed in nopython mode pipeline (step: nopython mode backend)
The 'parallel' target is not currently supported on Windows operating systems when using Python 2.7, or on 32 bit hardware.
```

jsignell:
Probably not caused by this PR, but legit.

jbednar:
I'll go ahead and merge this even though I think we do need py27 tests, but apparently they shouldn't be on Appveyor anyway, because of limitations in Numba's parfors support on Windows.  So that will need to be a separate PR.

PR #813, opened 2019-10-13
Brendancol/bump mapping improvements

brendancol:
- [x] big `bump` in performance
- [x] refactored Geography notebook bump section with better description of user-defined `height_func` and better looking output with better tree visibility


brendancol:
duplicate

PR #818, opened 2019-10-23
Allow remote filesystems in spatial read_parquet and to_parquet

### LINKED ISSUE #817: Support remote filesystems in spatial read_parquet and to_parquet
I'd like to use the `to_parquet` and `read_parquet` functions in `datashader/spatial/points.py` with a remote filesystem (in this particular case with s3) if possible. E.g. something like:

```python
import datashader.spatial.points as dsp
storage_options = {"key": "...", "secret": "..."}
dsp.read_parquet("s3://bucket/file.parquet", storage_options=storage_options)
```

I'm happy to submit a PR, but first wanted to check if this type of functionality is in-scope for datashader.

jrbourbeau:
This PR updates `read_parquet` and `to_parquet` in `datashader/spatial/points.py` to allow for reading from and writing to remote filesystems

Closes #817


jrbourbeau:
Just wanted to note that I haven't yet tested if this works with file-like objects as proposed in https://github.com/pyviz/datashader/issues/817#issuecomment-545482897. Happy to do so though

martindurant:
I certainly can't think of a better way to fo than to use get_fs_token_paths, but it does seem a touch roundabout. Perhaps fsspec could benefit from more helpers along these lines.

+1 from me. You could test against one of the builtin FSs in fsspec (like memory or FTP).

jrbourbeau:
Thanks for the feedback @martindurant! I've added a test that uses the `"memory://"` protocol

> it does seem a touch roundabout

Do you mean that I'm getting `get_fs_token_paths` from `dask`, which itself imports `get_fs_token_paths` from `fsspec`?

martindurant:
I meant using the function `get_fs_token_paths` at all, doesn't quite seem right for this use case.

jbednar:
Sounds like no one has a better suggestion, so I'll go ahead and merge. Thanks!

datashader/tests/test_spatial.py
```diff
@@ -149,3 +149,15 @@ def test_validate_parquet_file(df, tmp_path):
...
 
     assert spf.spatial is None
+
+
+def test_filesystem_protocol(df, tmp_path):
+    # Work around https://bugs.python.org/issue33617
+    tmp_path = str(tmp_path)
+    p = 5
+    # Use an filesystem protocol
+    path = "memory://" + os.path.join(tmp_path, 'spatial_points_1.parquet')
```
jbednar:
```suggestion
    path = "memory://" + tmp_path + '/spatial_points_1.parquet')
```
(Suggestion from Martin Durant, as memoryfs at least currently expects posix rather than OS-native paths.)

martindurant: Just `path = "memory://tmp_path/spatial_points_1.parquet'` for simplicity. I hope it works!

jbednar: @jrbourbeau , can you please apply the above suggestion to see if it will fix the Appveyor errors?  It seems like it probably will, but I do not have write access to your fork to try it.

jbednar: > "memory://tmp_path/spatial_points_1.parquet'

Here `tmp_path` is a variable, right? Not the literal string "tmp_path"? So I think the path needs to be constructed using code like in my suggestion, not hardcoded like that.

jbednar: Oh, but in that case I guess tmp_path itself may contain backslashes? If so then this suggestion won't work at all, right?

martindurant: It is indeed a variable, but doesn't need to be, since we are to use a temporary in-memory filesystem anyway. I want to avoid anything happening due to string handling, such as "C:\\" messing up what is seen as the "parent" directory. That's not really a test of the filesystem mechanism, since each backend has different definitions of what valid paths look like.

jbednar: Ok.  As I can't edit this, I'll merge and fix it up on master, but I'll have to back out the changes if they don't work out.

jrbourbeau:
Thanks @jbednar @martindurant

PR #821, opened 2019-10-25
Updates for compatibility with cudf v0.10

jonmmease:
Various fixes for compatibility with version cudf version 0.10

### DIFF
--- a/conda.recipe/meta.yaml
+++ b/conda.recipe/meta.yaml
@@ -30,7 +30,8 @@ requirements:
     {% endfor %}
     # adding pyct here is temporary (this conda recipe template will disappear when new pyctdev is released)
     - pyct
-
+  run_constrained:
+    - cudf >=0.10.0
 test:
   imports:
     - datashader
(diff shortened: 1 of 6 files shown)


jbednar:
Which versions of cudf are now supported?  Can you express that with a conditional dependency in the conda setup?

jonmmease:
> Which versions of cudf are now supported? Can you express that with a conditional dependency in the conda setup?

Good point, I'll do some tests and update the conda recipe.

jonmmease:
Version constraint added. Unless others object, I'm going to make the lower bound 0.10.0 since cudf is still moving really fast, and these changes would require some fallback logic in order to support older versions as well.

jbednar:
I'm happy with that; if we ran into someone with a strong constraint pinning them to an older cudf version we could consider adding such support at that time, but meanwhile let's do the thing we can maintain most easily.  It's tricky when we're depending on something with a version like 0.10, because by semantic versioning we've no idea whether their next releases will break anything, yet if we pin directly to 0.10 and no higher, we'll guarantee that it won't support the next release.  So we're damned either way!  We'll just have to quickly cut a new datashader release if the next cudf version is incompatible.

PR #823, opened 2019-10-31
WIP: Viewshed Performance Increases

brendancol:
The following PR improves the performance of datashader's performance by changing the structures and not the algorithm (taken from GRASS).

In testing, the performance increase was about 75x:

Screen Shot 2019-10-30 at 10 58 25 PM

This isn't ready to merge yet...fixing some tests...


brendancol:
Looks like past commit got pulled in here...going to recreate...

PR #824, opened 2019-10-31
Viewshed Performance Fixes

brendancol:
The following PR is a significant refactor of viewshed to address performance problems.

The general approach can be summarized as "Don't change the algorithm, change the data structures".  In this spirit, no more `TreeValue` class etc. and instead np.array and ufuncs are used as much as possible.

Screen Shot 2019-10-30 at 10 58 25 PM

### DIFF
--- a/datashader/spatial/viewshed.py
+++ b/datashader/spatial/viewshed.py
@@ -8,6 +8,59 @@ from math import pi as PI
 import numba as nb
 from numba import jit
 
+E_ROW_ID = 0
+E_COL_ID = 1
+E_ELVEV_0 = 2
+E_ELVEV_1 = 3
+E_ELVEV_2 = 4
+E_ANG_ID = 5
+E_TYPE_ID = 6
+
+S_DIST2VP = 0
+S_ROW_ID = 1
+S_COL_ID = 2
+S_GRAD_0 = 3
+S_GRAD_1 = 4
+S_GRAD_2 = 5
+S_ANG_0 = 6
+S_ANG_1 = 7
+S_ANG_2 = 8
+
+TN_KEY_ID = 0
+TN_GRAD_0 = 1
+TN_GRAD_1 = 2
+TN_GRAD_2 = 3
+TN_ANG_0 = 4
+TN_ANG_1 = 5
+TN_ANG_2 = 6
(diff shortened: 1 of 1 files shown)


datashader/spatial/viewshed.py
```diff
@@ -1,30 +1,23 @@
...
 import numpy as np
-import math
-
-from math import atan, sqrt, fabs
-from math import pi as PI
-
 import numba as nb
 from numba import jit
+from math import atan, sqrt, fabs
+from math import pi as PI
```
jbednar: Usually imports should be in order from least to most specific, so stdlib imports like those from `math` here should move _above_ those from `numba`.

jbednar: In fact those should go at the very top, above xarray and numpy as well.

datashader/spatial/viewshed.py
```diff
@@ -87,12 +53,7 @@
 
 # this value is returned by findMaxValueWithinDist() if there is no key within
 # that distance
-SMALLEST_GRAD = float('-inf')
-
-PROJ_LL = 0
-PROJ_NONE = -1
-
-NAN = -9999999999999999
+SMALLEST_GRAD = -9999999999999999999999.0
```
jbednar: What made this have to revert to an arbitrary constant? A switch from float to int?

brendancol: there were warning when placing this value `-inf` values in arrays. This may not be as pretty, but `float(-inf)` and `np.NINF` through invalid value warnings.

brendancol: the NAN constand is no longer being used

brendancol: This may or may not be related to your question, but the NOJIT tests revealed an issue where numba was impliciting casting floats to ints, when they should have been ints to begin with. a few days ago I fixed the NOJIT case by manually calling `int()` in some spots.  To be clear, numba was being helpful...but I want to avoid the additional casting so did some more refactoring.  the int NAN value was no longer need and also the `_is_null` wrapper around the int NAN constrant was removed.

datashader/spatial/viewshed.py
brendancol: @jbednar done

PR #837, opened 2019-11-29
More pyviz -> holoviz (and test external codecov rename).

ceball:
Code coverage links were 404, so tried to update at codecov.io. Fix travis badge. Also, update travis and github links (which were still working).


ceball:
I should have labelled this as draft - I was testing to see if coverage appeared. However, I don't see anything even after these changes and changes at codecov.io. 

I don't think I set up coverage originally, but I can't remember.

ceball:
(Not that merging is a problem.)

Actually, I don't see anything about coverage at all in the repository, unless I'm confused...

jbednar:
I don't think coverage is set up, but it certainly should be.

ceball:
Ok. I see the same for geoviews and hvplot (i.e. coverage.io badges apparently not going anywhere).

PR #838, opened 2019-11-29
Collect test coverage

ceball:
I've never used coverage.io before. I copied what's in panel (don't know what it all does), and I added the token to travis.


ceball:
(draft til we see if coverage shows up)

ceball:
Something has shown up https://codecov.io/gh/holoviz/datashader, indicating coverage will show up for master. Don't know about getting coverage reports on pull requests - we'll see if that starts happening or if we need to turn on something else.

PR #843, opened 2019-12-06
fix dead links in "Getting Started -> Pipeline"

tyrasd:


tyrasd:
probably related: https://datashader.org/getting_started/Pipeline.html currently has a dead link in this place as well (to https://datashader.org/user_guide/1-Plotting_pitfalls.html).

Issue #835, opened 2019-11-25
Geo-operations drop coordinates


philippjfr:
Operations like hillshade, slope, aspect, ndvi etc. all seem to return xarray DataArrays where the dimensions are unnamed and no coordinates are defined. I'd expect all of these operations to retain the dimensions and coordinates of the inputs.

brendancol:
@philippjfr good catch.  I will take this one and submit PR asap

PR #844, opened 2019-12-06
Adds missing `dims` and `coords` to geo transfer functions

brendancol:
Adds missing `dims` and `coords` to geo transfer functions (e.g. slope, aspect, ndvi, hillshade).


brendancol:
fixes #835

Issue #852, opened 2019-12-16
Lower bound is non-inclusive on polygon aggregation


philippjfr:
I'd expect the x- and y-range lower bound to be inclusive in the aggregation: 

```
import datashader as ds
from spatialpandas import GeoDataFrame, GeoSeries
from spatialpandas.geometry import Polygon

p1 = Polygon([[0, 0, 2, 0, 1, 1, 0, 0]])
p2 = Polygon([[0, 0, 1, 1, 0, 1, 0, 0]])

gdf = GeoDataFrame({'geometry': GeoSeries([p1, p2])})

cvs = ds.Canvas(plot_width=5, plot_height=5)

cvs.polygons(gdf, geometry='geometry')
```

i.e. in this output I'd expect the first row and column to be at least partially filled:

```
<xarray.DataArray (y: 5, x: 5)>
array([[False, False, False, False, False],
       [False,  True,  True,  True,  True],
       [False,  True,  True,  True,  True],
       [False,  True,  True,  True, False],
       [False,  True,  True,  True, False]])
Coordinates:
  * x        (x) float64 0.2 0.6 1.0 1.4 1.8
  * y        (y) float64 0.1 0.3 0.5 0.7 0.9
```

A visualization of the polygon with the aggregate on top:

![bokeh_plot - 2019-12-16T154316 511](https://user-images.githubusercontent.com/1550771/70915954-21b5f480-201b-11ea-8823-0b389ecfc040.png)

Looking at that it also looks like (0, 3) and (2, 4) are filled even though they shouldn't be.

jonmmease:
For reference, for pixels that fall exactly on a polygon edge, the expected behavior is:
 - If if edge is direct up or to the leftthen the pixel is filled.
 - Otherwise, if the edge is directed down or to the right, it is not filled.

I'll need to dig in a bit more, but there might be a half pixel offset going on here.

Issue #806, opened 2019-10-09
Trimesh rasterization seems slightly off


ronaldsipkema:
Please edit the template below, attempting to fill in as much info as possible (it's all optional, but the more you fill in, the faster we can help). Thanks!

When pasting code, tracebacks, or console output, please enclose in triple backticks (https://help.github.com/articles/basic-writing-and-formatting-syntax/#quoting-code).

## Description

While playing around with Canvas.trimesh and rasterizing simple geomtery it seems that the rule described in the documentation ("a pixel (bin) is treated as belonging to a given triangle if its center falls either inside that triangle or along its top or left edge") is not adhered to, i.e., some pixels that do not belong to a triangle get a value, and the other way around.

![rasterized_circle](https://user-images.githubusercontent.com/5845408/66497539-e946e700-eabc-11e9-8444-3ce0921bc2d1.png)

## How to reproduce

```
import pandas as pd
import shapely.geometry
from scipy.spatial import Delaunay
import datashader as ds
import matplotlib.pyplot as plt


def triangulate(vertices, x="x", y="y"):
    triang = Delaunay(vertices[[x,y]].values)
    return pd.DataFrame(triang.simplices, columns=['v0', 'v1', 'v2'])


shape = shapely.geometry.Point(0, 0).buffer(1)
vertices = pd.DataFrame(shape.exterior.coords, columns=('x', 'y'))
vertices['z'] = 1.0

simplices = triangulate(vertices)

canvas = ds.Canvas(plot_width=10, plot_height=10,
                x_range=(-1, 1), y_range=(-1, 1))

canvas.trimesh(vertices, simplices, agg=ds.any()).plot(alpha=0.5)
plt.plot(vertices.x, vertices.y)

```

jbednar:
Thanks for the report!  I keep hoping to have a chance to look into it, but haven't yet.  I can see that the array pattern shown does not match the Shapely circle, but as the triangles are not visible I can't see how the triangles relate to the array pattern, so I can't see where the discrepancy arises.

jbednar:
Oh, and I wasn't able to run the code above, due to a Pandas issue:

```
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-1-d9e37701cbad> in <module>
     12 
     13 shape = shapely.geometry.Point(0, 0).buffer(1)
---> 14 vertices = pd.DataFrame(shape.exterior.coords, columns=('x', 'y'))
[...]

PR #853, opened 2019-12-16
Fix half-pixel offset error in Polygon rasterization

jonmmease:
Closes https://github.com/holoviz/datashader/issues/852

Subtract a half-pixel offset when mapping vertex coords to into canvas space. This way the scanline algorithm is checking whether the center of each pixel is in the polygon, rather than the lower left corner.

```python
import datashader as ds
from spatialpandas import GeoDataFrame, GeoSeries
from spatialpandas.geometry import Polygon

p1 = Polygon([[0, 0, 2, 0, 1, 1, 0, 0]])
p2 = Polygon([[0, 0, 1, 1, 0, 1, 0, 0]])

gdf = GeoDataFrame({'geometry': GeoSeries([p1, p2])})

cvs = ds.Canvas(plot_width=5, plot_height=5)

cvs.polygons(gdf, geometry='geometry')
```
```
<xarray.DataArray (y: 5, x: 5)>
array([[ True,  True,  True,  True,  True],
       [ True,  True,  True,  True, False],
       [ True,  True,  True,  True, False],
       [ True,  True,  True, False, False],
       [ True,  True,  True, False, False]])
Coordinates:
  * x        (x) float64 0.2 0.6 1.0 1.4 1.8
  * y        (y) float64 0.1 0.3 0.5 0.7 0.9
```

cc @philippjfr


jbednar:
It would be great to see the same fix for https://github.com/holoviz/datashader/issues/806 :-)

philippjfr:
Looks good, thanks for the quick fix. Seems like there's a flake. np.nan imported but unused in test_polygons.py.

jonmmease:
@philippfr thanks! I can push the flake fix in an hour or so, but feel free to push to the branch if you want.

@jbednar yeah, that does look like the same kind of issue.  I’m not nearly as familiar with the Trimesh logic though, so I don’t want to make that change in this PR.  But I can take a look at it next time i’m working in that codepath.

jbednar:
Thanks!

PR #856, opened 2019-12-22
Try removing nbsmoke pin.

ceball:


setup.py
```diff
@@ -42,7 +42,7 @@
         'pytest-cov',
         'codecov',
         'flake8',
-        'nbsmoke ==0.2.8',  # test pinning to allow hv.extension
```
ceball: @jsignell (or @jbednar), do you have any more details about this? Did nbsmoke 0.3 break something? (The change was made in #829)

jbednar: No idea, but given that hv.extension has been available for a very long time, the pinning might have been put there to avoid nbsmoke 0.2.7 and earlier, rather than 0.3 and later.

jbednar: The PR shows that the pin was changed from >0.2.6 to ==0.2.8, which I think indicates that 0.3 is not the one that was being avoided.  So I assume the pinning can be changed to >=0.2.8.

jsignell: I suspect that adding holoviews to the examples was the real fix and this can be undone. If it passes tests, it can definitely be undone.

jbednar: Ok, sounds good. @ceball , shall I merge?

ceball: nbsmoke 0.4.0 is not on defaults (though it's on conda-forge). Apart from that it should be fine.

PR #861, opened 2020-01-16
GPU Quadmesh support

jonmmease:
This PR adds support for aggregating `quadmesh` glyphs using cuda.  The cuda codepath is invoked when the data backing the provided `xarray.DataArray` is a `cupy.ndarray`.

I'm seeing 20-50x speedups in both the rectilinear and curvilinear cases.

Here are some examples: https://anaconda.org/jonmmease/gpu_quadmesh_pr/notebook

### DIFF
--- a/datashader/data_libraries/xarray.py
+++ b/datashader/data_libraries/xarray.py
@@ -5,13 +5,18 @@ from datashader.core import bypixel
 import xarray as xr
 from datashader.utils import Dispatcher
 
+try:
+    import cupy
+except ImportError:
+    cupy = None
 
 glyph_dispatch = Dispatcher()
 
 
 @bypixel.pipeline.register(xr.Dataset)
 def xarray_pipeline(df, schema, canvas, glyph, summary):
-    return glyph_dispatch(glyph, df, schema, canvas, summary)
+    cuda = cupy and isinstance(df[glyph.name].data, cupy.ndarray)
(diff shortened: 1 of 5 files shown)


philippjfr:
I certainly wouldn't mind to announce this along with the other GPU support in HoloViews 1.13.0.

jonmmease:
FWIW, I'm starting to look at GPU support for raster, but that's going be much more involved and not ready for a while.  So, I'd say go ahead with releasing Quadmesh support (whenever that is) separately from raster.

jbednar:
Would it be possible to run a benchmark where you convert a raster to a quadmesh and compare GPU quadmesh vs. CPU raster?  That might be a speedup already, and then we can wait for the real raster speedup when it's ready.

jonmmease:
> Would it be possible to run a benchmark where you convert a raster to a quadmesh and compare GPU quadmesh vs. CPU raster?

Sure, that's a good idea.

jonmmease:
Unfortunately there's not much benefit (at least on my workstation).  The raster implementation is parallelized well on the CPU and is much faster than the single threaded quadmesh implementation.

The GPU quadmesh is about on par with the raster (with a factor of 2 either way depending on input size).

jbednar:
Good to know, thanks!  It's always interesting to see how hardware and algorithm choice interact to give the final performance on any particular problem.

PR #872, opened 2020-01-31
Add optimized raster codepath to quadmesh glyph with GPU support

jonmmease:
## Overview
This PR introcuces a new optimized codepath to the quadmesh glyph for handling uniformly spaced rasters. This is a first step towards the goal unifying the functionality of the raster and quadmesh glyphs.

Like the rest of the quadmesh codepaths (rectilinear and curvilinear), this new raster codepath supports GPU processing when provided with a cupy backed xarray Dataset as input. Unlike the other codepaths,  CPU processing is multi-threaded.

## Benchmarking
This codepath generally matches or slightly exceeds the performance of the current `raster` implementation in CPU mode, and exceeds this performance in GPU mode by a factor of ~15-20x on my workstation.

See https://anaconda.org/jonmmease/quadmesh_raster_gpu_pr/notebook for specific upsampling and downsampling benchmark results

## Not implemented
The following features of the current `raster` implementation are *not* implemented in this PR:
 - linear upsample interpolation (current behavior is equivalent to `raster`'s `'nearest'`)
 - Dask support
 - custom `nan_value` specification
 - The `layer` argument for working with 3D input arrays


jonmmease:
After a bit of digging, it looks like these CI failures are due to some incompatibilites with the pandas 1.0 release.

jbednar:
- [ ] We'll also need to mention in the Raster docs that people with a GPU should consider using the quadmesh glyph for their rasters when the above conditions are met.

jonmmease:
@jbednar tests passing now that pandas 1.0 compatibility was merged.

Issue #140, opened 2016-04-14
Generalizing support for categorical data


jbednar:
Right now, we offer the `count_cat` reduction operator, which effectively does the `count` reduction separately per category.  Most of the other reduction operators would also be meaningful to do categorically, and it would be a pain to implement a categorical version of each one as it is needed.  

Instead, we should be able to implement a general-purpose categorical operator, similar to the `summary` operator, which takes a scalar reduction operator and applies it per category.  @jcrist suggests a syntax like `by('category_column', mean('some_other_column'))`, which seems clear to me as well.  I think we should do this as soon as we need to apply any operator other than `count` to categories, since making a general version is probably not much more difficult than copying count_cat and changing the counting bits.

PR #875, opened 2020-02-14
Generalize categorical aggregation and support span in _colorize

maihde:
I have categorical data where I want to sum a column, but grouped by category.  The `count_cat` functionality simply counts the number of rows of each category.  This PR creates a new aggregation called `sum_cat`.

In addition, I often need to produce multiple datashader visualizations that are colorized where the alpha range is distributed across a fixed span.  This PR enables support for using `span` on categorical data.

Please provide comments or suggestions on how to improve the PR if necessary.

Practical examples can be found here:
https://github.com/spectriclabs/jupyter-notebooks/blob/master/Datashader-Improvements.ipynb

### DIFF
--- a/datashader/reductions.py
+++ b/datashader/reductions.py
@@ -143,7 +143,7 @@ class by(Reduction):
     ----------
     column : str
         Name of the column to aggregate over. Column data type must be
-        categorical. Resulting aggregate has a outer dimension axis along the
+        categorical. Resulting aggregate has an outer dimension axis along the
         categories present.
     reduction : Reduction
         Per-category reduction function.
@@ -502,6 +502,7 @@ class max(FloatingReduction):
 
 class count_cat(by):
     """Count of all elements in ``column``, grouped by category.
+    Alias for `by(...,count())`, for backwards compatibility.
 
     Parameters
     ----------
@@ -725,4 +726,3 @@ __all__ = list(set([_k for _k,_v in locals().items()
                     if isinstance(_v,type) and (issubclass(_v,Reduction) or _v is summary)
                     and _v not in [Reduction, OptionalFieldReduction,
                                    FloatingReduction, m2]]))
-
--- a/datashader/reductions.py
+++ b/datashader/reductions.py
@@ -51,6 +51,29 @@ class category_codes(Preprocess):
         else:
             return df[self.column].cat.codes.values
 
+class category_values(Preprocess):
+    """Extract multiple columns from a dataframe as a numpy array of values."""
+    def __init__(self, columns):
+        self.columns = list(columns)
+
+    @property
+    def inputs(self):
+        return self.columns
+
+    def apply(self, df):
+        if cudf and isinstance(df, cudf.DataFrame):
+            import cupy
+            if df[self.columns[1]].dtype.kind == 'f':
+                nullval = np.nan
+            else:
+                nullval = 0
+            a = df[self.columns[0]].cat.codes.to_gpu_array()
+            b = df[self.columns[1]].to_gpu_array(fillna=nullval)
+            return cupy.stack((a, b), axis=-1)
+        else:
+            a = df[self.columns[0]].cat.codes.values
+            b = df[self.columns[1]].values
+            return np.stack((a, b), axis=-1)
 
 class Reduction(Expr):
     """Base class for per-bin reductions."""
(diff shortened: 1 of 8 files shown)


jbednar:
This looks great!  Thanks for working on this.  We've had an issue open for years now to generalize count_cat (https://github.com/holoviz/datashader/issues/140), but the limitation has never affected our own work, so it's never gotten addressed.  And I don't think we'd ever noticed that span wasn't implemented for categoricals! Some notes:

- [ ] There are various tests failing right now that would need to be addressed
- [ ] Originally, the idea was to create a single categorical "meta-aggregator" that would accept any reduction function and compute it categorically. I.e., instead of count_cat(col) and sum_cat(col), it would be categorical(count,col) and categorical(sum,col).  We could then offer count_cat and sum_cat as simple macros that would be useful when using a string to specify the aggregator, but wouldn't need to try to implement each of the various reduction operators categorically. Having implemented a second categorical operator, how difficult do you think it would be to make it be a single general operation with a reduction-operator argument?
- [ ] Once it's part of the codebase, it would need to be documented in the user guide, but it would be ok to leave that job for me since you've already made such nice examples.

maihde:
Thanks James,

* I'll review the tests that are failing, but I don't believe they are due
to my changes because the master branch appears to fail tests without my
changes included.

* I'll take a look at refactoring into a more generic categorical
aggregator.  I don't think it will be that difficult.

* I'll take a stab at helping with the documentation, but will greatly
appreciate your assistance as well.

jbednar:
It looks like the failing tests on master are due to API changes for Pandas 1.0, which we'll address separately. For now we've pinned Pandas <1.0, so can you please rebase this PR (or patch in the changes from https://github.com/holoviz/datashader/pull/876) so that the tests can run properly?

maihde:
James,

I've fixed up the tests and my branch now passes without error.  I'm in the
process of trying to make the more generic categorical reduction, but I
figured that might be best handled as a separate PR.

~Michael

jbednar:
Sounds good!  I'll leave this open until then, as it's already useful as-is, but can close it as soon as the more general one is available. Thanks so much!

maihde:
James, take a look at the most recent push of a generic categorical
reduction.  If this is what you were looking for, I can work on some
additional unit tests.

Thanks,
Michael

jbednar:
Cool, thanks! Presumably this implementation would now replace count_cat and sum_cat?  Presumably we'd leave a wrapper to support count_cat for old code, but now simply calling the new code, and we'd delete sum_cat entirely.
 
For the API for the new functionality, I think we should probably call it "by" as suggested in #140, because even though it is currently limited to a categorical axis, in the future it doesn't seem terribly difficult to generalize this approach to handle numerical axes as well (by first binning the value then treating it like a category). So for now it would be:

```
class by(Reduction):
    """Apply the provided reduction separately per categorical ``column`` value.
    Parameters
    ----------
    column : str
        Name of the column to aggregate over. Column data type must be
        categorical. Resulting aggregate has a outer dimension axis along the
        categories present.
    reduction : Reduction
        Per-category reduction function.
    """
```

maihde:
That all sounds good.  I didn't want to go to the hassle of replacing
count_cat/sum_cat and such until I had confirmation that I was on the right
track.  I will finish up the implementation and change the name to 'by' and
update the PR.

maihde:
I've just pushed up an update the renames the reduction to `by()`, removes
`sum_cat`, refactors `count_cat` to use the new `by`, and adds some
additional unit-tests.

maihde:
@jbednar any thoughts on this PR?  I think I have finished everything and this is ready to merge into master.

datashader/reductions.py
```diff
@@ -380,7 +500,7 @@ def _combine(aggs):
         return np.nanmax(aggs, axis=0)
 
 
-class count_cat(Reduction):
+class count_cat(by):
     """Count of all elements in ``column``, grouped by category.
```
jbednar:
```suggestion
    """Count of all elements in ``column``, grouped by category.
    Alias for `by(...,count())`, for backwards compatibility.
```

examples/environment.yml
```diff
@@ -20,7 +20,7 @@ dependencies:
  - networkx>=2.0
  - numba
  - numpy
- - pandas
+ - pandas >=0.24.1,<1.0
```
jbednar:
```suggestion
 - pandas >=0.24.1
```

(master should now be able to support Pandas >1.0, though this PR may need to be rebased)

doc/api.rst
```diff
@@ -66,6 +66,7 @@ Reductions
    any
    count
    count_cat
+   sum_cat
```
jbednar:
```suggestion
   by
```

mattastley:
@jbednar I merged in the minor changes and it should be ready to merge into master.

jbednar:
Thanks!!!

PR #877, opened 2020-02-14
Remove apparently redundant specifications of pandas in dependencies

ceball:
Anything in install_requires will always be present, so doesn't need to be specified elsewhere.

Maybe there have been (or will be) times when the pandas dependency must be specified differently in some particular situation (e.g. building the website), but looks like not at the moment.

Unless there's some problem I don't know about...?


jbednar:
Thanks!

PR #880, opened 2020-02-18
ImportError -> Exception for guards around cupy/cudf import

jonmmease:
Currently, datashader imports `cupy` and `cudf` inside `try`/`except` blocks, and proceeds gracefully when an `ImportError` is raised.

I ran into a situation today where I was using a conda environment that had cupy installed, but without an NVIDIA GPU. When attemping to import cupy, an `AttributeError` was raised inside cupy.  since it was not an `ImportError`, this exception propagates all the way up and makes Datashader unusable in that environment.

With these changes, if any exception is raised while attempting to import `cupy` or `cudf`, Datashader will continue on without GPU support.

### DIFF
--- a/datashader/core.py
+++ b/datashader/core.py
@@ -20,12 +20,12 @@ from . import reductions as rd
 
 try:
     import cudf
-except ImportError:
+except Exception:
     cudf = None
 
 try:
     import dask_cudf
-except ImportError:
+except Exception:
     dask_cudf = None
 
 class Axis(object):
@@ -201,7 +201,7 @@ class Canvas(object):
                         source.spatial is not None and
                         source.spatial.x == x and source.spatial.y == y and
                         self.x_range is not None and self.y_range is not None):
-            
+
                     source = source.spatial_query(
                         x_range=self.x_range, y_range=self.y_range)
             glyph = Point(x, y)
--- a/datashader/data_libraries/__init__.py
+++ b/datashader/data_libraries/__init__.py
@@ -14,5 +14,5 @@ try:
     import dask_cudf as _dask_cudf  # noqa (Test dask_cudf installed)
     from . import dask_cudf         # noqa (API import)
 
-except ImportError:
+except Exception:
     pass
--- a/datashader/data_libraries/xarray.py
+++ b/datashader/data_libraries/xarray.py
@@ -7,7 +7,7 @@ from datashader.utils import Dispatcher
 
 try:
     import cupy
-except ImportError:
+except Exception:
     cupy = None
 
 glyph_dispatch = Dispatcher()
--- a/datashader/glyphs/area.py
+++ b/datashader/glyphs/area.py
@@ -11,7 +11,7 @@ from numba import cuda
 try:
     import cudf
     from ..transfer_functions._cuda_utils import cuda_args
-except ImportError:
+except Exception:
     cudf = None
     cuda_args = None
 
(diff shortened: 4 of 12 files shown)


jbednar:
Thanks. Wouldn't it be safer to check for either ImportError or AttributeError, to avoid hiding other errors that might be meaningful?

jbednar:
I guess there's a middle ground to consider as well, which would be to leave it checking for Exception everywhere except core.py. In core.py, you could first checking for ImportError or AttributeError and continue happily for either one, and then for general Exception you could print a warning that an unexpected error was encountered, but then still continue anyway.  That way it would be robust against future errors in the cu modules, but would still report when something unusual happens.

ceball:
> Wouldn't it be safer to check for either ImportError or AttributeError, to avoid hiding other errors that might be meaningful?

I'm a bit out of touch now, but I think what Jon has done is the standard thing that everyone does. You don't want any problem during import of an optional module to stop someone from being able to import and use your module, and trying to figure out what specifically to catch in all circumstances for an external package can be a bit tricky - hence catching everything. (Not that this would necessarily avoid all problems resulting from an attempted import gone wrong, but I guess it gets most.)

But, this means that people who _should have_ or _meant to use_ an optional module never learn it's not being used (except e.g. through poor performance). I guess that's what Jim is getting at? 

If the optional module told you all the things to watch out for being raised on import in various scenarios, or you figured them out, that would be ok. I'm not sure everything's known here, though? (Plus, just catching classes of exception probably isn't enough to be totally sure, vs catching specific ones.)

Otherwise, more extensive logging could be one general option (although people would have to look at the logs).

I guess another option could be telling people to declare that they want something, e.g. "DEFINITELY_USE_MY_GPU=1 python -c import datashader" kind of a thing, then telling only those people they don't have what they wanted because e.g. some import failed. This seems unwieldy. 

Or the above, but via the test suite. I.e. tell people that after installing, if they want to be sure it's working with their optional module/hardware/whatever, they should run the test suite with DEFINITELY_USE_MY_GPU=1 or --definitely-use-my-gpu or whatever.

Another possibility might be that we could check if one or modules _exists_ (i.e. should be importable) before trying to import it, and only try to actually import it if so; if it subsequently fails to import (for any reason), print a warning and the traceback (on the assumption that someone who has it meant for it to be used).

jonmmease:
Thanks for weighing in @jbednar and @ceball,

> Wouldn't it be safer to check for either ImportError or AttributeError, to avoid hiding other errors that might be meaningful?

I didn't dig into the stack trace, but I suspect that the `AttributeError` that I hit wasn't actually an intended user-facing exception from cupy.  So I guess it didn't seem right to me to add a special case for it in Datashader.

> But, this means that people who should have or meant to use an optional module never learn it's not being used (except e.g. through poor performance). I guess that's what Jim is getting at?

So this is a nice thing about the Datashader design. GPU support is always and only performed if you pass in a GPU data structure.  If we automatically enabled GPU acceleration on the traditional data structures when a compatible GPU was available, then I would agree that it wouldn't be wise to swallow general exceptions without alerting the user.  But in our case, if a user wants GPU support they will need to import cudf/cupy themselves to build up the right data structure, in which case they will see the full import error for themselves.

So that was my reasoning.

ceball:
Ha ha, sorry, my answer was too generic and hasty. 

I haven't had chance to try the GPU support yet 🙁

I'll shut up 😂

jbednar:
Ok, sounds good! Thanks.

PR #882, opened 2020-02-21
Do not automatically trigger the creation of a spatialpandas spatial index

jonmmease:
Slight change in Datashader's use of `spatialpandas` that I made in order to match the legacy performance of https://examples.pyviz.org/osm/osm-1billion.html#osm-gallery-osm-1billion.

With this change, Datashader will use the per-partition spatial index if one is present but it will not trigger the creation of a new spatial index.  Datashader will still always use/create a spatialindex across all of the partitions in order to judge which partitions are needed.


jbednar:
Looks good. Feel free to merge once the PR build passes.

PR #888, opened 2020-04-06
Fixed InteractiveImage bokeh 2.0 compatibility

philippjfr:
Makes InteractiveImage compatible with bokeh 2.0 and deprecates it.

### DIFF
--- a/datashader/bokeh_ext.py
+++ b/datashader/bokeh_ext.py
@@ -4,10 +4,10 @@ from distutils.version import LooseVersion
 
 import uuid
 import json
+import warnings
 
 import numpy as np
 import bokeh
-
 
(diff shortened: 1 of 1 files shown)


datashader/bokeh_ext.py
```diff
@@ -219,6 +234,10 @@ class InteractiveImage(object):
 
     def __init__(self, bokeh_plot, callback, delay=200, timeout=2000, throttle=None,
                  **kwargs):
+        warnings.warn('InteractiveImage has been deprecated as of 0.11.0. '
```
jbednar, line 237: It should say that it's been deprecated since 0.8.0, though at that time it was only in the release notes.

PR #894, opened 2020-04-09
Update deprecations

philippjfr:
Deprecates:

- `datashader.geo`
- `datashader.spatial`

### DIFF
--- a/datashader/geo.py
+++ b/datashader/geo.py
@@ -5,17 +5,28 @@ This module contains geoscience-related transfer functions whose use is complete
 
 from __future__ import division, absolute_import
 
+import warnings
+
 import pandas as pd
 import numpy as np
 import datashader.transfer_functions as tf
 
 from datashader import Canvas
 from datashader.colors import rgb
-from datashader.utils import ngjit, lnglat_to_meters   # noqa (API import)
+from datashader.utils import (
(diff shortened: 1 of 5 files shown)


datashader/geo.py
```diff
@@ -5,19 +5,30 @@
...
 from xarray import DataArray
 
 __all__ = ['mean', 'binary', 'slope', 'aspect', 'ndvi', 'hillshade', 'generate_terrain',
            'lnglat_to_meters']
 
 
+warnings.warn(
+    "The datashader.geo module is deprecated as of version 0.11.0. "
+    "Its contents have migrated to the xarray_spatial library.",
+    VisibleDeprecationWarning
```
jbednar, line 28: Should we link to the websites for xarray-spatial and spatialpandas libraries?  It's odd for an error message to do that in general, but those aren't necessarily easy to find. A google search for "xarray-spatial" doesn't show the right library on the first page of results.

datashader/spatial/__init__.py
```diff
@@ -8,3 +11,13 @@
 
 from datashader.spatial.viewshed import viewshed  # noqa (API import)
 from datashader.spatial.zonal import zonal_stats  # noqa (API import)
+
+from ..utils import VisisbleDeprecationWarning
```
jbednar, line 15:
```suggestion
from ..utils import VisibleDeprecationWarning
```

Issue #895, opened 2020-04-12
_colorize throw exception if input is all zeros


maihde:
If the input to colorize is an all zero array, an exception is throw.

ValueError: zero-size array to reduction operation minimum which has no identity

The expected behavior is that an image is rendered with all pixels at min-alpha.  The fix is to check the `total` array with `if not np.all(total)` before executing `offset = total[total > 0].min()`.  I will be submitting a PR with test-case and fix shortly.

jbednar:
Sounds good, thanks!

PR #896, opened 2020-04-12
Misc fixes for colorize

maihde:
This PR includes two fixes.  If you would like them split into separate PR, let me know. 

The first is a fix for #895, which is a simple fix to handle the corner case where the input is all zeros.

The second is a fix to correctly handle fixed spans in `_colorize`.  In the original implementation found in #875 the alpha wouldn't be correctly calculated.  This would be especially apparent in corner cases where min-span was greater than the smallest value of max-span was less than the largest values.  This has been fixed and additional unittests have been added to verify alpha is being correctly calculated.

### DIFF
--- a/datashader/transfer_functions/__init__.py
+++ b/datashader/transfer_functions/__init__.py
@@ -357,22 +357,31 @@ def _colorize(agg, color_key, how, span, min_alpha, name):
     # min/max of the data
     if span is None:
         span = [np.nanmin(total).item(), np.nanmax(total).item()]
+        offset = span[0]
         if how != 'eq_hist':
             span = _normalize_interpolate_how(how)([0, span[1] - span[0]], 0)
+        # If there are elements that are zero or lower, they should be fully
+        # transparent and masked out
+        if offset <= 0:
+            mask = mask | (total <= 0)
+            # If at least one element is not masked, use the minimum as the offset
+            # otherwise the offset remains at zero
+            if not np.all(mask):
+                offset = total[total > 0].min()
     else:
         if how == 'eq_hist':
             # For eq_hist to work with span, we'll need to store the histogram
             # from the data and then apply it to the span argument.
             raise ValueError("span is not (yet) valid to use with eq_hist")
+        # even in fixed-span mode cells with 0 should remain fully transparent
+        # i.e. a 0 will be fully transparent, but any non-zero number will
+        # be clipped to the span range and have min-alpha applied
+        mask = mask | (total <= 0)
+        # clip all other values to the fixed span
+        masked_clip_2d(total, mask, *span)
+        offset = span[0]
         span = _normalize_interpolate_how(how)([0, span[1] - span[0]], 0)
 
-    offset = np.array(span, dtype=data.dtype)[0]
-    if offset == 0:
-        mask = mask | (total <= 0)
-        # If at least one element is not masked, use the minimum as the offset
-        # otherwise the offset remains at zero
-        if not np.all(mask):
-            offset = total[total > 0].min()
     a = _normalize_interpolate_how(how)(total - offset, mask)
     # Interpolate the alpha values
     a = interp(a, array(span),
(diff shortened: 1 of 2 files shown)


datashader/transfer_functions/__init__.py
```diff
@@ -357,19 +357,31 @@ def _colorize(agg, color_key, how, span, min_alpha, name):
     # min/max of the data
     if span is None:
         span = [np.nanmin(total).item(), np.nanmax(total).item()]
+        offset = span[0]
         if how != 'eq_hist':
             span = _normalize_interpolate_how(how)([0, span[1] - span[0]], 0)
+        # If there are elements that are zero or lower, they should be fully
+        # transparent and masked out
```
jbednar, line 364: I'm having trouble following this logic. Negative values in the original data are valid for displaying, so why would they be masked out here?

philippjfr: As we were just discussing, maybe this was written without taking into account the fact that categorical aggregates can now be performed with any reduction, including mean, sum etc and can therefore be negative. Maybe we need some way to indicate counts?

maihde: @jbednar @philippjfr in my primary use case the issue wasn't the new reductions, but rather using `_colorize` with a fixed span (which is also a new feature).  If I set span to sometime like (10, 100) and I have pixel with a count of 5 the old code would subtract 10 and arrive at -5.  I believe the correct behavior is similar to how fixed spans behave in other parts of the code.  All non-zero numbers are clipped to a range 10-100.

You can see these issues by checking out commit `bea6f40` which has the unittest showing the issue, while the following commit fixes the issue.  In the unittests I tests various scenarios where the fixed spans overlap the actual data in different ways and then verify that the alpha channels are calculated as expected.

maihde: Sorry, I see what you are asking about now, my bad!

Yes this was protective code to deal with incoming data that had negative values.  I'll write a unittest for this case to demonstrate the issue.  Sorry for both the initial confusion and not including a test the first time.

maihde: Okay, I have pushed a test up, but while implementing it really made me question what the correct behavior is (as I think are as well)

The [original code](https://github.com/holoviz/datashader/blob/7c34b668508deb04dffd7a3db076faa618138da6/datashader/transfer_functions/__init__.py#L358) prior to categorical aggregations supporting sum etc. had this code:

```
    if offset == 0:
        mask = mask | (total <= 0)
```

Since the second line had `total <= 0` I took that to mean that now that offset could be negative, the intention was to mask out negative values as well.  Anyways, I spent some time trying to think of how to get the desired behavior in all cases but haven't come across anything.  I'm happy to implement whatever you think best...or treat this question as a separate PR/feature.

philippjfr: Thanks for your efforts on this. I'll think about it a bit and get back to you. Would be great to get this into the imminent 0.11.0 release.

jbednar: Yes; this section of the codebase is a mix of code that was written assuming colorizing would only be done with counts (and thus always non-negative), and other parts that were updated.  So it's tough to reason about it and detect all the places that need to be updated so that they work properly with the new `by()` support.  I think the changes made in this PR so far definitely need to be made in some form, but they need to go further before they actually match the full range of what colorize can now be asked to render.

maihde: @jbednar @philippjfr as it relates to the 0.11.0 release.  This PR should still be identical behavior to the introduction of `by()` and `span` support and it fixes a fundamental issue with `span` that was in my PR #875.  In other words, I don't think this PR breaks anything that wasn't already broken and fixes something that should work.

jbednar: Ok, I'll merge this now, and then work on master to try to do something about negative values.

Issue #900, opened 2020-04-21
by() reduction doesn't combine with mean(), std() etc.


o-smirnov:
#### ALL software version info
DS master in a fresh virtualenv.

#### Description of expected behavior and the observed behavior

``by()`` doesn't seem to be able to combine itself with the more complex reductions like ``mean()``. I think it ignores ``_build_bases`` when constructing its ``append`` method, but I'm not adept enough with the code yet to fix it myself.

#### Complete, minimal, self-contained example code that reproduces the issue

```python
import numpy as np
import dask.dataframe
import datashader
import pandas as pd

if __name__ == '__main__':
    pf = pd.DataFrame(dict(a=np.arange(10), b=np.arange(10), c=np.arange(-5,5), cat=[0,0,0,1,1,1,2,2,2,3]))
    ddf = dask.dataframe.from_pandas(pf, npartitions=1)
    ddf = ddf.categorize('cat')
    print(ddf)

    canvas = datashader.Canvas(10, 10)

    raster = canvas.points(ddf, 'a', 'b', datashader.count_cat('cat'))
    print("count_cat ok")

    raster = canvas.points(ddf, 'a', 'b', datashader.mean('c'))
    print("mean ok")

    raster = canvas.points(ddf, 'a', 'b', datashader.by('cat', datashader.mean('c')))
    print("by(cat, mean(c)) ok")
```

#### Stack traceback and/or browser JavaScript console output

```
(sms) oms@tshikovski:~/projects/shadeMS$ python ./test-ds.py 
Dask DataFrame Structure:
                   a      b      c              cat
npartitions=1                                      
0              int64  int64  int64  category[known]
9                ...    ...    ...              ...
Dask Name: categorize_block, 2 tasks
count_cat ok
mean ok
Traceback (most recent call last):
  File "/home/oms/.venv/sms/lib/python3.6/site-packages/toolz/functoolz.py", line 456, in memof
    return cache[k]
KeyError: ((<datashader.reductions.by object at 0x7fcc800a9978>, dshape("""{
  a: int64,
  b: int64,
  c: int64,
  cat: categorical[[0, 1, 2, 3], type=int64, ordered=False]
  }"""), <datashader.glyphs.points.Point object at 0x7fcc800a9dd8>), frozenset({('cuda', False)}))

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "./test-ds.py", line 27, in <module>
[...]

PR #902, opened 2020-04-24
Add support for by with multiple base reductions

philippjfr:
This allows for reductions like `mean`, `var` and `std` to be used with `by`. This is achieved through three steps:

1. If a reduction has multiple bases (e.g. `mean` has `sum` and `count`) then create bases like `by('cat', base_reduction)`
2. Ensure that the `_append` method correctly handles the additional categorical dimension by wrapping the function
3. Ensure that during finalization the bases are combined e.g. `sum`/`count`, while also handling the categorical dimension

There are a few assumption made here:

* Any reduction with multiple bases that is to be supported with `by` must define a `_finalize` method
* Any reduction with an `append` that requires more than one value will need special handling (e.g. see handling of `m2` here)

Fixes https://github.com/holoviz/datashader/issues/900

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,11 +31,11 @@ stages:
   - name: docs
     if: tag =~ ^v(\d+|\.)*[^a-z]\d*$ OR tag = website
   - name: docs_dev
-    if: tag =~ ^v(\d+|\.)*[a-z]\d*$ OR tag = website_dev OR commit_message =~ /^.*(website_dev).*$/
+    if: tag =~ ^v(\d+|\.)+([a-z]|rc)\d+$ OR tag = website_dev OR commit_message =~ /^.*(website_dev).*$/
   - name: conda_dev_package
-    if: tag =~ ^v(\d+|\.)+[a-z]\d+$
+    if: tag =~ ^v(\d+|\.)+([a-z]|rc)\d+$
   - name: pip_dev_package
-    if: tag =~ ^v(\d+|\.)+[a-z]\d+$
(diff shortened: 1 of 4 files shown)


philippjfr:
Seemingly the code path when the JIT is disabled is not handling temp aggregates correctly, which is required for m2/var/std reductions. I don't know why that is but it works fine with the jit enabled. Maybe I should just skip the tests when JIT is disabled for now?

philippjfr:
Okay, hopefully fixed.

.travis.yml
```diff
@@ -31,11 +31,11 @@ stages:
...
     if: tag =~ ^v(\d+|\.)*[^a-z]\d*$ OR tag = website
   - name: docs_dev
-    if: tag =~ ^v(\d+|\.)*[a-z]\d*$ OR tag = website_dev OR commit_message =~ /^.*(website_dev).*$/
+    if: tag =~ ^v(\d+|\.)+([a-z]|rc)\d+$ OR tag = website_dev OR commit_message =~ /^.*(website_dev).*$/
   - name: conda_dev_package
-    if: tag =~ ^v(\d+|\.)+[a-z]\d+$
+    if: tag =~ ^v(\d+|\.)+([a-z]|rc)\d+$
   - name: pip_dev_package
-    if: tag =~ ^v(\d+|\.)+[a-z]\d+$
+    if: tag =~ ^v(\d+|\.)+([a-z]|rc)\d+$
```
philippjfr, line 38: Didn't mean to add this here, but it's not harmful so I'll leave it in (it allows rc tags).

PR #908, opened 2020-05-01
Fix _colorize alpha for case of no provided span

jbednar:
PR #875 introduced a regression, with categorical colormapping computing an incorrect alpha range. Specifically, in the previous release the output in the "Original Image" cell below used the full range of alpha values, resulting in bright colors:

image

while with current Datashader master the alpha range is very small:

image

With git bisect it was clear that these changes were from the changes to _colorize() in #875 to allow specifying a span.  On investigation, the issue was that previously when no span was set, the range of alpha values was set based on the interpolated (scaled) data, while after #875 the range was set using unscaled data (which in this example is a much smaller range than the 0 to 255 scaled range).

Because there have been changes to this code since #875 (and specifically in #896), it was difficult to see how to restore the original behavior, but I've done my best.  The results look appropriate to me with and without a specified span on these examples now.

image

But some part of the changes in #896 may have been lost, e.g. whatever `masked_clip_2d(total, mask, *span)` was doing here. Plus I have not yet tried to address the clearly inappropriate behavior for negative values, which should _not_  be masked off as they are now.

### DIFF
--- a/datashader/transfer_functions/__init__.py
+++ b/datashader/transfer_functions/__init__.py
@@ -357,7 +357,7 @@ def _colorize(agg, color_key, how, span, min_alpha, name):
     # min/max of the data
     if span is None:
         # Currently masks out zero or negative values, but will need fixing         
-        offset = total.min()
+        offset = np.nanmin(total)
         if offset <= 0:
             mask = mask | (total <= 0)
             # If at least one element is not masked, use the minimum as the offset
@@ -377,9 +377,8 @@ def _colorize(agg, color_key, how, span, min_alpha, name):
         offset = np.array(span, dtype=data.dtype)[0]
         if offset <= 0:
             mask = mask | (total <= 0)
-            offset = total[total > 0].min()
-        a_ = _normalize_interpolate_how(how)(total - offset, mask)
         span = _normalize_interpolate_how(how)([0, span[1] - span[0]], 0)
+        a_ = _normalize_interpolate_how(how)(total - offset, mask)
 
     # Interpolate the alpha values
     a = interp(a_, array(span),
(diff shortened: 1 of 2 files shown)


datashader/transfer_functions/__init__.py
```diff
@@ -356,18 +356,16 @@ def _colorize(agg, color_key, how, span, min_alpha, name):
...
-        span = [np.nanmin(total).item(), np.nanmax(total).item()]
-        offset = span[0]
-        if how != 'eq_hist':
-            span = _normalize_interpolate_how(how)([0, span[1] - span[0]], 0)
-        # If there are elements that are zero or lower, they should be fully
-        # transparent and masked out
-        if offset <= 0:
+        # Currently masks out zero or negative values, but will need fixing         
+        offset = np.nanmin(total)
+        if offset <= 0 and total.dtype.kind == 'u':
```
jbednar, line 361:
```suggestion
        if offset == 0 and total.dtype.kind == 'u':
```

(a unit can't be less than zero, right?)

datashader/transfer_functions/__init__.py
```diff
@@ -356,18 +356,16 @@ def _colorize(agg, color_key, how, span, min_alpha, name):
...
     # min/max of the data
     if span is None:
-        span = [np.nanmin(total).item(), np.nanmax(total).item()]
-        offset = span[0]
-        if how != 'eq_hist':
-            span = _normalize_interpolate_how(how)([0, span[1] - span[0]], 0)
-        # If there are elements that are zero or lower, they should be fully
-        # transparent and masked out
-        if offset <= 0:
+        # Currently masks out zero or negative values, but will need fixing
```
jbednar, line 359:
```suggestion
```

PR #909, opened 2020-05-02
Handle negative values in categorical colorize

philippjfr:
I've been testing with the following code:

```python
import numpy as np
import holoviews as hv
hv.extension('bokeh')
coords = [np.array([0, 1]), np.array([2, 5])]
dims = ['y_axis', 'x_axis']
cat_agg = tf.Image(np.array([[(0, -30, 0), (-18, 0, -18)],
                                [(-2, -2, -2), (-18, 0, 0)]]),
                        coords=(coords + [['a', 'b', 'c']]),
                        dims=(dims + ['cats']))
colors = [(255, 0, 0), '#0000FF', 'orange']
img = tf.shade(cat_agg, color_key=colors, how='linear', span=(-36, 0), min_alpha=20)
img2 = tf.shade(cat_agg, color_key=colors, how='linear')
hv.RGB(img.data.view('uint8').reshape((2, 2, 4))[::-1, :]) +\
hv.RGB(img2.data.view('uint8').reshape((2, 2, 4))[::-1, :])
```

### DIFF
--- a/datashader/transfer_functions/__init__.py
+++ b/datashader/transfer_functions/__init__.py
@@ -163,7 +163,9 @@ def eq_hist(data, mask=None, nbins=256*256):
 
     data2 = data if mask is None else data[~mask]
     if data2.dtype == bool or np.issubdtype(data2.dtype, np.integer):
-        hist = np.bincount(data2.astype('i8').ravel())
+        if data2.dtype.kind == 'u':
+             data2 = data2.astype('i8')
+        hist = np.bincount(data2.ravel())
         bin_centers = np.arange(len(hist))
         idx = int(np.nonzero(hist)[0][0])
         hist, bin_centers = hist[idx:], bin_centers[idx:]
--- a/datashader/transfer_functions/__init__.py
+++ b/datashader/transfer_functions/__init__.py
@@ -334,8 +334,8 @@ def _colorize(agg, color_key, how, span, min_alpha, name):
             # otherwise the offset remains at zero
             if not np.all(mask):
                 offset = total[total > 0].min()
-        masked_total = np.where(~mask, total, np.nan)
-        a_scaled = _normalize_interpolate_how(how)(masked_total - offset, mask)
+            total = np.where(~mask, total, np.nan)
+        a_scaled = _normalize_interpolate_how(how)(total - offset, mask)
         norm_span = [np.nanmin(a_scaled).item(), np.nanmax(a_scaled).item()]
     else:
         if how == 'eq_hist':
@@ -346,9 +346,10 @@ def _colorize(agg, color_key, how, span, min_alpha, name):
         # i.e. a 0 will be fully transparent, but any non-zero number will
         # be clipped to the span range and have min-alpha applied
         offset = np.array(span, dtype=data.dtype)[0]
-        if offset == 0  and total.dtype.kind == 'u':
+        if np.nanmin(total) == 0 and total.dtype.kind == 'u':
             mask = mask | (total <= 0)
-        masked_clip_2d(data, mask, *span)
+            total = np.where(~mask, total, np.nan)
+        masked_clip_2d(total, mask, *span)
         a_scaled = _normalize_interpolate_how(how)(total - offset, mask)
         norm_span = _normalize_interpolate_how(how)([0, span[1] - span[0]], 0)
 
--- a/datashader/transfer_functions/__init__.py
+++ b/datashader/transfer_functions/__init__.py
@@ -358,7 +358,7 @@ def _colorize(agg, color_key, how, span, min_alpha, name):
     if span is None:
         # Currently masks out zero or negative values, but will need fixing         
         offset = np.nanmin(total)
-        if offset <= 0:
+        if offset <= 0 and total.dtype.kind == 'u':
             mask = mask | (total <= 0)
             # If at least one element is not masked, use the minimum as the offset
             # otherwise the offset remains at zero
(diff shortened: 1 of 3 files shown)


jbednar:
This is addressing a remaining issue from https://github.com/holoviz/datashader/pull/908 , which is that negative data values were being cropped.  The colorize code was originally written just for count_cat, which (being counts) always returns non-negative values, and so it was safe to use zero as the "missing data" value used as a mask.  Now that other aggregates can be used, this PR changes the aggregate datatype to an unsigned int when working with counts, allowing us to preserve this masking behavior for counts while allowing zero and negative values for non-count data.

datashader/transfer_functions/__init__.py
```diff
@@ -322,9 +289,10 @@ def _interpolate(agg, cmap, how, alpha, span, min_alpha, name):
 
 def _colorize(agg, color_key, how, span, min_alpha, name):
     if cupy and isinstance(agg.data, cupy.ndarray):
-        from ._cuda_utils import interp
+        from ._cuda_utils import interp, masked_clip_2d 
         array = cupy.array
     else:
+        from ._cpu_utils import masked_clip_2delse:
```
jbednar, line 295:
```suggestion
        from ._cpu_utils import masked_clip_2d:
```

datashader/transfer_functions/__init__.py
```diff
@@ -358,7 +326,7 @@ def _colorize(agg, color_key, how, span, min_alpha, name):
     if span is None:
         # Currently masks out zero or negative values, but will need fixing         
         offset = np.nanmin(total)
-        if offset <= 0:
+        if offset <= 0 and total.dtype.kind == 'u':
```
jbednar, line 329:
```suggestion
        if offset == 0 and total.dtype.kind == 'u':
```

datashader/transfer_functions/__init__.py
```diff
@@ -375,8 +343,9 @@ def _colorize(agg, color_key, how, span, min_alpha, name):
         # i.e. a 0 will be fully transparent, but any non-zero number will
         # be clipped to the span range and have min-alpha applied
         offset = np.array(span, dtype=data.dtype)[0]
-        if offset <= 0:
+        if offset <= 0  and total.dtype.kind == 'u':
```
jbednar, line 346:
```suggestion
        if offset == 0  and total.dtype.kind == 'u':
```

datashader/transfer_functions/__init__.py
```diff
@@ -163,7 +163,7 @@ def eq_hist(data, mask=None, nbins=256*256):
 
     data2 = data if mask is None else data[~mask]
     if data2.dtype == bool or np.issubdtype(data2.dtype, np.integer):
-        hist = np.bincount(data2.ravel())
+        hist = np.bincount(data2.astype('i8').ravel())
```
philippjfr, line 166: Might be better:

```suggestion
        if data2.dtype.kind == 'u':
             data2 = data2.astype('i8')
        hist = np.bincount(data2.ravel())
```

datashader/transfer_functions/__init__.py
```diff
@@ -375,8 +346,10 @@ def _colorize(agg, color_key, how, span, min_alpha, name):
         # i.e. a 0 will be fully transparent, but any non-zero number will
         # be clipped to the span range and have min-alpha applied
         offset = np.array(span, dtype=data.dtype)[0]
-        if offset <= 0:
+        if np.nanmin(total) == 0 and total.dtype.kind == 'u':
```
philippjfr, line 349: If the data is of `uint` type we should **always** mask out zeros even with a specified span, otherwise zero values end up being black in the aggregate which is wrong.

datashader/transfer_functions/__init__.py
```diff
@@ -375,8 +346,10 @@ def _colorize(agg, color_key, how, span, min_alpha, name):
...
             mask = mask | (total <= 0)
+            total = np.where(~mask, total, np.nan)
+        masked_clip_2d(total, mask, *span)
```
philippjfr, line 352: Clip values outside the provided span ensuring their alpha values do not over/underflow during interpolation.

datashader/transfer_functions/__init__.py
```diff
@@ -358,12 +328,13 @@ def _colorize(agg, color_key, how, span, min_alpha, name):
...
         # Currently masks out zero or negative values, but will need fixing         
         offset = np.nanmin(total)
-        if offset <= 0:
+        if offset == 0 and total.dtype.kind == 'u':
             mask = mask | (total <= 0)
             # If at least one element is not masked, use the minimum as the offset
             # otherwise the offset remains at zero
             if not np.all(mask):
                 offset = total[total > 0].min()
+            total = np.where(~mask, total, np.nan)
```
philippjfr, line 337: This should ensure that zero values are not given a minimum alpha.

philippjfr:
I've commented with some of the reasoning/assumptions I'm making. Reviewing the rest of the code I've got some more comments:

```python
r[mask] = g[mask] = b[mask] = 255
```

This isn't clear to me, why set masked out values to white? They already have zero alpha, we are destroying information here.

In reviewing `_interpolate` (i.e. non-categorical colorize) I also started thinking about these lines:

```python
        if np.issubdtype(data.dtype, np.integer):
            mask = data == 0
```

This does not appear valid for `sum` or other non-count aggregates to me, why should zero values be masked out in this case? Now that we use uints for count aggregates I think zero should be handled just like any other value for regular integer types.

philippjfr:
Okay, I'll merge and tag another dev release for now.

Issue #899, opened 2020-04-20
Problems with _colorize()


o-smirnov:
### ALL software version info

0.10.0 release, but same problem on master.

#### Description of expected behavior and the observed behavior

I'm using an aggregator similar to ``color_cat()`` to colorize data. The resulting raster comes out very under-saturated, as it seems to have only two values in the alpha channel: 0 and 40.

Stepping through with the debugger, I can see the problem is here: https://github.com/holoviz/datashader/blob/master/datashader/transfer_functions/__init__.py#L375

I haven't got an explicit span set, so ``span`` is equal to [0, max(total)] at this point (the latter being some very large number), while the ``a`` array is normalized to [0,1]. Calling interp with a huge span and such a small ``a`` results in all non-zero values of a pinned to min_alpha=40.

Comparing to the implementation of ``_interpolate``, span should be set to min(a),max(a) at this point for correct results...

o-smirnov:
Looking at the code, I think there's actually a more insidious problem if ``_colorize`` is used together with a  ``by()`` reduction. Let's say I do ``by('category_column', mean('value_column'))`` (I'd need to get #900 to work first, but for the sake of argument let's say it does), and the mean of the value column is negative in some categories. I think this bit of code implicitly assumes that the result of the reduction is positive (no doubt because it started life when the only reduction was "count"):

```python
    total = data.sum(axis=2)
    # zero-count pixels will be 0/0, but it's safe to ignore that when dividing
    with np.errstate(divide='ignore', invalid='ignore'):
        r = (data.dot(rs)/total).astype(np.uint8)
        g = (data.dot(gs)/total).astype(np.uint8)
        b = (data.dot(bs)/total).astype(np.uint8)
```

...and also the same assumption down below, where ``total`` is masked on >0. So it's all going to break down with reductions that can return negative values.

I guess the solution is to shift and rescale the data into a positive interval first up. In fact, as a workaround, I can do this to the raster before I even call ``shade``.

o-smirnov:
Looking at the code, I think there's actually a more insidious problem if ``_colorize`` is used together with a  ``by()`` reduction. Let's say I do ``by('category_column', mean('value_column'))`` (I'd need to get #900 to work first, but for the sake of argument let's say it does), and the mean of the value column is negative in some categories. I think this bit of code implicitly assumes that the result of the reduction is positive (no doubt because it started life when the only reduction was "count"):

```python
    total = data.sum(axis=2)
[...]

PR #910, opened 2020-05-04
Additional fixes for colorize span

jbednar:
So far just comments, but I think there also should be handling for min_alpha in the case where the span includes no values.

### DIFF
--- a/datashader/geo.py
+++ b/datashader/geo.py
@@ -24,7 +24,8 @@ __all__ = ['mean', 'binary', 'slope', 'aspect', 'ndvi', 'hillshade', 'generate_t
 
 warnings.warn(
     "The datashader.geo module is deprecated as of version 0.11.0. "
-    "Its contents have migrated to the xarray_spatial library.",
+    "Its contents have migrated to the xarray_spatial library, ",
+    "github.com/makepath/xarray-spatial.",
     VisibleDeprecationWarning
 )
 
--- a/datashader/geo.py
+++ b/datashader/geo.py
@@ -21,10 +21,9 @@ from xarray import DataArray
 __all__ = ['mean', 'binary', 'slope', 'aspect', 'ndvi', 'hillshade', 'generate_terrain',
            'lnglat_to_meters']
 
-
 warnings.warn(
     "The datashader.geo module is deprecated as of version 0.11.0. "
-    "Its contents have migrated to the xarray_spatial library, ",
+    "Its contents have migrated to the xarray_spatial library, "
     "github.com/makepath/xarray-spatial.",
     VisibleDeprecationWarning
 )
--- a/datashader/reductions.py
+++ b/datashader/reductions.py
@@ -410,10 +410,7 @@ class sum(FloatingReduction):
 
     @staticmethod
     def _combine(aggs):
-        missing_vals = np.isnan(aggs)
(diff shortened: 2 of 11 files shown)


philippjfr:
>I think there also should be handling for min_alpha in the case where the span includes no values.

I think I disagree here, the current behavior is consistent with the way it has always worked in the non-categorical case and how other plotting libraries treat color limits. Anything below the lower **should** be assigned the `min_alpha` value.

jbednar:
> I think I disagree here, the current behavior is consistent with the way it has always worked in the non-categorical case and how other plotting libraries treat color limits. Anything below the lower bound should be assigned the min_alpha value

I've added examples of using min_alpha together with span, and I _think_ you're right; it seems to behave consistently. It would be great if you could look over the new examples (in the new section "2_Pipeline.ipynb#Controlling-ranges-for-colormapping") and see if you think the behavior is as you'd expect.

There are also a few open issues remaining from recent PRs/issues:

- [ ] @philippjfr: _interpolate "if np.issubdtype(data.dtype, np.integer): mask = data == 0" -- This does not appear valid for sum or other non-count aggregates to me, why should zero values be masked out in this case? Now that we use uints for count aggregates I think zero should be handled just like any other value for regular integer types.
- @jbednar: I think I agree, though note that we will then no longer be able to mask out pixels with no data in them, for integer aggregates other than count.  I do think that's safer, on balance, though. I tried changing that line to test "if data.dtype.kind == 'u':" instead, but saw no difference.

- [ ] Need examples of non-count `by` aggregates

jbednar:
Fixes #910.

jbednar:
Ok, Philipp and I had a productive debugging session where we really tried to pin down where things were behaving differently between _interpolate and _colorize, between sum and mean, and between sum/mean and count.  We found a variety of issues due to the code never previously having supported floats or nans, and we think we managed to address all those.  The behavior now looks good for aggregating sums and means of positive values, which is now illustrated in new examples in 2_Pipeline.ipynb.  The behavior for aggregating non-categorical negative values also looks good, again illustrated with new examples showing what I consider to be appropriate ways to work with aggregations of positive and negative values.

- [ ] The last remaining known issue with these changes is to validate or improve the behavior for aggregating positive and negative values categorically.  I added a new example that illustrates what it currently does, but I'll have to think about what it _should_ do in that case, and then we can finally merge this.
- [ ] Well, after adding some tests corresponding to the new cases added to 2_Pipeline.ipynb.

jbednar:
Discussion for the zero and negative case is at https://github.com/holoviz/datashader/issues/899#issuecomment-628386981; short answer is yes, it is broken.

jbednar:
I've now changed _colorize to subtract the minimum value from the data by default before calculating colors, which now lets it work with negative values and seems to have essentially no impact on the typical cases that previously worked, i.e. counts, since those nearly always have a pixel with 0 anyway, so that using the minimum is the same as the previous default of using 0.

I also added special handling for computing a color for pixels that are zero (after subtracting the baseline), which previously was causing some pixels to end up black.  Now they instead are colored with the average color of the categories present (albeit at zero strength) in that pixel, which allows us to avoid ever having undefined colors.

I added new examples to 2_Pipeline.ipynb and documented the `color_baseline` argument, so I think it's ready to go once the tests pass (and new tests are added!).

PR #913, opened 2020-05-06
Fixed indexing errors in QuadMesh in python2

philippjfr:
In Python2 QuadMesh aggregation was erroring because the src indexes were floats not integers.

See https://travis-ci.org/github/holoviz/holoviews/jobs/683632264 for the error.


jbednar:
Good catch, thanks!

PR #916, opened 2020-05-13
Implementation of Xiaolin Wu anti-aliased line drawing

esc:
Partially implements https://github.com/holoviz/datashader/pull/916 .


datashader/core.py
```diff
@@ -257,6 +258,8 @@ def line(self, source, x=None, y=None, agg=None, axis=0, geometry=None):
         geometry : str
             Column name of a LinesArray of the coordinates of each line. If provided,
             the x and y arguments may not also be provided.
+        antialias : bool
+            If True, do anti-aliasing.
```
jbednar, line 262:
```suggestion
            If True, draw anti-aliased lines, distributing the aggregate value across neighboring pixels to more closely approximate the line shape. If False, each position on the line affects only a single pixel, resulting in line shapes that are blocky but easier to reason about.
```

datashader/glyphs/line.py
```diff
@@ -547,19 +561,131 @@ def map_onto_pixel(sx, tx, sy, ty, xmin, xmax, ymin, ymax, x, y):
...
+    xend = _draw_endpoint(agg, steep,   (x1, y1), grad)
+    for x in range(xstart, xend):
+        y = int(intery)
+        _draw_pixel(agg, _flipxy(x, y, steep), _myrfpart(intery))
+        _draw_pixel(agg, _flipxy(x, y+1, steep), _myfpart(intery))
+        intery += grad
+
+@ngjit
+def _draw_pixel(agg, px, value):
+    """ Xiaolin Wu utility. """
```
jbednar, line 588:
```suggestion
    """ Xiaolin Wu utility. """
```

It's great for methods to have docstrings, but only if the docstring conveys something beyond what's already clear from the method and class names.

datashader/core.py
esc: Sure!

datashader/glyphs/line.py
esc: Yes, they clearly can't stay like this. I could either remove them or flesh them out a bit more. What is your preference?

esc:
So this is slowly maturing and I wanted to share the script I have been using to test this code:

```python
import xarray as xr
import datashader as ds
import pandas as pd
import numpy as np
from matplotlib.cm import binary

cmap01 =([tuple(v*255. for v in tuple(rgb)[:3])
          for rgb in binary(np.linspace(0, 1))])
cmap02 =([tuple(v*255. for v in tuple(rgb)[:3])
          for rgb in binary(np.linspace(0, 1))])[::-1]
cmap03 = [(x, x, x) for x in np.linspace(0,255, 2**16-1)]
cmap04 = ['gainsboro', 'lightgrey', 'silver', 'darkgrey', 'grey', 'dimgrey']


regaular_cvs = ds.Canvas(plot_width=500, plot_height=500,
                        x_range=(0, 499), y_range=(0, 499))
reduced_cvs = ds.Canvas(plot_width=200, plot_height=200,
                        x_range=(149, 349), y_range=(149, 349))


def draw_line(cvs, p1, p2, antialias):
    xs, ys = np.array([p1[0], p2[0]]), np.array([p1[1], p2[1]])
    points = pd.DataFrame({'x': xs, 'y': ys, 'val': 5.0})
    return cvs.line(points, 'x', 'y', agg=ds.reductions.max("val"),
                    antialias=antialias)


def draw_lines(cvs, points, antialias):
    aggs = []
    for ((x1, y1), (x2, y2)) in points:
        aggs.append(draw_line(cvs, (x1, y1), (x2, y2), antialias))
    return xr.concat(aggs, 'stack').sum(dim='stack')


def draw_multi_segment_line(cvs, points, antialias):
    x, y = [], []
    for (x1, y1) in points:
        x.append(x1)
        y.append(y1)
    xs, ys = np.array(x), np.array(y)
    points = pd.DataFrame({'x': xs, 'y': ys, 'val': 5.0})
    agg = cvs.line(points, 'x', 'y', agg=ds.reductions.max("val"),
                    antialias=antialias)
    # This is required for the line to render properly
    return xr.concat([agg], 'stack').sum(dim='stack')


def draw_image(aggregators, cmap=cmap01, filename='test.png'):
    img = ds.transfer_functions.shade(aggregators, cmap=cmap)
    img = ds.transfer_functions.set_background(img, '#ffffff')
    print("Saving: ", filename)
    img.to_pil().save(filename)


def test_001():
    points = []
    for a in range(10, 550, 60):
        points.append(((10, 10), (490, a)))
    return points, "test_001"

def test_002():
    points = []
    for a in range(10, 550, 60):
        points.append(((10, 10), (a, 490)))
    return points, "test_002"

def test_003():
    points = []
    for a in range(10, 550, 60):
        points.append(((490, 490), (10, a)))
    return points, "test_003"

def test_004():
    points = []
    for a in range(10, 550, 60):
        points.append(((490, 490), (a, 10)))
    return points, "test_004"

def test_005():
    points = [
        ((10, 10),  (10, 490)),
        ((10, 490), (20, 10)),
        ((20, 10),  (20, 490)),
        ((20, 490), (40, 10)),
        ((40, 10),  (40, 490)),
        ((40, 490), (80, 10)),
        ((80, 10),  (80, 490)),
        ((80, 490), (160, 10)),
        ((160, 10), (160, 490)),
        ((160, 490),(320, 10)),
        ((320, 10), (320, 490)),
        ((320, 490),(490, 10)),
        ((490, 10), (490, 490)),
    ]
    return points, "test_005"

def test_006():
    points = [
        (10, 10),
        (10, 490),
        (20, 10),
        (20, 490),
        (40, 10),
        (40, 490),
        (80, 10),
        (80, 490),
        (160, 10),
        (160, 490),
        (320, 10),
        (320, 490),
        (490, 10),
        (490, 490),
    ]
    return points, "test_006"


antialias_options = ((True, "antialias"), (False, "noaa"))
canvas_options = ((regaular_cvs, "normal"), (reduced_cvs, "clipped"))

for antialias, aa_descriptor  in antialias_options:
    for canvas, canvas_descriptor in canvas_options:
        for func in (test_001, test_002, test_003, test_004, test_005):
            points, name = func()
            aggregators = draw_lines(canvas, points, antialias)
            draw_image(aggregators,
                       filename=f"{name}_{aa_descriptor}_{canvas_descriptor}.png")

        for func in (test_006, ):
            points, name = func()
            aggregator = draw_multi_segment_line(canvas, points, antialias)
            draw_image(aggregator,
                    filename=f"{name}_{aa_descriptor}_{canvas_descriptor}.png")

```

esc:
The test basically produces a set of test images from a matrix of combinations.

esc:
I'm going to rebase this now.

setup.py
```diff
@@ -44,7 +44,8 @@
         'nbsmoke[all] >=0.4.0',
         'fastparquet >=0.1.6',  # optional dependency
         'holoviews >=1.10.0',
-        'pyarrow'
+        'pyarrow',
+        'netcdf'
```
jbednar, line 48:
```suggestion
        'netcdf4'
```

jbednar: (May fix the tests...)

esc: I'll give it a try

esc:
It seems like the tests are now failing with legitimate errors, I'll investigate.

esc:
This PR will likely need a more recent Numba. I think it will not pass, until 0.51.0 becomes available from the defaults channel on anaconda.org

esc:
I used an f-string in my antialias PR, now the 2.7 build is unhappy, does datashader still supp. 2.7? I think Datashader should probably orient itself along Numpy.

jlstevens:
Tests now passing other than the py2 tests which won't work with the required numba version for antialiasing.

jlstevens:
I have checked out this fork and can confirm everything is working well for the examples generated by the test script. 

One thing I noticed is that the test data is rather large 14MB and I think it would be fine to regenerate the data with a 10x reduction in size for the canvas in each direction (i.e using 50x50 and 20x20 arrays). Looking at the PNG images clearly shows the effect of antialiasing and I think this would be a more suitable test data set (~400KB).

Overall, the main thing to discuss before merging this PR is the required numba version (0.51.0, released on August 12) which is very recent, not yet available on the defaults channel and would require us to drop python 2 (this PR fails to compile for numba versions prior to 0.49 and 0.47 is the last version that supports python 2).

philippjfr:
My suggestion is to keep the old code for the line rendering algorithm and conditionally use that for older numba releases. We can then drop py2 and this old code branch in the same release. Also not comfortable dropping support for old numba at this point since 0.48 and 0.51.0 are the only releases that actually work properly and 0.51.2 is not yet released.

esc:
@jlstevens @philippjfr thanks for the feedback, I'll patch it up accordingly soon.

jlstevens:
I've been trying to see how antialising affects the datashader [Trimesh example](https://datashader.org/user_guide/Trimesh.html) but without success:

image

image

As printed, the top version has antialiasing on (i.e `cvs.line` with `antialising=True`) and the bottom has it off (i.e `cvs.line` with `antialising=False`).  Unfortunately, I can't see the difference as both seem to have about the same number of jaggies!

I tested that I am using this PR by running the script posted above which does show a difference in that case (antialiased on the left in case it wasn't obvious):

image

I wonder what is different about the use of `cvs.line` that makes the benefit of antialiasing obvious in the test case but not the trimesh case...

philippjfr:
Is antialiasing implemented for all line rendering modes such as `LineAxis1` (which is what we use for rendering trimesh vertices in HoloViews at least)?

jlstevens:
After a little bit of discussion, this is my understanding about why the trimesh isn't being antialiased:

1. `LineAxis1` *does* have an antialiasing implementation but only the `sum` and `max` aggregators are supported.
2. For this example, only the `count` and `any` aggregators can be used which means antialiasing cannot be applied (as there is no value column).

Unfortunately, I think this example represents a common use case where people use datashader to visualize dense trimeshes (which won't necessarily have a value column).

esc:
I updated the sizes of the test-data and replaced the large files in the git history using rebase, hence the force push.

The file sizes are now:

```
💣 zsh» ls -lah datashader/tests/data/*.{png,nc}                                                                                                                   :(
-rw------- 1 vhaenel staff 9.2K Sep  4 13:54 datashader/tests/data/test_001_antialias_clipped.nc
-rw------- 1 vhaenel staff  432 Sep  4 13:54 datashader/tests/data/test_001_antialias_clipped.png
-rw------- 1 vhaenel staff  19K Sep  4 13:54 datashader/tests/data/test_001_antialias_normal.nc
-rw------- 1 vhaenel staff 1.2K Sep  4 13:54 datashader/tests/data/test_001_antialias_normal.png
-rw------- 1 vhaenel staff 9.2K Sep  4 13:54 datashader/tests/data/test_001_noaa_clipped.nc
-rw------- 1 vhaenel staff  203 Sep  4 13:54 datashader/tests/data/test_001_noaa_clipped.png
-rw------- 1 vhaenel staff  19K Sep  4 13:54 datashader/tests/data/test_001_noaa_normal.nc
-rw------- 1 vhaenel staff  560 Sep  4 13:54 datashader/tests/data/test_001_noaa_normal.png
-rw------- 1 vhaenel staff 9.2K Sep  4 13:54 datashader/tests/data/test_002_antialias_clipped.nc
-rw------- 1 vhaenel staff  412 Sep  4 13:54 datashader/tests/data/test_002_antialias_clipped.png
-rw------- 1 vhaenel staff  19K Sep  4 13:54 datashader/tests/data/test_002_antialias_normal.nc
-rw------- 1 vhaenel staff 1.1K Sep  4 13:54 datashader/tests/data/test_002_antialias_normal.png
-rw------- 1 vhaenel staff 9.2K Sep  4 13:54 datashader/tests/data/test_002_noaa_clipped.nc
-rw------- 1 vhaenel staff  196 Sep  4 13:54 datashader/tests/data/test_002_noaa_clipped.png
-rw------- 1 vhaenel staff  19K Sep  4 13:54 datashader/tests/data/test_002_noaa_normal.nc
-rw------- 1 vhaenel staff  479 Sep  4 13:54 datashader/tests/data/test_002_noaa_normal.png
-rw------- 1 vhaenel staff 9.2K Sep  4 13:54 datashader/tests/data/test_003_antialias_clipped.nc
-rw------- 1 vhaenel staff  487 Sep  4 13:54 datashader/tests/data/test_003_antialias_clipped.png
-rw------- 1 vhaenel staff  19K Sep  4 13:54 datashader/tests/data/test_003_antialias_normal.nc
-rw------- 1 vhaenel staff 1.2K Sep  4 13:54 datashader/tests/data/test_003_antialias_normal.png
-rw------- 1 vhaenel staff 9.2K Sep  4 13:54 datashader/tests/data/test_003_noaa_clipped.nc
-rw------- 1 vhaenel staff  202 Sep  4 13:54 datashader/tests/data/test_003_noaa_clipped.png
-rw------- 1 vhaenel staff  19K Sep  4 13:54 datashader/tests/data/test_003_noaa_normal.nc
-rw------- 1 vhaenel staff  593 Sep  4 13:54 datashader/tests/data/test_003_noaa_normal.png
-rw------- 1 vhaenel staff 9.2K Sep  4 13:54 datashader/tests/data/test_004_antialias_clipped.nc
-rw------- 1 vhaenel staff  506 Sep  4 13:54 datashader/tests/data/test_004_antialias_clipped.png
-rw------- 1 vhaenel staff  19K Sep  4 13:54 datashader/tests/data/test_004_antialias_normal.nc
-rw------- 1 vhaenel staff 1.1K Sep  4 13:54 datashader/tests/data/test_004_antialias_normal.png
-rw------- 1 vhaenel staff 9.2K Sep  4 13:54 datashader/tests/data/test_004_noaa_clipped.nc
-rw------- 1 vhaenel staff  181 Sep  4 13:54 datashader/tests/data/test_004_noaa_clipped.png
-rw------- 1 vhaenel staff  19K Sep  4 13:54 datashader/tests/data/test_004_noaa_normal.nc
-rw------- 1 vhaenel staff  454 Sep  4 13:54 datashader/tests/data/test_004_noaa_normal.png
-rw------- 1 vhaenel staff 9.2K Sep  4 13:54 datashader/tests/data/test_005_antialias_clipped.nc
-rw------- 1 vhaenel staff  285 Sep  4 13:54 datashader/tests/data/test_005_antialias_clipped.png
-rw------- 1 vhaenel staff  19K Sep  4 13:54 datashader/tests/data/test_005_antialias_normal.nc
-rw------- 1 vhaenel staff 1.4K Sep  4 13:54 datashader/tests/data/test_005_antialias_normal.png
-rw------- 1 vhaenel staff 9.2K Sep  4 13:54 datashader/tests/data/test_005_noaa_clipped.nc
-rw------- 1 vhaenel staff  139 Sep  4 13:54 datashader/tests/data/test_005_noaa_clipped.png
-rw------- 1 vhaenel staff  19K Sep  4 13:54 datashader/tests/data/test_005_noaa_normal.nc
-rw------- 1 vhaenel staff  368 Sep  4 13:54 datashader/tests/data/test_005_noaa_normal.png
-rw------- 1 vhaenel staff 9.2K Sep  4 13:54 datashader/tests/data/test_006_antialias_clipped.nc
-rw------- 1 vhaenel staff  285 Sep  4 13:54 datashader/tests/data/test_006_antialias_clipped.png
-rw------- 1 vhaenel staff  19K Sep  4 13:54 datashader/tests/data/test_006_antialias_normal.nc
-rw------- 1 vhaenel staff 1.3K Sep  4 13:54 datashader/tests/data/test_006_antialias_normal.png
-rw------- 1 vhaenel staff 9.2K Sep  4 13:54 datashader/tests/data/test_006_noaa_clipped.nc
-rw------- 1 vhaenel staff  139 Sep  4 13:54 datashader/tests/data/test_006_noaa_clipped.png
-rw------- 1 vhaenel staff  19K Sep  4 13:54 datashader/tests/data/test_006_noaa_normal.nc
-rw------- 1 vhaenel staff  334 Sep  4 13:54 datashader/tests/data/test_006_noaa_normal.png
```

esc:
The total size of the test data for this PR is now at `884K`.

datashader/glyphs/line.py
jbednar: More info would be great.

esc:
@jlstevens this is ready for the next round of reviews. I fixed an off-by-on-error and fixed the code to work/compile with older Numba versions. So the tests now pass on Numba 0.48.0 but w/o anti-aliasing and pass, with anti-aliasing, on Numba 0.51.2. So it's probably worth checking both those Numba versions manually during testing.

esc:
@jbednar it might do, yes. I think the only way to circumvent that, I can think of right now, is to always use a 'safe draw pixel'.

esc:
@jbednar it might be possible to craft a test case for this.

jbednar:
Maybe do unsafe_draw_pixel for `start:end-1`, then do safe_draw_pixel for `end-1:end`?

esc:
> Maybe do unsafe_draw_pixel for `start:end-1`, then do safe_draw_pixel for `end-1:end`?

The y-value isn't looped over but determined from the x-value -- so that proposed solution is not feasible, I think.

jlstevens:
Travis tests are passing and I've checked that 0.47 seems to be working fine as long as you don't try to use antialiasing.

esc:
Commit https://github.com/holoviz/datashader/pull/916/commits/c3ac6191c29199ec5a16dd93390c1ad875d3e7ad adds a test case for lines on the edges:

Anti-aliased:
 
test_007_antialias_normal

Non anti-aliased:

test_007_noaa_normal

As you can see, the lines at the edges will be drawn using the "inner two pixels" as it were. So it will always use pixels with coordinates inside the canvas. This is especially well visible in the anti-aliased example above.

jlstevens:
Looks good to me! @jbednar ?

jbednar:
@esc , I'm a bit confused. The anti-aliased line looks great in the above screenshot, and indeed I can see that when connecting a line segment between (x1,y1) and (x2,y2), where x1,y2,y1,y2 are chosen as the centers of pixels, the antialiased bits are contained within the inclusive ranges (x1:x2, y1:y2). I.e., there shouldn't be any out-of-bounds drawing in that case.  But wasn't 
https://github.com/holoviz/datashader/pull/916/commits/7729722caa89bdf1715bad0fe8af3875675eb653 designed to catch such out-of-bounds drawing, indicating that it was still happening?  Seems like it can't be both ways, i.e. that edges will always be drawn with the "inner two pixels" and that https://github.com/holoviz/datashader/pull/916/commits/7729722caa89bdf1715bad0fe8af3875675eb653 is needed.

To me it seems like there will always be the potential for out of bounds drawing when antialiasing is used.  E.g. if the bounds covers the inclusive range (-1:1,-1:1), what happens for a horizontal line segment starting at (-1,0.9999995) and ending at (1,0.9999999)? Towards the end point, isn't most of the weight of those last few pixels going to be out of bounds, and thus won't the antialiasing code try to draw it there unless explicitly cropped?

esc:
Following up on this: the anti-alias implementation uses the existing `liang-barsky` algorithm to clip the line inside the viewport/canvas. So, I believe that cropping is actually a built-in feature of the current framework and thus is (currently) non-optional. So, none of the line drawing algorithms will try to draw out of bounds, except the anti-aliasing one if and only if it is a straight line, and in this case, only empty/white pixels will be drawn out of bounds and this is the specialized condition that 7729722 addresses.

stuartarchibald:
@esc perhaps try https://github.com/stuartarchibald/datashader/commit/b59757d97778d1a30bed16c3c9351d30ff228c7d to remove the restriction on Numba version.

datashader/glyphs/line.py
esc: I will resolve this now as the code in question no longer exists.

esc:
@stuartarchibald thanks, I cherr-picked the commit locally and ran the tests with Numba 0.48.0 and 0.51.2 with and w/o JIT enabled and they passed in all instances. Thank you!

stuartarchibald:
> @stuartarchibald thanks, I cherr-picked the commit locally and ran the tests with Numba 0.48.0 and 0.51.2 with and w/o JIT enabled and they passed in all instances. Thank you!

No worries :)

jlstevens:
If no one knows yet, I'll check if 0.47 in particular works as that is our main Python 2 version for datashader...

stuartarchibald:
Think flake8 compliance is broken https://travis-ci.org/github/holoviz/datashader/jobs/726274853#L1705-L1706

jlstevens:
For reference:

```
./datashader/core.py:17:1: F401 '.utils.numba_version' imported but unused
./datashader/glyphs/line.py:7:1: F401 'datashader.utils.numba_version' imported but unused
./datashader/tests/test_antialias.py:124:1: F401 'pytest' imported but unused
```

I made these fixes myself then realized I couldn't push to the fork!

esc:
flake8 fixed and rebased.

esc:
With a newer Numba, the tests also pass on Windows. With the old 0.48 the tests fail on Windows.

esc:
🎉

PR #922, opened 2020-05-25
Fix zonal stat example in the docs

philippjfr:
Since we added special handling for zeros in a uint array but not an int array the zonal stats example that casts to int types does not mask out the zero values and we get a red background:

Before:

download (1)

After:

download (2)

I've also taken the opportunity to add an explicit error for mismatching shapes on stack because I just came across a user who got caught out by this.

### DIFF
--- a/datashader/transfer_functions/__init__.py
+++ b/datashader/transfer_functions/__init__.py
@@ -118,7 +118,7 @@ def stack(*imgs, **kwargs):
     """
     if not imgs:
         raise ValueError("No images passed in")
-    shapes = ()
+    shapes = []
     for i in imgs:
         if not isinstance(i, Image):
             raise TypeError("Expected `Image`, got: `{0}`".format(type(i)))
--- a/datashader/transfer_functions/__init__.py
+++ b/datashader/transfer_functions/__init__.py
@@ -118,9 +118,14 @@ def stack(*imgs, **kwargs):
     """
     if not imgs:
         raise ValueError("No images passed in")
+    shapes = ()
     for i in imgs:
         if not isinstance(i, Image):
             raise TypeError("Expected `Image`, got: `{0}`".format(type(i)))
+        elif not shapes:
+            shapes.append(i.shape)
+        elif shapes and i.shape not in shapes:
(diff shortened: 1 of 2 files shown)


philippjfr:
Alternatively I could also cast to `uint` types but the entire line seems highly non-obvious:

download (3)

We need to very clearly state the change in assumptions we've made about ints and uints in shade.

jbednar:
@brendancol , just highlighting this change in case it needs corresponding handling in xarray-spatial.  You don't need to do anything in this PR or on Datashader, but there may be changes needed in xarray-spatial or its docs.

datashader/transfer_functions/__init__.py
```diff
@@ -118,9 +118,14 @@ def stack(*imgs, **kwargs):
...
     if not imgs:
         raise ValueError("No images passed in")
+    shapes = []
     for i in imgs:
         if not isinstance(i, Image):
             raise TypeError("Expected `Image`, got: `{0}`".format(type(i)))
+        elif not shapes:
+            shapes.append(i.shape)
+        elif shapes and i.shape not in shapes:
+            raise ValueError("The stacked images must have the same shape.")
```
jbednar, line 128: Thanks. What was the error message or symptom before adding this exception?

philippjfr: See https://discourse.holoviz.org/t/how-to-plot-an-observer-on-a-georeferenced-tif-file/540/2

```
~/opt/anaconda3/envs/grass/lib/python3.7/site-packages/toolz/functoolz.py in flip(func, a, b)
    735     [1, 2, 3]
    736     """
--> 737     return func(b, a)
    738 
    739 

TypeError: ufunc 'over' not supported for the input types, and the inputs could not be safely coerced to any supported types according to the casting rule ''safe''
```

jbednar: Thanks.

jbednar:
I've updated the docs in https://github.com/holoviz/datashader/pull/922 to mention the change to uint32 in what seem like the appropriate places.

PR #926, opened 2020-05-25
Implements tree reduction in the dask layer

o-smirnov:
Following discussions in https://github.com/ratt-ru/shadeMS/issues/29 (in particular, using small chunks in the dask layer would, counterintuitive, explode RAM usage), @sjperkins replaced the current chunk aggregation code in dask.py with a tree reduction.  

This has been tested extensively with https://github.com/ratt-ru/shadeMS, and was found to reduce RAM usage considerably. Not tested in a CUDA context at all though, so if somebody more knowledgable than me can take a look at it, that'd be great.


sjperkins:
It looks like https://github.com/holoviz/datashader/pull/926/commits/a95dd1ae0fa203af4bea2885c29b0d78b6ac11aa doesn't handle the full range of types that can be present on Dataframe Arrays and is breaking the test suite. I'll work on a fix.

sjperkins:
@jbednar I'd be interested in your thoughts on this PR and the use of a [Reduction Operator](https://en.wikipedia.org/wiki/Reduction_Operator) to aggregate the images produced in each chunk.

o-smirnov:
@jbednar any chance to look at this?

o-smirnov:
@jbednar, ping. It would be very nice to get this (and #927) merged so we can do a proper shadeMS release in time for [ADASS](https://adass2020.es/). I attach a copy of my poster in the hope that it motivates: [shadeMS ADASS Poster.pdf](https://github.com/holoviz/datashader/files/5467971/shadeMS.ADASS.Poster.pdf)

datashader/data_libraries/dask.py
```diff
@@ -1,9 +1,12 @@
 from __future__ import absolute_import, division
 
 import dask
+import dask.array as da
 import dask.dataframe as dd
 from collections import OrderedDict
 from dask.base import tokenize, compute
+import numpy as np
+import pandas as pd
```
philippjfr, line 9: We should start adding isort or some other import linting, but could you move these up with the dask import.

philippjfr (approved):
I'm in favor of merging this, but are there any concrete benchmarks about the impact on performance and memory usage?

datashader/data_libraries/dask.py
jbednar, line 9:
```suggestion
from collections import OrderedDict
import numpy as np
import pandas as pd
import dask
import dask.array as da
import dask.dataframe as dd
from dask.base import tokenize, compute
```

datashader/data_libraries/dask.py
```diff
@@ -70,18 +76,103 @@ def default(glyph, df, schema, canvas, summary, cuda=False):
     y_mapper = canvas.y_axis.mapper
     extend = glyph._build_extend(x_mapper, y_mapper, info, append)
 
-    def chunk(df):
+    # Here be dragons
```
jbednar, line 79: Agreed, but explaining precisely what those dragons are would be helpful! :-)

datashader/data_libraries/dask.py
```diff
@@ -70,18 +76,103 @@ def default(glyph, df, schema, canvas, summary, cuda=False):
...
+                continue
+            else:
+                dtypes.append(subdtype)
+        else:
+            dtypes.append(dt)
+
+    dtype = np.result_type(*dtypes)
+    # Create a meta object so that dask.array doesn't try to look
+    # too closely at the type of the chunks it's wrapping
+    # they're actually dataframes, tell dask they're ndarrays
```
jbednar, line 104: Why?

jbednar (approved):
The changes here are _way_ over my head, so I have no idea how to tell whether they are a good idea. The basic idea of a tree reduction definitely makes sense, but I wouldn't be able to detect problems with this implementation if there are any, and as @philippjfr suggests it would be good to know what the actual impact on performance is, given that it's a clearly more complex approach than the current one.

jbednar:
Oh, and the poster looks amazing!  I'm so glad that Datashader has been useful for this, and I apologize that it's taken so long for me to catch up to all the cool things you guys are doing!

o-smirnov:
> but are there any concrete benchmarks about the impact on performance and memory usage?

I ran a few measurements for https://github.com/ratt-ru/shadeMS/issues/29 and https://github.com/ratt-ru/shadeMS/issues/34 back in the day. Basically, without this PR, my plotting problems were blowing out the RAM (on a 512GB box), so the comparison was rather binary. But I can try to do some tests with a reduced dataset.

datashader/data_libraries/dask.py
sjperkins: Low-level dask graphs can represent a very broad range of operations and data.  Dask Arrays and Dataframes are simply metadata describing the layer of an associated Low-Level Dask Array. For example:

```python
>>> dsk = {
      ('a', 0, 0): (np.full, (2, 2), 1),
      ('a', 0, 1): (np.full, (2, 2), 2),
      ('a', 1, 0): (np.full, (2, 2), 3),
      ('a', 1, 1): (np.full, (2, 2), 4)}

>>> chunks = ((2, 2), (2, 2))
>>> array = da.Array(dsk, 'a', chunks, meta=np.empty((0,0), dtype=np.int32))
>>> array
dask.array<a, shape=(4, 4), dtype=int32, chunksize=(2, 2), chunktype=numpy.ndarray>
>>> array.compute()                                                                                                                                                      
Out[7]: 
array([[1, 1, 2, 2],
       [1, 1, 2, 2],
       [3, 3, 4, 4],
       [3, 3, 4, 4]])
```

In the above example, the low-level graph, **chunks** describe the purported size of the arrays produced by the 4 dask tasks, while the **meta** describes the **type** of each chunk, here represented with an empty array. This is to support arrays other that NumPy (e.g. [sparse](https://sparse.pydata.org) arrays). At low-level graph

The dragon is that it is entirely possibly to construct **dsk** in a way that disagrees with the metadata and *this doesn't matter at all at the graph level*. The only time there's a problem is:

1. if ``array.compute()`` is directly called (Because dask will try to construct an array of type **meta**)
2. If the output of ``arrays`` tasks are used as input to tasks that expect ndarrays.

(1) doesn't apply in this case  and here, a dask Array is constructed from tasks producing Pandas Dataframes which are passed into tasks (combine, aggregate),  that expect Dataframes, circumventing (2).

The final dask Array produced by the reduction describes a task that does return an ndarray so everything works out, under the hood.

o-smirnov: Right, so in summary, no issues in the way it's currently used in the codebase (given that everything goes via dataframes), just leaving a signpost for future developers to look out for if they start hacking the code for another purpose?

sjperkins: Yep, I can leave a link to the explanation in the code.

o-smirnov:
@philippjfr, I added some benchmarks here: https://github.com/ratt-ru/shadeMS/issues/34#issuecomment-722485980.

Bottom line is, it depends on many factors, but the tree reduction version never does worse, and in some regimes (big problems, medium chunk sizes) does considerably better.

o-smirnov:
This is perhaps the most striking example of where the master branch almost runs aground -- check that RAM curve just going up and up -- while tree reduction just trundles along happily. (This tends to happen when a large number of categories, 64 in this case, is employed):

| master version | tree reduction |
|-|-|
|image|image|

jbednar:
I've fixed tests on master, so could you please rebase?

o-smirnov:
All good to go.

jbednar:
Thanks for the contribution!  Sorry I couldn't get the tests fixed in time for the conference presentation; all I accomplished last week is checking election results about 40,000 times a day. :-/  Should be ready to make a new release soon.

o-smirnov:
Aye, I think the election immobilized everyone... anyway, I realized we'll need a release anyway before I can release shadeMS to PyPI, so it was never going to be on time. Motivated people have been able to install off the github branch anyway.

jbednar:
I've tagged a dev release that's building now; see https://travis-ci.org/github/holoviz/datashader/builds/743042551 . If it completes successfully, you'll be able to do `conda install -c pyviz/label/dev datashader=0.11.2a1` try it out with this PR's contents and also #927's.

o-smirnov:
Thanks -- any idea why I can't ``pip install --pre datashader==0.11.2a1``? The badge for the dev release is up on https://pypi.org/project/datashader/, but pip refuses to see it. I've never played with pip and dev releases before, so I'm fully ignorant here...

jbednar:
The conda package is up and I've verified that it installs successfully using `conda install -c pyviz/label/dev datashader=0.11.2a1`. Looks like the pip build didn't run because of a previous unrelated job, so I restarted it. When https://travis-ci.org/github/holoviz/datashader/jobs/743042557 completes I'd hope that either a0 or a1 will be installable with pip (not sure which one this run is, as I tagged it twice! :-).

o-smirnov:
Build failed:
```
pkg_resources.DistributionNotFound: The 'rfc3986>=1.4.0' distribution was not found and is required by the application
```

jbednar:
Yep; looks like twine has undeclared dependencies. :-(  I've forced those to be installed, with the new build at https://travis-ci.org/github/holoviz/datashader/builds/743245089 

(It's been running for nearly 3 hours now!).  The result will be v0.11.2a2 if it succeeds.

jbednar:
It did not; apparently a further missing dependency. Another try: https://travis-ci.org/github/holoviz/datashader/builds/743321413

o-smirnov:
Failed again, sadly...

jbednar:
Aye.  Still building with another minor tweak. Rather than post all the URLs, just look for a green one with a version number in this table https://travis-ci.org/github/holoviz/datashader/builds/ and when there is one, that's the version that can be installed on pip.  (It's already installable on conda.)

jbednar:
The pip build finally completed, and says that it uploaded datashader-0.11.2a5.tar.gz to https://test.pypi.org/legacy/ (https://travis-ci.org/github/holoviz/datashader/jobs/743444276). I assume that means that it should be available from PyPI, but `pip install --pre datashader==0.11.2a5` doesn't find any alpha releases.  Anyone know whether the upload command needs updating? I don't have any experience with pip dev releases myself...

o-smirnov:
Not me... @Athanaseus or @gijzelaerr, any idea?

gijzelaerr:
I think probably PyPI doesn't handle extensions very well, and I recommend against using them in general, just use simple `x.y.z` versioning.

o-smirnov:
Thing is, I remember pushing out packages like radiopadre 1.0pre9, and PyPI was fine with that (and knew that 1.0pre9 < 1.0). So I'm not sure what exactly is different here!

philippjfr:
The difference is that we upload dev releases to test.pypi.org not the main pypi.org repo. We've long debated changing that but for now @jbednar or I could manually copy it over.

jbednar:
If pip handles them properly at the main repo, i.e. `pip install datashader` only installs actual releases, then I think they should automatically go to the main repo.  Alternatively, is there an easy way for people to install dev releases from test.pypi.org?

jbednar:
In any case, I do seem to remember pip now handling dev releases fine, though I don't remember the details, and we definitely aren't ready to make a full x.y.z release while CUDA and py2 are so badly broken.

philippjfr:
Yes support for --pre is pretty universal at this point, only very old pip installations won't handle it I believe. For now you can add --index-url https://test.pypi.org --pre to install from the other repo.

o-smirnov:
> For now you can add --index-url https://test.pypi.org --pre to install from the other repo

But I can't put that into my ``setup.py``, or can I? For dev purposes I'm fine installing straight off github -- it's for the pre-release that I need a package on pypi.

gijzelaerr:
2 things:

* afaik test.pypi.org is for testing the upload of your package, not for alpha packages.
* A custom upload repo is a pypi configuration setting, and should be set in the pypi configuration.

Issue #907, opened 2020-04-30
_colorize can exceed available memory if there is a large number of unique categories


maihde:
The pipeline description says the following:

```
...only one stage (Aggregation) requires access to the entire dataset. The remaining stages use a fixed-sized data structure regardless of the input dataset...
```

However, I have observed that if I have a dataset with a **very large number** of unique categories (for this example let's say > 50,000) then colorize will allocate a very large array.

```
  File "datashader/transfer_functions/__init__.py", line 352, in _colorize
    r = (data.dot(rs)/total).astype(np.uint8)
MemoryError: Unable to allocate 27.2 GiB for an array with shape (256, 256, 55804) and data type int64
```

It appears that the `_colorize` datastructure is only fixed in the image size dimensions and scales O(n) in the category dimension.

#### ALL software version info
Datashader 0.10.0

#### Description of expected behavior and the observed behavior
Ideal behavior is that _colorize could handle an arbitrary number of categories, similar to other aspects of Datashader.  Improving the memory utilization of colorize would be second best.  Documentation of limitation and/or work around is also be fine.

#### Complete, minimal, self-contained example code that reproduces the issue

Will provide in follow-up.  Thanks as always.

jbednar:
Datashader's aggregation is designed to work with arbitrarily large input datasets but it assumes that the output from the aggregation will fit into memory on a single compute node. The output from an aggregation step always scales with the width and height of the array of pixels, and also (for categorical aggregations) scales with the number of categories. I.e., categorical aggregates are of size _height * width * num_categories_.  In general, these are appropriate assumptions if rendering to an image to display on a computer screen, as such screens are limited in their x and y resolution and humans cannot reliably distinguish very many colors in an image anyway. (Nearly all categorical colormaps include fewer than 20 colors, and the largest that we distribute (in colorcet) include only 256.)

It is possible to support aggregating into arbitrarily large height, width, and num_categories, but doing so efficiently requires pre-sorting the data into partitions by spatial location and category so that each compute node processes only a restricted range of categories and/or spatial locations. Such partitioning can already be done using spatialpandas for location and a pandas/dask groupby for category, but those are both expensive operations and not always practical to do ahead of time. 

Without such partitioning and pre-sorting, each compute node would need to run through the entire dataset and simply discard points not falling into its own partition and category, which is simple enough to do but is very inefficient, performing the computation in the same time as currently done but with _m_ times as many nodes as the current approach, for _m_ partitions of the output array. 

[...]

Issue #10, opened 2016-01-21
Add cache buster to http requests 


brendancol:

jbednar:
I'd rather add a real request specification, not just a random id, so that browser caching is used when it really makes sense (as it will when switching between widget values, e.g. different transfer functions).  But that requires constructing some string that encapsulates everything about the request that could change.

brendancol:
So in the PR above I DID go the UUID route.  This is a temporary fix, and http will be removed once bokeh plot frame height/width are exposed to the server side.  The other params (transfer func, agg, field...) are not actually consumed from the url, so thought that might be confusing if they are included only to bust cache.

PR #927, opened 2020-05-29
suggested implementation for category_modulo and category_binning

o-smirnov:
I suggested this in #907 -- I would like to humbly submit a proposed implementation.

## Background

I often need to colourize (categorize) data using (a) indices from an integer column (possibly modulo some preset number of categories), or (b) by binning a float-valued column and assigning categories based on which bin a value falls into.

The data is typically too big to be in core, and is computed or loaded in a lazy fashion using dask dataframes and https://github.com/ska-sa/dask-ms. I have not found a way to map this into a Categorical column without triggering a computation (and the ensuring disk I/O), which defeats the purpose of the lazy-evaluation dask layer.

In addition, in #907 @maihde requested a mechanism for mapping a large number of categories into a reduced number of colours.

## Proposal

This PR does the following:

* Tweaks the ``by()`` reduction so that it deals with an abstract "categorizer" object (rather than explicitly using ``category_codes`` or ``category_values``)

* Implements two new categorizers, ``category_modulo`` and ``category_binning``, derived from ``category codes``. These assign categories based on the two use cases described above.

* The ``by`` constructor will now accept a categorizer object, as an alternative to a column name. Thus e.g.

  ```
  by(category_binning('x', 0, 10, 16))
  ```
  constructs a by-reduction using 16 categories, with category 0 being 0<=x<10, category 1 being 10<=x<20, etc.

  Calling the constructor with a column name implicitly constructs a ``category_codes``, thus retaining the old behaviour.

## TODO

* The cuda code path of these two new categorizers currently throws a ``NotImplementedError``. It's probably trivial to implement in cuda (as a minimum, categorization can be done on the CPU, then a to_gpu_array() can be called), but I haven't got the knowledge.

* @maihde's ``simplify_categories()`` function can be implemented as another categorizer.


o-smirnov:
Hmm ok, I gotta get the tests to work first, clearly...

jbednar:
This looks great, thanks!  I recently proposed implementing binning of numeric dimensions in https://github.com/holoviz/datashader/pull/875#issuecomment-589821583 to make 3D aggregations, and it's fun to see it actually appear! Once the tests pass I'm happy to try it out and give feedback.

o-smirnov:
Thanks, @jbednar! I'm quite keen for this and #926 to go in, because then I can release shadems on PyPI.

This PR should also allow other interesting user-defined categorizers, for example an outer product over mutiple categories, or a category remapper...

I might need some help with the Travis build though. It seems everything passes except this one thing:
```
  File "/home/travis/miniconda/envs/3.6/bin/datashader", line 11, in <module>
    load_entry_point('datashader', 'console_scripts', 'datashader')()
  File "/home/travis/build/holoviz/datashader/datashader/__main__.py", line 9, in main
    return pyct.cmd.substitute_main('datashader',args=args)
  File "/home/travis/miniconda/envs/3.6/lib/python3.6/site-packages/pyct/cmd.py", line 455, in substitute_main
    args.func(args) if hasattr(args,'func') else parser.error("must supply command to run")
  File "/home/travis/miniconda/envs/3.6/lib/python3.6/site-packages/pyct/cmd.py", line 394, in <lambda>
    parser.set_defaults(func=lambda args: fn(name, **{k: getattr(args,k) for k in vars(args) if k!='func'} ))
TypeError: fetch_data() got an unexpected keyword argument 'verbose'
```
....which has absolutely nothing to do with any code I touched, so I'm at a loss where to start here.

o-smirnov:
> I might need some help with the Travis build though. It seems everything passes except this one thing:

Actually it looks like master is currently failing in exactly the same way, so I'll just wait for you to fix it and merge master in again...

jbednar:
Ok, I've asked @kebowen730 to look into that. Stay tuned!

o-smirnov:
@jbednar with the latest changes to master all the tests pass. Could you please look into merging this?

jbednar:
Sure, but at the moment I'm distracted by SciPy2020, and haven't had a chance to look at it yet. Soon!

o-smirnov:
@jbednar any chance to look at this?

datashader/reductions.py
```diff
@@ -139,26 +258,37 @@ def _finalize(bases, cuda=False, **kwargs):
...
-        categories present.
+    cats: str or CategoryPreprocess instance
+        Name of column to aggregate over, or a categorizer object that returns categories.
+        Resulting aggregate has an outer dimension axis along the categories present.
     reduction : Reduction
         Per-category reduction function.
     """
-    def __init__(self, cat_column, reduction):
-        self.columns = (cat_column, getattr(reduction, 'column', None))
+    def __init__(self, cats, reduction):
```
jbednar, line 270: Is it necessary to rename cat_column here? That seems to hurt backwards compatibility if anyone was supplying the keyword argument, and at least in typical cases the name is still accurate.

datashader/reductions.py
```diff
@@ -43,39 +43,158 @@ def apply(self, df):
...
+    def apply(self, df):
+        result = (df[self.column] - self.offset) % self.modulo
+        if cudf and isinstance(df, cudf.DataFrame):
+            return result.to_gpu_array()
+        else:
+            return result.values
+
+class category_binning(category_modulo):
+    """
+    A variation on category_codes that assigns categories by binning a continuously-valued column.
```
jbednar, line 117:
```suggestion
    A variation on category_codes that assigns categories by binning a continuous-valued column.
```

datashader/reductions.py
```diff
@@ -43,39 +43,158 @@ def apply(self, df):
...
+    The number of categories returned is always nbins+1.
+    The last category (nbin) is for NaNs in the data column, as well as for values under/over the binned
+    interval (when include_under or include_over is False).
+
+    Parameters
+    ----------
+    column:   column to use
+    bin0:     lower bound of first bin
+    binsize:  bin size
```
jbednar, line 126: Is it important that the bin size be exactly as specified? I would think that a more natural interface would be to specify the lower and upper bounds of the range and then nbins, with binsize being computed as (upper-lower)/nbins.  That way people can calculate lower and upper from their data and easily try different nbins, whereas how they have to write an expression for binsize to make it work like that.

jbednar (commented):
Looks good; sorry for the delay in reviewing! It's hard to test it out without examples and tests, though.  Would it be possible to add those?

datashader/reductions.py
o-smirnov: Agreed, I don't remember why I even renamed it, but I put it back.

datashader/reductions.py
o-smirnov: Good suggestion.

o-smirnov:
Thanks for the review @jbednar, I've implemented the suggested fixes, and there's now a whole bunch of tests for the reductions added in. They pass for me locally.

o-smirnov:
I think master itself is currently not passing tests (I also see something like "fixture 'benchmark' not found" failing elsewhere). So once master is fixed, this should be good to go.

jbednar:
I've fixed tests on master, so could you please rebase?

o-smirnov:
Good to go!

jbednar:
Thanks so much for the contribution; this is really cool stuff!  I'll have to think about how to highlight it in the docs; any suggestions or sample bits of code welcome!

o-smirnov:
Sure, shall I just write something up right here and you can cut-and-paste appropriately?

jbednar:
That would be great, thanks!

o-smirnov:
Let's say you have a dataframe with columns named ``x``, ``y``, ``gender``, ``age`` and ``weight``. Gender is a categorical column with N categories, age is integer, the others are floats.

## Counting categories

The traditional ``count_cat`` aggregator can be used to aggregate points categorized by gender.

```python
    agg = canvas.points(df,'x','y', agg=ds.count_cat('gender'))
```

The resulting cube has three axes: x, y and gender. Each pixel in the cube will contain a count of the points of the appropriate gender that fall into the corresponding x,y bin.

The ``by`` aggregator is a generalization of this. It is constructed with two arguments: a column, and a reduction function. The following is an exact equivalent of ``ds.count_cat('gender')``:

```python
    agg = canvas.points(df,'x','y', agg=ds.by('gender', ds.count()))
```

## Aggregating statistics by category

However, more elaborate reduction functions can also be supplied:

```python
    agg = canvas.points(df,'x','y', agg=ds.by('gender', ds.mean('weight')))
```

This returns a 3D cube where each x, y, gender pixel gives the *mean weight* of that gender over the x, y bin.

## Categorizing by non-categorical columns

The examples above only work with a categorial column type. What if one wanted to categorize by a non-categorical column such as age (or even weight)? This can be done by creating a *categorizer* object for that column, and giving it to ``by``:

```python
    cat = ds.category_modulo('age', modulo=10, offset=16)
    agg = canvas.points(df,'x','y', agg=ds.by(cat, ds.mean('weight')))
```

This returns a 3D cube containing 10 slices. The category is computed as ``(age - 16)%10``. Thus, the first slice will aggregate the mean weight (over an x, y bin) for ages 16, 26, 36, ..., the second slice ages 17, 27, 37, ...., etc.

The previous example is, admittedly, a contrived example. Here is something more realistic: let us look at the standard deviation in weight for particular age brackets:

```python
    cat = ds.category_binning('age', lower=20, higher=100, nbins=8, include_under=False, include_over=False)
    agg = canvas.points(df,'x','y', agg=ds.by(cat, ds.std('weight')))
```

This returns a 3D cube containing 9 (nbins+1) slices. Slice 0 gives the stddev in weight (per each x, y bin) for ages [20,30), slice 1 for ages [30,40), ..., slice 7 for ages [90,100). The last slice, #8, is the "odd bin", i.e. it catches all "other" categories -- in this case, it gives the stddev in weight for ages below 20, over 100, and for an age of NaN (the latter would only be a possibility if age was a float, of course).

If we were to give ``include_under=True`` (default), ages below 20 would be included in the first bin, nominally [20,30). Likewise, with ``include_over=True``, ages >= 100 will be included in the last bin, [90,100). In this case, the "odd bin" will only contain points with NaN ages.

Binning can also be done over a float-valued column:

```python
    cat = ds.category_binning('weight', lower=0, higher=200, nbins=10)
    agg = canvas.points(df,'x','y', agg=ds.by(cat, ds.max('age')))
```
This returns a 3D cube containing 11 slices. Each point gives the maximum age over a particular x, y and weight bin. The last slice (#10) will catch negative and NaN weights, as well as weights >= 200.

## Custom categorizers

It is possible to implement your own custom categorizations. You must derive a subclass from ``category_codes``, ``category_modulo`` or ``category_binning``. You must then implement the following methods ``__init__``, ``_hashable_inputs``, ``categories``, ``validate``, and ``apply``. See e.g. the implementation of ``category_modulo`` in ``reductions.py`` for an example.

Issue #932, opened 2020-06-19
Numba <0.49 prevents installation with cuDF>=0.14


AjayThorve:
Referencing issue #841 which mentions the reason for numba version pinning. Hoping to get this to work with newer version of rapids [cuDF](https://github.com/rapidsai/cudf), which requires numba version be >=0.49

FYI: I have tried building datashader from source and it seems to work with numba 0.49

philippjfr:
There was a major performance regression in numba 0.49 so we added that pin. Now that 0.50 is out we should unpin and release again.

jbednar:
Yes. I'm travelling and unlikely to get to it this week, but @philippjfr , if you have verified that 0.50 fixes the performance issues, then please do tag a new Datashader release.

jonmmease:
I can try to run the GPU tests this evening with cuDF 0.14+ and numba 0.50+.  @philippjfr would you have time to double check the CPU performance regression?

philippjfr:
Sorry should have replied, I've already checked and the performance seems fine with 0.50.0.

philippjfr:
Would appreciate you running the GPU tests, I've set up an environment with cudf 0.14 but haven't had time to upgrade my drivers yet, which seems to be required.

jonmmease:
Sure thing. I'll update this issue this evening after giving it a go. If things are in good shape, could you cut a patch release over the next few days?

jonmmease:
@philippjfr @jbednar I spent a while testing against cudf 0.14 and numba 0.50 and there's a big roadblock.

A PR (https://github.com/numba/numba/pull/5556) in numba 0.50 changed the semantics of `cuda.atomic.max` and `cuda.atomic.min`. These functions used to treat NaNs as missing values, making it possible to initialize an array of NaNs and then use the atomic operations to compute mins/maxes without needing a special case the first time the reduction applied to an element.  Now in numba 0.50, these functions propagate NaNs, meaning that this workflow will always result in an array of NaNs.  Unfortunately, I don't see a way to work around this in Datashader without introducing a race condition.

A request to reproduce the old behavior (under the names `nanmin`/`nanmax`) is being discussed in https://github.com/numba/numba/issues/5820. We won't be able to update to numba 0.50+ until this is implemented, and at that point we'll either need to drop numba <0.50, or include some special case logic to choose the right function based on the numba version.

philippjfr:
[...]

PR #934, opened 2020-06-30
Fixes issue with Numba CUDA target altering max/min semantics.

stuartarchibald:
As title.

Closes https://github.com/holoviz/datashader/issues/932

### DIFF
--- a/datashader/transfer_functions/_cuda_utils.py
+++ b/datashader/transfer_functions/_cuda_utils.py
@@ -1,6 +1,8 @@
 from __future__ import division
+from distutils.version import LooseVersion
 from math import ceil, isnan, nan
 from numba import cuda
+import numba
 import cupy
 import numpy as np
 
@@ -70,13 +72,35 @@ def masked_clip_2d(data, mask, lower, upper):
     masked_clip_2d_kernel[cuda_args(data.shape)](data, mask, lower, upper)
 
 
+# Behaviour of numba.cuda.atomic.max/min changed in 0.50 so as to behave as per
+# np.nanmax/np.nanmin
+if LooseVersion(numba.__version__) >= LooseVersion("0.51.0"):
+    @cuda.jit
+    def cuda_atomic_nanmin(ary, idx, val)
+        return cuda.atomic.nanmin(ary, idx, val)
+    @cuda.jit
+    def cuda_atomic_nanmax(ary, idx, val)
+        return cuda.atomic.nanmax(ary, idx, val)
+elif LooseVersion(numba.__version__) <= LooseVersion("0.49.0"):
+    @cuda.jit
+    def cuda_atomic_nanmin(ary, idx, val)
+        return cuda.atomic.min(ary, idx, val)
+
+    @cuda.jit
+    def cuda_atomic_nanmax(ary, idx, val)
+        return cuda.atomic.max(ary, idx, val)
+else:
+    # WHAT DO YOU WANT TO DO HERE?
+    pass
+
+
 @cuda.jit
 def masked_clip_2d_kernel(data, mask, lower, upper):
     i, j = cuda.grid(2)
     maxi, maxj = data.shape
     if i >= 0 and i < maxi and j >= 0 and j < maxj and not mask[i, j]:
-        cuda.atomic.max(data, (i, j), lower)
-        cuda.atomic.min(data, (i, j), upper)
+        cuda_atomic_max(data, (i, j), lower)
+        cuda_atomic_min(data, (i, j), upper)
 
 
 # interp
--- a/datashader/reductions.py
+++ b/datashader/reductions.py
@@ -8,6 +8,8 @@ import xarray as xr
 
 from datashader.glyphs.glyph import isnull
 from numba import cuda as nb_cuda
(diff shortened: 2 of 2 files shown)


stuartarchibald:
This is entirely untested and relies on https://github.com/numba/numba/pull/5941, open question about what to do for version Numba 0.50. which has neither the old atomic min/max behaviour nor the new nanmin/nanmax functions.

datashader/transfer_functions/_cuda_utils.py
```diff
@@ -70,13 +72,35 @@ def masked_clip_2d(data, mask, lower, upper):
...
+    def cuda_atomic_nanmax(ary, idx, val)
+        return cuda.atomic.nanmax(ary, idx, val)
+elif LooseVersion(numba.__version__) <= LooseVersion("0.49.0"):
+    @cuda.jit
+    def cuda_atomic_nanmin(ary, idx, val)
+        return cuda.atomic.min(ary, idx, val)
+
+    @cuda.jit
+    def cuda_atomic_nanmax(ary, idx, val)
+        return cuda.atomic.max(ary, idx, val)
```
jbednar, line 91: Are the `def`s really needed here?  Would it not be sufficient to do `cuda_atomic_nanmin = cuda.atomic.nanmin` in the first `if` and `cuda_atomic_nanmin = cuda.atomic.min` in the second?

datashader/transfer_functions/_cuda_utils.py
```diff
@@ -70,13 +72,35 @@ def masked_clip_2d(data, mask, lower, upper):
...
+else:
+    # WHAT DO YOU WANT TO DO HERE?
```
jbednar, line 93:
```suggestion
    assert("Datashader's CUDA support requires numba!=0.50.0")
```

datashader/transfer_functions/_cuda_utils.py
```diff
@@ -70,13 +72,35 @@ def masked_clip_2d(data, mask, lower, upper):
...
+
 @cuda.jit
 def masked_clip_2d_kernel(data, mask, lower, upper):
     i, j = cuda.grid(2)
     maxi, maxj = data.shape
     if i >= 0 and i < maxi and j >= 0 and j < maxj and not mask[i, j]:
-        cuda.atomic.max(data, (i, j), lower)
-        cuda.atomic.min(data, (i, j), upper)
+        cuda_atomic_max(data, (i, j), lower)
+        cuda_atomic_min(data, (i, j), upper)
```
jbednar, line 103:
```suggestion
        cuda_atomic_nanmax(data, (i, j), lower)
        cuda_atomic_nanmin(data, (i, j), upper)
```

datashader/transfer_functions/_cuda_utils.py
stuartarchibald: Unfortunately I think they are (and these may need to be device functions too), the `cuda.atomic.max/min` are stubs that are basically a token known to the compiler, if they appear as globals I don't think the compiler will know what to do with them.

stuartarchibald:
Thanks for the review @jbednar, once 0.51 is out I'll revisit (or someone else can, I have no particular attachment to the patch, feel free to push to my branch :)) with view to test this and make it actually work opposed to it just being a guess!

datashader/transfer_functions/_cuda_utils.py
jbednar: Ok, thanks.

dHannasch:
> Thanks! I haven't tested it, but hopefully @jonmmease and/or @philippjfr can.

If anyone's up for testing this, I've fixed the picky little errors at https://github.com/holoviz/datashader/pull/940. (See https://github.com/stuartarchibald/datashader/pull/1.)
Note however that https://github.com/holoviz/datashader/pull/940 has one more change that might be important. @philippjfr [asked](https://github.com/holoviz/datashader/pull/940#discussion_r453246153) why we couldn't just import `numba.cuda.atomic.nanmin` and `numba.cuda.atomic.nanmax` directly (that is, why do we need to wrap them in `@cuda.jit def`s?), and I didn't have an answer, so I tried splitting the difference and using them directly in reductions.py `_append_cuda`, but using the `@cuda.jit` wrappers in `masked_clip_2d_kernel`. So if something goes wonky in `_append_cuda` but not in `masked_clip_2d_kernel`, blame me, not @stuartarchibald.

Also, if you're going to manually test, best to just test Python 3.7 for now. There are currently unrelated problems with Dask that make the test suite fail on 2.7 and 3.6.

stuartarchibald:
xref https://github.com/numba/numba/pull/6020 which is @gmarkall carrying on with some extra details on top of the original PR on the Numba side.

AjayThorve:
@dHannasch I tested this with numba 0.51(https://github.com/numba/numba/releases/tag/0.51.0rc1), `python 3.7.8`.

The version check seems to work as well. Just a couple of fixes required

philippjfr:
I'd like to go ahead and merge and then fix things up in a separate PR. Any objections?

AjayThorve:
> I'd like to go ahead and merge and then fix things up in a separate PR. Any objections?

@philippjfr that works

PR #937, opened 2020-07-05
numba-0.50 is out with the performance regression fix, isn't it ?

stonebig:
see https://github.com/numba/numba/issues/5746#issuecomment-638072219


philippjfr:
Yes but numba 0.50 also completely breaks GPU support.

jbednar:
Right; we can't quite update yet; see https://github.com/numba/numba/issues/5820 .

PR #939, opened 2020-07-11
ENH: updated draft of MPL artist [WIP]

nvictus:
Working on resolving issues with @tacaswell's #200 at Scipy 2020 sprints along with @manzt.

The `DSArtist` now takes in a `datashader.Pipeline` object and so far can handle the case of a 2D raster with a quantitative colormap but not the other 3D categorical case when a color_key is used. 

We currently infer the colormap by applying the datashader pipeline's operations manually rather than using the callable itself. We use the aggregation part of the pipeline (agg, transform_fn) to get a vmin and vmax in order to build and set a matplotlib colormap and norm.

image

We'll keep working on the categorical case, but we wanted to share this now and also see if there is still interest in merging this into datashader.


philippjfr:
Neat, I'll try to play around with this tonight.

jbednar:
There's definitely still interest in merging this into Datashader, just way too many unfunded promises to too many people. :-) Anything you can do to make it usable is great!

nvictus:
Great to hear that the interest is there! Totally understand being spread thin. 

Trevor and I put some more effort into supporting (1) quantitative rasters with a colorbar by registering a matplotlib norm and cmap, and (2) categorical rasters with a legend by providing a function to generate legend elements. Both methods involve inferring the colors by shoving artificial data through the pipeline, but it seems to work just fine. Cursor events also report the actual data instead of image values.

Here's what it looks like:

image

philippjfr:
@nvictus Would you mind rebasing the PR? I've fixed the CI issues.

nvictus:
Done!

philippjfr:
@nvictus what's the status here? Should we merge and release this? Do you want to make further tweaks?

nvictus:
HI @philippjfr. I added some docstrings and a convenience function `dsshow` to mirror the `imshow`/`matshow` functionality in matplotlib. I also added a couple basic tests.

I chose to duplicate the `Pipeline` API in `dsshow` to skip the intermediate step of creating a pipeline object, especially since we don't actually end up calling the object itself in the artist. But I'm open to suggestions.

Other than that and any external suggestions, I think I'm done tweaking.  If @tacaswell approves of the `dsshow` API, I vote to merge.

@tacaswell, FYI I also set interpolation to "none" by default as it was generating artifacts.

tacaswell:
I only gave it a very quick review and nothing raised any flags.  I think the API an `dsshow` should be more driven by what makes sense for datashader which I am not the best person to comment on.

I am super stoked this got pushed across the finish line :) !

jbednar:
This looks really cool.  I'm still trying to get my head around it, but I don't think the colorbar-inferring code is working.  E.g. I tried using the following test script adapted from your example:

```
import numpy as np, pandas as pd, datashader as ds, matplotlib
from datashader.mpl_ext import dsshow
matplotlib.use('macosx')

from functools import partial
dyn = partial(ds.tf.dynspread, max_px=40, threshold=0.5)


n = 1000000

df = pd.DataFrame({
    'x': np.r_[np.random.randn(n) - 0.5, 0.03*np.random.randn(n) + 0.5],
    'y': np.r_[np.random.randn(n),       0.03*np.random.randn(n)],
    'c': pd.Categorical(np.r_[['cloud 1'] * n, ['cloud 2'] * n])
})

da1 = dsshow(
    df,
    ds.Point('x', 'y'),
    spread_fn=dyn,
    aspect='equal'
)

plt.colorbar(da1);
plt.draw()
plt.show()
```

The hover shows that the actual values are around 100 in the very center of this plot:

image

but from the colorbar you would expect those values to be in the 80000 range.  Here I've taken your second Gaussian and reduced its radius so that it's all concentrated in a few pixels (barely visible in the above plot), and hovering over that small patch shows that the counts in that region are much higher:

image

Those are the values I'd expect when focusing a large distribution into a small region, so I think the plotting and the hovering are working fine. But the colorbar is acting like the numerical range is linearly distributed over the color range, when it clearly isn't when using the default "eq_hist" color shading. 

We've recently gone through a painful process of trying to add colorbar support for colormappers like eq_hist to Bokeh, and at one point @jlstevens did try a similar "linspace ramp" approach for computing colorbars. Maybe he can recall why we decided that it ultimately would not work. We ended up with a different approach that was merged at https://github.com/bokeh/bokeh/pull/10240 , which supports JavaScript-based histogram equalization colormapping in a way that the bins remain available for use when building a colorbar. Actually building the colorbar is something that @mattpap is working on this week.  It's been a painful slog at the Bokeh level, but I guess for Matplotlib it can be slightly easier because the process can happen in Python rather than JS. I don't think it can be done by transforming a ramp, though!

philippjfr:
Not sure this will be helpful here but note that I ported my Javascript implementation of eqhist to a matplotlib Normalize class, which you can find [here](https://github.com/holoviz/holoviews/pull/4567/files#diff-7835bbf73589d76bfa74d6de328be2eeR393).

91184816-5c654600-e6ed-11ea-8585-410497f5949a

This produces a colorbar that is at least accurate even if it doesn't yet implement appropriate ticking behavior and therefore causes values at the bottom of the range to bunch up.

nvictus:
Right, I see now what the problem is. Normalizations like linear and log are invariant to the data so a linear ramp should translate into the right colors, but `eq_hist` normalization is a function of the data values being displayed, so if you feed it a ramp, you're changing the mapping.

One problem is that `color_fn`/`tf.shade` encapsulates both data normalization and colormapping and isn't introspectable, so even with the `EqHistNormalize` class, there's no way of telling `DSArtist` which matplotlib `Normalize` class to use given an opaque datashading pipeline.

Is there currently a way in datashader to separate the concerns of normalization and coloring? Maybe `color_fn` should always assume the data is mapped to [0, 1] and rely on a `Normalize` class to interpolate data values to that range? How is it done in holoviews?

philippjfr:
>One problem is that color_fn/tf.shade encapsulates both data normalization and colormapping and isn't introspectable, so even with the EqHistNormalize class, there's no way of telling DSArtist which matplotlib Normalize class to use given an opaque datashading pipeline.

You would not be using `tf.shade` with EqHistNormalize, you'd simply use the aggregation output and then let matplotlib handle the normalization colormapping.

jbednar:
Right. Datashader's tf.shade() is necessary only in one of two situations: (1) no other colormapper is available (e.g. when rendering directly to PNG, with no plotting library installed), or (2) you need a colormapping feature available in tf.shade but not available in your plotting library.

The two main features of tf.shade() not commonly available in plotting libraries are histogram equalization and categorical color mixing. Bokeh now has histogram equalization, and if you use Philipp's EqHistNormalize you can get similar support for Matplotlib, which means that you should never need tf.shade() for single-category plotting with a plotting library, and you'll have all the features of your plotting library available (colorbars, hovering, etc.)

We have not yet figured out how to implement categorical color mixing in Bokeh, and I haven't thought about how to do that in Matplotlib, so categorical plots would still need to use tf.shade() if you want the colors mixed. (Just overlaying the colors works fine in Bokeh and presumably Matplotlib, but that doesn't achieve the color mixing for categories that tf.shade does to avoid overplotting and ambiguity.)

jbednar:
Note that if you follow the above advice, you won't be able to use Datashader's Pipeline object as-is, because it returns an RGB image rather than the underlying aggregate. To avoid that problem, you could make a small change to Pipeline where you optionally skip the color_fn step:

```
bins = core.bypixel(self.df, canvas, self.glyph, self.agg)
trans = self.transform_fn(bins)        
img = trans if self.color_fn is None else self.color_fn(trans)
return self.spread_fn(img)
```
This will mean that spread_fn may get passed an aggregate rather than an image, but that's ok now that spreading is supported for aggregates and not just images (and in fact works better on aggregates).

Alternatively, if you look at the Pipeline code, you can see that the actual code is trivial, so you can simply dispense with Pipeline here and use your own custom pipeline inside the DSArtist or as a separate object.

jbednar:
Actually, I see that you are already breaking up the pipeline inside your DSArtist, so I'd recommend you simply ignore Datashader's Pipeline and accept your own arguments for the stages you're using, so that you can apply them yourself however you like.

jbednar:
@nvictus , we're working on a new Datashader release, and I'd love to have good Matplotlib support in it!  Any chance you could try out Philipp's [EqHistNormalize](https://github.com/holoviz/holoviews/pull/4567/files#diff-7f4d270c5d618d32d9ebdc7faceb0bcccca93b8be2c0a935651a8adfe80ea12fL397) class and get that integrated in here? If you start to work on it again, first please rebase the PR so that tests will pass with your changes and to avoid conflicts.

nvictus:
Alright, I finally found some time to work in the `EqHistNormalize` by delegating normalization and colormapping of 2D aggregates to matplotlib. To make the logic less chaotic, I broke up `DSArtist` into `QuantitativeDSArtist` for 2D aggregation and `CategoricalDSArtist` for 3D categorical aggregation which is still handled by `tf.shade`. 

The front-end `dsshow` function dispatches to the appropriate artist class: I simply assumed that categorical aggregators must be  subclasses of `ds.reductions.by` and are otherwise single-category. If that's not the case, let me know if there's a clean way to distinguish the two cases a priori or if we should expose separate API functions.

Now, I'm not sure what to do about `spread_fn`. If spreading can now be applied to aggregates before coloring (as part of `transform_fn`), is there still a need for it?

nvictus:
Btw, I tried to preserve some of the `tf.shade` semantics in `dsshow`: 

* the `norm` argument accepts "linear", "log", "cbrt" and "eq_hist" as values, which produce a compatible matplotlib norm.
* the `cmap` argument accepts either a matplotlib colormap name/instance or a list of colors, as in `tf.shade`. To replicate the case of a transparency-based colormap with a single color + min_alpha + alpha, I added a convenience function `alpha_colormap` to generate such a colormap.

akhmerov:
Just a quick check: what do you think about using sorting for histogram normalization, see [here](https://gitlab.kwant-project.org/qt/cookbook/-/blob/master/cmap-normalizers.py#L5-18) for a minimal implementation.

datashader/mpl_ext.py
```diff
@@ -0,0 +1,633 @@
...
+    glyph : Glyph
+        The glyph to bin by.
+    reduction : Reduction, optional, default: :class:`~.count`
+        The reduction to compute per-pixel.
+    transform_fn : callable, optional
+        A callable that takes the computed aggregate as an argument, and
+        returns another aggregate. This can be used to do preprocessing before
+        passing to the ``color_fn`` function.
+    spread_fn : callable, optional
+        ???
```
jbednar, line 461: Don't forget to delete this bit if not supporting spread_fn.

datashader/mpl_ext.py
```diff
@@ -0,0 +1,633 @@
...
+    the artist can be used to make a colorbar with ``fig.colorbar``.
+
+    If the aggregation is 3D categorical (i.e. generates a composited
+    image from several categorical components), you can use the
+    :meth:`CategoricalDSArtist.get_legend_elements` method to obtain patch
+    handles that can be passed to ``ax.legend`` to make a legend.
+
+    Examples
+    --------
+    Generate two gaussian point clouds and plot (1) the density as a
```
jbednar, line 527:
```suggestion
    Generate two Gaussian point clouds and plot (1) the density as a
```

jbednar (commented):
@nvictus, looks great!

- Sure, the spread_fn doesn't need to be a separate argument now, as long as there are examples showing how to supply spreading to the transform_fn argument.
- The tf.shade support looks perfect.
- Seems like the main thing left now is just documentation. Maybe a new section at the end of getting_started/Interactivity.ipynb?

@akhmerov, I don't have any _a_ _priori_ opinion about sorting versus the current eq_hist implementation. Sorting does sound slow, but maybe it's not? Does it behave any differently, e.g. avoiding the large gaps in the colorbar for small integer counts that the current implementation has? If it has such advantages, we might consider offering it separately, e.g. as `eq_sort`.

philippjfr:
The sorting approach is almost certainly going to be faster than my iterative, reverse eq_hist implementation, which is used here, but I'd also love to see a detailed comparison between the two.

nvictus:
Two small updates:

1. I noticed some weird colorbar stability issues with `EqHistNormalize`. It's seems to be due to the fact that the source data for histogram normalization is consumed each time the norm is called and there seem to be some extra calls happening in the background with different inputs. I found a solution that seems to work: consume the source data when the norm is rescaled using `autoscale` or `autoscale_None`. This already happens in the DSArtist every time the pipeline is run to adjust `vmin` and `vmax`. The same strategy should work if we switch to @akhmerov's sorting based method.

2. I figured out how to intercept the final image when letting matplotlib do the shading, so I was able to resurrect `spread_fn`. Given that spreading can now be done on aggregates, I thought it would make sense to make the names of `transform_fn` and `spread_fn` more generic, so I tentatively renamed them `on_agg` and `on_image`, respectively. What do you guys think?

> Seems like the main thing left now is just documentation. Maybe a new section at the end of getting_started/Interactivity.ipynb?

I'll look into it!

jbednar:
> I tentatively renamed them on_agg and on_image, respectively.

Sounds good! In our group we usually use the word "hook" for parameters like that, which would be "agg_hook" (or "post_agg_hook" to be completely explicit) and "image_hook" or "post_image_hook", but I don't have any strong opinion.

nvictus:
I've changed the parameter names to `agg_hook` and `shade_hook` as suggested and I've added a section to `examples/getting_started/3_Interactivity.ipynb`.

jbednar:
Thanks for the docs! Those are really helpful and I think I can finish them up from there. However, as you can see from the failing tests, there appears to be an issue. It's currently failing only on Ubuntu, and initially I was unable to reproduce any problem on my Mac. However, it looks like it's actually a transient problem that could indicate something more serious, because when I run `pytest --nbsmoke-run -k .ipynb examples/getting_started/3_Interactivity.ipynb` multiple times locally, some of the time it fails, saying `nbclient.exceptions.DeadKernelError: Kernel died` and `ERROR    traitlets:client.py:690 Kernel died while waiting for execute reply.` That seems very worrisome, but I'm not sure how to debug it.

nvictus:
It does seem worrisome. I use an Ubuntu machine and I haven't been able to reproduce it yet.

According to [this](https://github.com/nteract/papermill/issues/476#issuecomment-590997645), it could be a message throughput issue, OOM or segfault.

jbednar:
I would expect it to be a segfault, because the volumes of data involved are not terribly large, and when the error happens, it happens pretty quickly. I would guess that it's always doing something incorrect, e.g. writing outside the bounds of an array, but that it only sometimes hits an area not allowed for writing and triggering a segfault. Numba turns off out of bounds warnings on arrays, so I tried disabling Numba via `env NUMBA_DISABLE_JIT=1 pytest --nbsmoke-run -k .ipynb examples/getting_started/3_Interactivity.ipynb`. It did take 10X longer to run, but I wasn't able to reproduce the issue, which could mean that it's a Numba bug. But if so, it's one that only seems to come up when using the new examples, because if I delete those examples from the notebook and run `pytest --nbsmoke-run -k .ipynb examples/getting_started/3_Interactivity.ipynb` 8 times I get 0 failures, while I get about 3 failures out of 8 runs when those examples are included.

nvictus:
Hmm... is disabling this cell alone sufficient to eliminate the failures?

image

nvictus:
Though, there is also axes sharing in this cell:

image

jbednar:
I was able to get it to fail even when the only two cells included in that section are `import matplotlib.pyplot as plt ; from datashader.mpl_ext import dsshow, alpha_colormap` and `dsshow(df, ds.Point('x', 'y'), norm='eq_hist');`. Just the import alone wasn't enough; out of a dozen runs it never failed. And `eq_hist` isn't needed; `linear` makes it fail as well. So a single `dsshow` is enough to do it.

jbednar:
It's tricky; I tried reducing it down to `from datashader.mpl_ext import dsshow ; dsshow(df, ds.Point('x', 'y'), norm='linear');`, and that still failed but only one time out of 16.  At that rate, it's going to be hard to show when it _doesn't_ happen, but at least this is a fairly minimal way to reproduce it.

nvictus:
I set up a new environment and I can reproduce this now. 

```
_______________________________________________ test session _______________________________________________
nbconvert.preprocessors.execute.DeadKernelError: Kernel died
-------------------------------------------- Captured log call ---------------------------------------------
ERROR    traitlets:execute.py:508 Kernel died while waiting for execute reply.
========================================= short test summary info ==========================================
FAILED examples/getting_started/3_Interactivity.ipynb::/home/nezar/local/devel/scipy2020/datashader/examples/getting_started/3_Interactivity.ipynb
============================================ 1 failed in 12.08s ============================================

```

I wonder if the failure is nbconvert-specific.

jbednar:
It's hard to imagine why it would be, but so far, yes.

nvictus:
Based on what nbsmoke does, I can reproduce with the following:

```python
from nbconvert.preprocessors import ExecutePreprocessor
import nbformat
import io

for i in range(16):
    print(i)
    nb = io.open('./examples/getting_started/3_Interactivity.ipynb', encoding='utf-8')
    notebook = nbformat.read(nb, as_version=4)
    ep = ExecutePreprocessor(allow_errors=False)
    output = ep.preprocess(notebook, {})
```

```
0
1
2
3
4
5
6
7
8
9
10
double free or corruption (out)
```

The error message does look like a segfault in the kernel.

jbednar:
You made it farther than I did, thanks!  I was hoping to get this into today's Datashader release, but I'll have to punt on that. We can have another release when it's ready.

nvictus:
That's too bad. Thanks for trying to push it to the finish line. 

FWIW, it's starting to seem to me like it could be a race condition between bokeh and matplotlib. I don't seem to get the failure if I comment out `hv.output(backend="bokeh")`.

nvictus:
I can narrow it down specifically to the use of the `hv.util.Dynamic` element. Removing it seems to eliminate the failure.

On the other hand, if I place a `hv.output(backend="matplotlib")` plot in a cell after it instead of using `dsshow`, I get the same failure.

jbednar:
I'm not sure I follow all the steps there, but are you saying you can reproduce this issue without using mpl_ext? If so, please post the script that could reproduce that, and we can test that independently of this PR. If it's independent of this PR, postponing a merge of this PR won't help! :-/

nvictus:
Right! Sorry for not being clear: yes, I can get nbsmoke to fail when running a notebook that completely lacks `mpl_ext`.

Take the current notebook `3_Interactivity.ipynb` without the `mpl_ext` additions and move the following cell to the very end:

```python
hv.output(backend="matplotlib")
agg2 = ds.Canvas().points(df,'x','y')
hd.datashade(points)  +  hd.shade(hv.Image(agg2))  +  hv.RGB(np.array(tf.shade(agg2).to_pil()))
```

Then try again after removing or commenting out `+ (pts * dynamic).relabel("Dynamic hover")` in the hover info section. In my hands, it ceases to fail after that.

jbednar:
Ok, I've been able to get the kernel to die on that notebook in Jupyter in a similar way on Datashader master, so whatever problems are involved here appear to be independent of this PR. So I re-ran the tests here, which happened to pass this time, and I'll go ahead and merge it, opening a separate issue for the segfaults.

@akhmerov , we may still want to switch to your eqhist implementation, but if so I'll do so separately from this PR.  @nvictus , I'll also finish the updates to 3_Interactivity.ipynb that were started here.

@nvictus, thanks so much for all your hard work here, and @tacaswell , thanks for getting the ball rolling initially!

philippjfr:
Glad to finally see this go in!

jbednar:
And thanks to @philippjfr for the eq_hist implementation! What a saga!

philippjfr:
Note that reverse mapping the eq_hist is the difficult bit, the forward pass is relatively straightforward but doesn't give you colorbar support. Haven't investigated in detail whether that's easier based on the sorting approach, suspect it might be though.

akhmerov:
> @akhmerov , we may still want to switch to your eqhist implementation, but if so I'll do so separately from this PR.

Sounds reasonable. Unfortunately I don't have bandwidth to work on that, so my ping was mainly an FYI.

nvictus:
Wonderful! Thanks, all! This was a fun project, and something I've wanted to have for a while. Also thanks to @manzt who worked with me to revive the initial work by @tacaswell.

manzt:
Thanks for pushing this all the way through, @nvictus!

tacaswell:
:tada: I'm super stoked that this got merged!

jbednar:
Yeah, but I've never gotten the damned website to build properly and show
it off, so I've never announced it!  Keeps getting bumped off my to-do
list...

PR #940, opened 2020-07-11
Try #934 with Numba installed from GitHub

dHannasch:
This PR is just an attempt to get some nice green checkmarks attached to @stuartarchibald's excellent https://github.com/holoviz/datashader/pull/934, which will make Datashader compatible with Numba 0.51 (see e.g. https://github.com/numba/numba/issues/5820#issuecomment-651402266).

https://github.com/numba/numba/pull/5941 hasn't even been merged yet, and Numba 0.51 certainly isn't out yet, but we can make sure all the gears mesh right now by installing from the git repository directly.

### DIFF
--- a/datashader/transfer_functions/_cuda_utils.py
+++ b/datashader/transfer_functions/_cuda_utils.py
@@ -76,18 +76,18 @@ def masked_clip_2d(data, mask, lower, upper):
 # np.nanmax/np.nanmin
 if LooseVersion(numba.__version__) >= LooseVersion("0.51.0"):
     @cuda.jit
-    def cuda_atomic_nanmin(ary, idx, val)
+    def cuda_atomic_nanmin(ary, idx, val):
         return cuda.atomic.nanmin(ary, idx, val)
     @cuda.jit
-    def cuda_atomic_nanmax(ary, idx, val)
+    def cuda_atomic_nanmax(ary, idx, val):
         return cuda.atomic.nanmax(ary, idx, val)
 elif LooseVersion(numba.__version__) <= LooseVersion("0.49.0"):
     @cuda.jit
-    def cuda_atomic_nanmin(ary, idx, val)
+    def cuda_atomic_nanmin(ary, idx, val):
         return cuda.atomic.min(ary, idx, val)
 
     @cuda.jit
-    def cuda_atomic_nanmax(ary, idx, val)
+    def cuda_atomic_nanmax(ary, idx, val):
         return cuda.atomic.max(ary, idx, val)
 else:
     # WHAT DO YOU WANT TO DO HERE?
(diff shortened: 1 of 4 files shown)


datashader/transfer_functions/_cuda_utils.py
```diff
@@ -70,13 +72,35 @@ def masked_clip_2d(data, mask, lower, upper):
...
+    @cuda.jit
+    def cuda_atomic_nanmin(ary, idx, val)
+        return cuda.atomic.min(ary, idx, val)
+
+    @cuda.jit
+    def cuda_atomic_nanmax(ary, idx, val)
+        return cuda.atomic.max(ary, idx, val)
+else:
+    # WHAT DO YOU WANT TO DO HERE?
+    pass
```
philippjfr: Does it have to be wrapped in a jitted function or would this work?

```suggestion
if LooseVersion(numba.__version__) >= LooseVersion("0.51.0"):
    from numba.cuda.atomic import nanmin as cuda_atomic_min, nanmax as cuda_atomic_max
elif LooseVersion(numba.__version__) <= LooseVersion("0.49.0"):
    from numba.cuda.atomic import min as cuda_atomic_min, max as cuda_atomic_max
else:
    cuda_min, cuda_max = None, None
```

(As long as we change it below and in the other file?

dHannasch: Good question. I don't know.

philippjfr: Cc @stuartarchibald

stuartarchibald: Worth a try but I'm fairly sure they won't be recognised, these "functions" aren't actually functions, but behave like them in machine code once compiled. They work via a bunch of redirections:
https://github.com/numba/numba/blob/bbb10ebfaabcf0c691ee822ef2c861c92ff3585a/numba/cuda/stubs.py#L379 is a marker so that type info can be associated here  https://github.com/numba/numba/blob/bbb10ebfaabcf0c691ee822ef2c861c92ff3585a/numba/cuda/cudadecl.py#L273-L284 and then lowered here https://github.com/numba/numba/blob/54fbed92162c7636b4a047cf92c57f775ad3cd7d/numba/cuda/cudaimpl.py#L530-L544 the compiler picks up the marker and does the right thing.

dHannasch: Huh. It sounded like it was working for @AjayThorve (https://github.com/holoviz/datashader/pull/934#issuecomment-671505652).
(He might have meant he tested the original unaltered #934, but that had a couple of SyntaxErrors as noted in https://github.com/stuartarchibald/datashader/pull/1 so I assume he meant he tested this version.)

No idea whether there might be more obscure problems that would only crop up in some situations, though. I really don't understand what's going on in any of the three pieces of code you linked (very much not a numba expert), so I can't speculate on how the environment might play a role.

AjayThorve: > Huh. It sounded like it was working for @AjayThorve ([#934 (comment)](https://github.com/holoviz/datashader/pull/934#issuecomment-671505652)).
Actually what worked for me was these changes (https://github.com/holoviz/datashader/pull/947) on top of #934 (which is now merged)

dHannasch: Ah, I see. Yeah, no particular reason to think we can do without the wrappers, then.

philippjfr:
Thanks for contributing these changes and sorry we didn't end up merging this particular PR. Is there anything in here that should still be merged? I'll close for now.

PR #943, opened 2020-07-12
Drop Python 2.7 support.

### LINKED ISSUE #942: Drop support for Python 2.7
#### Is your feature request related to a problem? Please describe.
Datashader currently cannot be used with newer versions of Numba, which makes it incompatible with other tools that rely on newer versions of Numba, see e.g. https://github.com/holoviz/datashader/issues/932 .

However, Numba has officially dropped support for Python 2.7 and Numba 0.51 is planned to be incompatible with Python 2.7 (https://github.com/numba/numba/commit/051bb50ca20f1850550428622134d6f59795097f).

#### Describe the solution you'd like
If Datashader officially drops support for Python 2.7, then it will be possible to use Datashader with Numba 0.51 (with some tweaks to Datashader's code (https://github.com/holoviz/datashader/pull/934)), thus enabling the use of Datashader alongside other tools.

#### Describe alternatives you've considered
The obvious alternative is for Numba to continue supporting Python 2.7, but Numba really does not want to do that (https://numba.pydata.org/numba-doc/latest/reference/python27-eol.html).

#### Additional context
It's worth noting that Datashader depends on a lot of other packages that have also dropped Python 2.7 support, including NumPy, pandas, Pillow, Bokeh, and SciPy. https://python3statement.org/
Soon enough, maintaining Python 2.7 support is going to require pinning a *lot* of versions, not just Numba.

dHannasch:
Resolves #942.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -73,12 +73,6 @@ jobs:
         - export NUMBA_DISABLE_JIT=1; doit test_unit_nojit
       after_success: codecov
 
-    # python 2 flake checking typically catches python 2 syntax
-    # errors where python 3's been assumed...
-    - <<: *default
-      env: DESC="py2 flakes" PYTHON_VERSION=2.7 OPTS=""
-      script: doit test_lint
-
     # test using python 3.7
     - <<: *default
       env: PYTHON_VERSION=3.7


philippjfr:
Thanks for this PR, we are in fact dropping Python2 support in the coming release (0.12) but are now switching to GitHub Actions.

PR #944, opened 2020-07-12
Fix AttributeError: module 'pandas.compat' has no attribute 'string_types'

### LINKED ISSUE #945: Test suite currently fails on Travis
The master branch, on Travis.

#### Description of expected behavior and the observed behavior
We would want the test suite to pass.

The test suite fails to pass on Travis. Inspecting the logs traces the AttributeError to dask/dataframe/core.py:2566. You can see this at https://github.com/holoviz/datashader/pull/939 and https://github.com/holoviz/datashader/pull/943 . (Neither of those pull requests has any changes that would cause a Dask error to appear, but if anyone is skeptical, we can re-run the Travis tests with no changes from master at all.) Fortunately, Dask has already released a bugfix.

(It's unclear why the problem only showed up now, but the environment on Travis isn't exactly the same every time --- presumably something changed in another package causing the bug to appear.)

The issue appears when running the test suite on Travis.

#### Stack traceback and/or browser JavaScript console output
https://travis-ci.org/github/holoviz/datashader/jobs/707234805

dHannasch:
Resolves #945 for Python 3 (making those pass as seen in this Travis build), but not for Python 2.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -12,7 +12,7 @@ install_requires = [
     # `conda install dask[complete]` happily gives you dask...which is
     # happily like pip's dask[complete]. (conda's dask-core is more
     # like pip's dask.)
-    'dask[complete] >=0.18.0',
+    'dask[complete] >=2.0.0',
     'toolz >=0.7.4',  # ? for some dask issue (dasks does only >=0.7.3)
     'datashape >=0.5.1',
     'numba >=0.37.0,<0.49',


philippjfr:
Seems like this should be on dask to appropriately pin pandas. Pinning dask so aggressively isn't ideal since there are many older dask/pandas combinations that will work just fine.

philippjfr:
Thanks again @dHannasch, but as I said above datashader should not be responsible for second-guessing dask pinning and they are in fact generally pretty good at getting bug fixe releases available in a timely manner.

PR #946, opened 2020-07-14
Implement 3D Volume regridding

philippjfr:
Implements basic 3D volume regridding only supporting mean downsampling and nearest upsampling for now.

- [ ] Double check array orientations
- [ ] Add tests
- [ ] Add docs

### DIFF
--- a/datashader/core.py
+++ b/datashader/core.py
@@ -11,11 +11,14 @@ from six import string_types
 from xarray import DataArray, Dataset
 from collections import OrderedDict
 
-from .utils import Dispatcher, ngjit, calc_res, calc_bbox, orient_array, \
-    compute_coords, dshape_from_xarray_dataset
+from .utils import (
+    Dispatcher, ngjit, calc_res, calc_bbox, orient_array,
+    compute_coords, calc_res3d, calc_bbox3d, compute_coords3d,
+    dshape_from_xarray_dataset,
+)
 from .utils import get_indices, dshape_from_pandas, dshape_from_dask
 from .utils import Expr # noqa (API import)
-from .resampling import resample_2d, resample_2d_distributed
+from .resampling import resample_2d, resample_2d_distributed, resample_3d
 from . import reductions as rd
 
 try:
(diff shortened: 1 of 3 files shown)


datashader/core.py
```diff
@@ -1131,10 +1148,173 @@ def raster(self,
...
         return DataArray(data, coords=coords, dims=dims, attrs=attrs)
 
+    def volume(self, source, nan_value=None, agg='mean', interpolate='nearest'):
+        """Sample a raster dataset by canvas size and bounds.
+
+        Handles 3D xarray DataArrays.
+
+        Missing values (those having the value indicated by the
+        "nodata" attribute of the raster) are replaced with `NaN` if
+        floats, and 0 if int.
```
jbednar, line 1158:
```suggestion
        floats, and 0 if int. Thus if 0 is not a suitable missing-value indicator, 
        for your integer data, convert it to a float type before calling this method.
```

datashader/core.py
```diff
@@ -1131,10 +1148,173 @@ def raster(self,
...
+
+        Parameters
+        ----------
+        source : xarray.DataArray or xr.Dataset
+            2D or 3D labelled array (if Dataset, the agg reduction must
+            define the data variable).
+        nan_value : int or float, optional
+            Optional nan_value which will be masked out when applying
+            the resampling.
```
jbednar, line 1167:
```suggestion
        nan_value : int or float, optional
            Optional nan_value that will be masked out when applying
            the resampling.
```

How does this interact with the "nodata" attribute of the raster? Does it override it? Should clarify that here.

datashader/core.py
```diff
@@ -1131,10 +1148,173 @@ def raster(self,
...
+        agg : Reduction, optional default=mean()
+            Resampling mode when downsampling raster.
+            options include: first, last, mean, mode, var, std, min, max
+            Accepts an executable function, function object, or string name.
+        interpolate : str, optional  default=linear
+            Resampling mode when upsampling raster.
+            options include: nearest, linear.
```
jbednar, line 1174:
```suggestion
            Options include: nearest, linear.
```

jbednar:
@philippjfr , can you compare your approach here with https://github.com/holoviz/datashader/pull/927 and see if there is anything you need to take from here? I haven't tried to see how the two binning approaches differ or are similar.

philippjfr:
Having trouble seeing how they relate to each other. Isn't #927 going from points -> grids and this is going from grids -> grids?

jbednar:
Right you are; this is 3D volume _re_-gridding, and the other is 3D volume gridding.  So, the actual task is the other way around -- please check if anything from #927 is relevant to this one. Thanks!

maximlt:
@philippjfr any update with this? Can we move forward?

PR #947, opened 2020-08-10
Supporting cuDF 0.14+

AjayThorve:
- removing numba pinning which prevented datashader 0.11 working with cuDF 0.14+
- fixed a couple of issues raised with the PR #934 

I have tested this with `cuDF 0.14 + numba 0.49`  & `cuDF 0.15(nightlies) + numba 0.51.0rc1`

cc @philippjfr @jonmmease

### DIFF
--- a/datashader/transfer_functions/_cuda_utils.py
+++ b/datashader/transfer_functions/_cuda_utils.py
@@ -75,18 +75,18 @@ def masked_clip_2d(data, mask, lower, upper):
 # Behaviour of numba.cuda.atomic.max/min changed in 0.50 so as to behave as per
 # np.nanmax/np.nanmin
 if LooseVersion(numba.__version__) >= LooseVersion("0.51.0"):
-    @cuda.jit
+    @numba.jit
     def cuda_atomic_nanmin(ary, idx, val)
         return cuda.atomic.nanmin(ary, idx, val)
-    @cuda.jit
+    @numba.jit
     def cuda_atomic_nanmax(ary, idx, val)
         return cuda.atomic.nanmax(ary, idx, val)
-elif LooseVersion(numba.__version__) <= LooseVersion("0.49.0"):
-    @cuda.jit
+elif LooseVersion(numba.__version__) <= LooseVersion("0.49.1"):
+    @numba.jit
     def cuda_atomic_nanmin(ary, idx, val)
         return cuda.atomic.min(ary, idx, val)
 
-    @cuda.jit
+    @numba.jit
     def cuda_atomic_nanmax(ary, idx, val)
         return cuda.atomic.max(ary, idx, val)
 else:
@@ -99,8 +99,8 @@ def masked_clip_2d_kernel(data, mask, lower, upper):
     i, j = cuda.grid(2)
     maxi, maxj = data.shape
     if i >= 0 and i < maxi and j >= 0 and j < maxj and not mask[i, j]:
-        cuda_atomic_max(data, (i, j), lower)
-        cuda_atomic_min(data, (i, j), upper)
+        cuda_atomic_nanmax(data, (i, j), lower)
+        cuda_atomic_nanmax(data, (i, j), upper)
 
 
 # interp
--- a/datashader/transfer_functions/_cuda_utils.py
+++ b/datashader/transfer_functions/_cuda_utils.py
@@ -76,22 +76,21 @@ def masked_clip_2d(data, mask, lower, upper):
 # np.nanmax/np.nanmin
 if LooseVersion(numba.__version__) >= LooseVersion("0.51.0"):
     @numba.jit
-    def cuda_atomic_nanmin(ary, idx, val)
+    def cuda_atomic_nanmin(ary, idx, val):
         return cuda.atomic.nanmin(ary, idx, val)
     @numba.jit
-    def cuda_atomic_nanmax(ary, idx, val)
+    def cuda_atomic_nanmax(ary, idx, val):
         return cuda.atomic.nanmax(ary, idx, val)
 elif LooseVersion(numba.__version__) <= LooseVersion("0.49.1"):
     @numba.jit
-    def cuda_atomic_nanmin(ary, idx, val)
+    def cuda_atomic_nanmin(ary, idx, val):
         return cuda.atomic.min(ary, idx, val)
 
     @numba.jit
-    def cuda_atomic_nanmax(ary, idx, val)
+    def cuda_atomic_nanmax(ary, idx, val):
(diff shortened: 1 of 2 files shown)


datashader/transfer_functions/_cuda_utils.py
```diff
@@ -75,32 +75,31 @@ def masked_clip_2d(data, mask, lower, upper):
...
 
 @cuda.jit
 def masked_clip_2d_kernel(data, mask, lower, upper):
     i, j = cuda.grid(2)
     maxi, maxj = data.shape
     if i >= 0 and i < maxi and j >= 0 and j < maxj and not mask[i, j]:
-        cuda_atomic_max(data, (i, j), lower)
-        cuda_atomic_min(data, (i, j), upper)
+        cuda_atomic_nanmax(data, (i, j), lower)
+        cuda_atomic_nanmax(data, (i, j), upper)
```
gmarkall, line 102: Changing `cuda_atomic_min` to `cuda_atomic_nanmax` looks a little strange here, as the original PR #934 kept `max` and `min`: https://github.com/holoviz/datashader/pull/934/files#diff-d28b7c34f16db5fd95f4c3a0723b9dc9R102-R103 - is this fixing another issue, or is it a copy/paste issue?

setup.py
```diff
@@ -15,7 +15,7 @@
     'dask[complete] >=0.18.0',
     'toolz >=0.7.4',  # ? for some dask issue (dasks does only >=0.7.3)
     'datashape >=0.5.1',
-    'numba >=0.37.0,<0.49',
+    'numba >=0.37.0',
```
philippjfr, line 18: I think it should be this:

```suggestion
    'numba <0.49.0,>=0.51.0',
```

datashader/transfer_functions/_cuda_utils.py
philippjfr: Seems to be something @jbednar suggested as a change in the original PR. Not quite sure why either. I don't think it makes much of a difference but also doesn't seem necessary.

datashader/transfer_functions/_cuda_utils.py
```diff
@@ -75,32 +75,31 @@ def masked_clip_2d(data, mask, lower, upper):
 # Behaviour of numba.cuda.atomic.max/min changed in 0.50 so as to behave as per
 # np.nanmax/np.nanmin
 if LooseVersion(numba.__version__) >= LooseVersion("0.51.0"):
-    @cuda.jit
-    def cuda_atomic_nanmin(ary, idx, val)
+    @numba.jit
```
philippjfr, line 78: Guess I'm missing something, what was wrong with `cuda.jit` here?

datashader/transfer_functions/_cuda_utils.py
gmarkall: @jbednar suggested changing `cuda_atomic_min` to `cuda_atomic_nanmin` - this changes it to `cuda_atomic_nanmax` - i.e. changing a `min` into a `max, which seems a more significant change.

philippjfr: Oh, totally missed that. Good catch!

setup.py
AjayThorve: It seems to work for version upto `numba 0.49.1`, how about 'numba <=0.49.1,>=0.51.0'

datashader/transfer_functions/_cuda_utils.py
AjayThorve: if I use the `@cuda.jit` decorator, I get the following error:

```bash
TypingError: Failed in nopython mode pipeline (step: nopython frontend)
Untyped global name 'cuda_atomic_nanmax': cannot determine Numba type of <class 'numba.cuda.compiler.Dispatcher'>
```
and since calling a numba.jit function from CUDA automatically compiles an equivalent CUDA device function.


Although going through some numba examples, I realized that `@cuda.jit(device=True)` would achieve the same device function, and the result works. I have updated the code. Thanks for pointing this out

datashader/transfer_functions/_cuda_utils.py
AjayThorve: Apologies for the confusion, this was a copy/paste issue, it should be `cuda_atomic_nanmin`.  I have updated the code

setup.py
philippjfr: 0.49 had a major performance regression affecting datashader and was excluded for that reason.

datashader/transfer_functions/_cuda_utils.py
philippjfr: Thanks for looking into it, really appreciate it!

datashader/transfer_functions/_cuda_utils.py
philippjfr: Thanks for the fix, and thanks @gmarkall for catching it.

setup.py
AjayThorve: got it. I have pushed the fix :)

setup.py
```diff
@@ -15,7 +15,7 @@
     'dask[complete] >=0.18.0',
     'toolz >=0.7.4',  # ? for some dask issue (dasks does only >=0.7.3)
     'datashape >=0.5.1',
-    'numba >=0.37.0,<0.49',
+    'numba <0.49.0,>=0.51.0',
```
jbednar, line 18: This change now claims that Datashader can work with Numba older than 0.37.  Is that really true?  I don't recall any patches added for working with older Numba versions like that.  I think we need to express that it works with Numba greater than 0.37 _except_ 0.49 and 0.50; not sure how to say that in this expression language.

philippjfr: That's true we should keep the lower pin but is `'numba >=0.37.0,<0.49.0,>=0.51.0'` even valid?

AjayThorve: I think in that case, something like this would work: `numba >=0.37.0, <0.49.0,>=0.51.0`

AjayThorve: @philippjfr `numba >=0.37.0,<0.49.0,>=0.51.0` this will not resolve as all 3 conditions will never be satisfied. This seems to work on my end: `'numba >=0.37.0, !=0.49.*, !=0.50.*'`

philippjfr: Right, that was my worry and thanks for figuring that out. Excluding 0.49 and 0.50 like that seems like the best option.

jbednar: Sounds good, thanks!

AjayThorve: fixed it

philippjfr:
Merging, I'll fix the CI issues shortly.

PR #948, opened 2020-08-12
Use language: python on travis

philippjfr:
Various CI fixes including pinning of pytest and switching to language: python image on Travis CI.


philippjfr:
@jonmmease Any idea what's going on with the spatialpandas tests here? I've already pinned pyarrow<1.0. I haven't looked at it deeply yet just checking if you have any immediate ideas?

philippjfr:
Oh nevermind, I think it's the old RaggedArray tests that are failing.

PR #954, opened 2020-08-24
Documentation updates for aggregate spreading

jlstevens:
Updates the docs following the merge of #771.


examples/getting_started/2_Pipeline.ipynb
```diff
@@ -675,6 +674,34 @@
...
+   "source": [
+    "tf.spread(img, mask=mask)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Although this approach continues to works and can be more concise than\n",
+    "aggregate spreading, it is *not* recommended as the operation is less\n",
```
jbednar, line 696:
```suggestion
    "Although both types of spreading are supported, RGB spreading\n",
    "is *not* recommended as the operation is less\n",
```

I don't think either approach has an overall advantage for concise code, because there are cases where agg spreading is much more concise, e.g. if you spread the agg once and then colormap it multiple times.  In that particular example we're showing multiple spreading types, but in general people probably look more often at different colormaps than at different spreading shapes, so I'd suspect that agg-based spreading is actually shorter overall. In any case it's not really relevant to mention here.

examples/getting_started/2_Pipeline.ipynb
```diff
@@ -675,6 +674,34 @@
...
+    "well defined in RGB space than aggregate space. This can be seen by\n",
+    "comparing the two visualizations above: the RGB spreading has fuzzy\n",
+    "orange areas showing through the higher valued purple distribution. In\n",
+    "addition, aggregate spreading better preserves the structure of the\n",
+    "aggregate data which is useful when using other visualization\n",
+    "strategies, such as using the [Bokeh hover inspection\n",
+    "tool](https://bokeh.org/) with the datashader support in\n",
+    "[HoloViews](http://holoviews.org/).\n",
```
jbednar, line 704:
```suggestion
    "addition, aggregate spreading preserves the numerical value of the\n",
    "aggregate data, which can then be used with later processing stages\n",
    "such as the [Bokeh hover inspection\n",
    "tool](https://bokeh.org/) and the datashader support in\n",
    "[HoloViews](http://holoviews.org/).\n",
```

jbednar:
Hmm; I see that the Spreading section is using an example that only makes sense after the Colormapping section, since categorical colors aren't introduced before that.  But logically it comes before colormapping in the pipeline.  Hmm; not sure how to make that flow linearly, then.

examples/getting_started/2_Pipeline.ipynb
jlstevens: Sure. I guess what I wrote was based on the particular example I was dealing with and probably can't be generalized to how everyone might be using datashader.

examples/getting_started/2_Pipeline.ipynb
jlstevens: Suggestion looks good, thanks!

jbednar:
Ok, looks good. I'll merge and we can reorganize later if we figure out a good way.

PR #955, opened 2020-08-25
Drop ds.geo and ds.spatial modules

thuydotm:


philippjfr:
I would expect all of `datashader.geo` and `datashader.spatial` to be removed at once after appropriate deprecation cycles.

thuydotm:
@philippjfr Can you kindly specify which version that both `datashader.geo` and `datashader.spatial` will be dropped? Should I continue working on this so that it will be ready whenever needed?

philippjfr:
@jbednar Any thoughts here? We've added deprecation warnings for both modules already, so do we drop them in 0.12?

jbednar:
Yes, I'm ready to drop them entirely for the next release.  We do need to update some of our examples still, but dropping them entirely will force us to do that!

thuydotm:
I think we can safely remove the whole `datashader.geo` module and point the user guide notebook `8_Geography` to xarray-spatial. In `datashader.spatial`, `proximity`, `viewshed` and `zonal` can be now dropped but we still need `points` for `Canvas.points`. Also, the example `2_Points` will stay unchanged. How about `hillbert_curve`? I don't see it in user-guide, also don't see it used yet. Should it be removed or not?

jbednar:
ds.spatial.points (including ds.spatial.hilbert_curve, part of its implementation) has moved to [spatialpandas](https://github.com/holoviz/spatialpandas), and so we should be depending on spatialpandas and xarray_interactive for _all_ of what's in ds.geo and ds.spatial; there shouldn't be anything left not covered by one of those two external packages. We do have to rewrite ds.Canvas.points() to call SpatialPandas instead of ds.spatial.points, though.

thuydotm:
@jbednar I see, thanks for clarifying.

thuydotm:
Since the spatial indexing logic has been moved to `spatialpandas`, I have removed the call to SpatialPointsFrame in ds.Canvas.points(). The 2 modules ds.geo and ds.spatial now totally removed. Could you review this PR and let me know if any change needed?

examples/user_guide/2_Points.ipynb
```diff
@@ -13,212 +13,28 @@
...
-   "metadata": {},
-   "outputs": [],
-   "source": [
-    "import holoviews as hv\n",
-    "from holoviews.operation.datashader import datashade\n",
-    "hv.extension('bokeh')\n",
-    "\n",
-    "datashade(hv.Points(sframe, kdims=['x', 'y'], vdims=['cat']), aggregator=ds.count_cat('cat')).opts(width=400, height=400)"
+    "In most cases, Datashader must iterate through your entire dataset to render any plot, because it cannot assume the datapoints have been sorted in any particular order. Thus, the aggregation performance is dependent on the number of datapoints in your entire dataframe, not just those in the current viewport (x and y range). If you have a large dataset covering a wide area and you want to support fast local operations (e.g. if you have data at a global level but analysis is typically done in small local regions), you may want to use [spatialpandas](https://github.com/holoviz/spatialpandas), which is designed to work with Datashader and supports spatial indexing using [Hilbert curves](https://en.wikipedia.org/wiki/Hilbert_curve). Please see the [spatialpandas Ov](https://github.com/holoviz/spatialpandas/blob/master/examples/Overview.ipynb) notebook.\n",
```
jbednar, line 16:
```suggestion
    "In most cases, Datashader must iterate through your entire dataset to render any plot, because it cannot assume the datapoints have been sorted in any particular order. Thus, the aggregation performance is dependent on the number of datapoints in your entire dataframe, not just those in the current viewport (x and y range). If you have a large dataset covering a wide area and you want to support fast local operations (e.g. if you have data at a global level but analysis is typically done in small local regions), you may want to use [spatialpandas](https://github.com/holoviz/spatialpandas), which is designed to work with Datashader and supports spatial indexing using [Hilbert curves](https://en.wikipedia.org/wiki/Hilbert_curve). Please see the [spatialpandas Overview](https://github.com/holoviz/spatialpandas/blob/master/examples/Overview.ipynb) notebook.\n",
```

examples/user_guide/8_Geography.ipynb
```diff
@@ -4,731 +4,39 @@
...
-    "\n",
-    "OBSERVER_X = -12.5\n",
-    "OBSERVER_Y = 10\n",
-    "\n",
-    "H = 400\n",
-    "W = 400\n",
-    "\n",
-    "canvas = ds.Canvas(plot_width=W, plot_height=H,\n",
-    "                   x_range=(-20, 20), y_range=(-20, 20))\n",
+    "These functionality is provided in [xarray-spatial](github.com/makepath/xarray-spatial) library.\n",
```
jbednar, line 16:
```suggestion
    "These functionality is provided in the [xarray-spatial](github.com/makepath/xarray-spatial) library.\n",
```

datashader/core.py
```diff
@@ -192,18 +192,7 @@ def points(self, source, x=None, y=None, agg=None, geometry=None):
...
-            if 'datashader.spatial.points' in sys.modules:
-                from datashader.spatial.points import SpatialPointsFrame
-                if (isinstance(source, SpatialPointsFrame) and
-                        source.spatial is not None and
-                        source.spatial.x == x and source.spatial.y == y and
-                        self.x_range is not None and self.y_range is not None):
-
-                    source = source.spatial_query(
-                        x_range=self.x_range, y_range=self.y_range)
             glyph = Point(x, y)
```
jbednar, line 196: This PR seems to remove support for SpatialPointsFrame entirely, while I think we should continue supporting spatial indexing, just no longer assume it's found in Datashader. So I *think* we need to reinstate the deleted code from core.py here, but check for spatialpandas instead. I didn't find SpatialPointsFrame in spatialpandas, though.  @jonmmease , can you suggest what changes we need to make here to avoid depending on the now-deleted spatial module and instead get this functionality from spatialpandas?

jbednar (changes requested):
Looks good, thanks!  Let's see what @jonmmease says about the spatialpandas support before merging.

datashader/core.py
jonmmease: This logic is specific to the old spatial points frame, so it's not needed when using spatialpandas.

jbednar: Thanks! How does the corresponding functionality work when using SpatialPandas?  I.e., how is the spatial range communicated from Datashader to SpatialPandas in that case?

jonmmease: To use the points glyph as an example, the Dask partition-level spatial indexing is done at

https://github.com/holoviz/datashader/blob/dd8f634eb9c61887255f8be80e2354cba55dcef6/datashader/core.py#L209-L215

Here `DaskGeoDataFrame.cx_partitions` is basically equivalent to `SpatialPointsFrame.spatial_query`

The within partition spatial indexing is done in
https://github.com/holoviz/datashader/blob/dd8f634eb9c61887255f8be80e2354cba55dcef6/datashader/glyphs/points.py#L269-L272

jonmmease:
Wanted to cross reference https://github.com/pyviz-topics/examples/pull/97. This was my old WIP PR to port the `SpatialPointsFrame` demo notebook to `spatialpandas`.  As I recall, at that time there was still some work needed in HoloViews to get spatialpandas `DaskGeoDataFrame` working with datashader.

datashader/core.py
jbednar: Perfect, thanks. Looks like this PR is mergeable then!

jbednar:
Thanks. Looks like this can be merged now, but we will need to address https://github.com/pyviz-topics/examples/pull/97 and https://github.com/holoviz/holoviews/issues/4234 before releasing Datashader.

jbednar (approved):

Issue #958, opened 2020-09-02
Canvas.raster(): wrong output coords when sampling on same x range and y range as of source input


thuydotm:
#### ALL software version info
datashader version 0.11.1

#### Description of expected behavior and the observed behavior
Seems like there is a bug in Canvas.raster() that makes the coords of output DataArray incompatible with the source input DataArray when sampling on the same x range and y range.

#### Complete, minimal, self-contained example code that reproduces the issue

```python
import pandas as pd
import datashader as ds
from datashader.transfer_functions import shade, stack

width, height  = 800, 600
df = pd.DataFrame({'y': [-3, -1, -4, -4, 1, 3], 'x': [-5, -8, -5, -5, 0, 5]})

cvs = ds.Canvas(plot_width=width, plot_height=height, x_range=(-10, 10), y_range=(-5, 4))

line_agg = cvs.line(df, x='x', y='y')
sample_raster_agg = cvs.raster(line_agg)

img1 = shade(sample_raster_agg)
img2 = shade(line_agg)

stack(img1, img2)
```

```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-3-fee18f015693> in <module>
      14 img2 = shade(line_agg)
      15 
----> 16 stack(img2, img1)

~\Anaconda3\envs\xarray-spatial\lib\site-packages\datashader\transfer_functions\__init__.py in stack(*imgs, **kwargs)
    134     imgs = xr.align(*imgs, copy=False, join='outer')
    135     with np.errstate(divide='ignore', invalid='ignore'):
--> 136         out = tz.reduce(tz.flip(op), [i.data for i in imgs])
    137     return Image(out, coords=imgs[0].coords, dims=imgs[0].dims, name=name)
    138 

~\Anaconda3\envs\xarray-spatial\lib\site-packages\toolz\functoolz.py in __call__(self, *args, **kwargs)
    301     def __call__(self, *args, **kwargs):
    302         try:
--> 303             return self._partial(*args, **kwargs)
    304         except TypeError as exc:
    305             if self._should_curry(args, kwargs, exc):

~\Anaconda3\envs\xarray-spatial\lib\site-packages\toolz\functoolz.py in flip(func, a, b)
    735     [1, 2, 3]
    736     """
--> 737     return func(b, a)
    738 
    739 

TypeError: ufunc 'over' not supported for the input types, and the inputs could not be safely coerced to any supported types according to the casting rule ''safe''

```

[...]

PR #959, opened 2020-09-04
Canvas.raster(): fix floating point error in output coords

thuydotm:
Addresses #958

When we resample a raster data set over a full extent canvas of same size, recalculating `coords` might cause extremely small floating-point errors. This leads to discrepancy between `coords` of input and output DataArray/Dataset (while we expect them to be exactly the same).

cc: @brendancol

### DIFF
--- a/datashader/core.py
+++ b/datashader/core.py
@@ -1130,17 +1130,23 @@ x- and y-coordinate arrays must have 1 or 2 dimensions.
         # Compute DataArray metadata
         if np.isclose(left, self.x_range[0]) and \
                 np.isclose(right, self.x_range[1]) and \
-                np.isclose(bottom, self.y_range[0]) and \
-                np.isclose(top, self.y_range[1]) and \
-                np.size(xvals) == self.plot_width and \
-                np.size(yvals) == self.plot_height:
-            # do not recalculate coords if resampling on same x range, y range,
-            #    and same canvas size as of source input
-            #    to avoid floating point representation error.
-            xs, ys = xvals, yvals
+                np.size(xvals) == self.plot_width:
+            # To avoid floating point representation error,
+            # do not recompute x coords if same x_range and same plot_width
+            xs = xvals
         else:
-            xs, ys = compute_coords(self.plot_width, self.plot_height,
-                                    self.x_range, self.y_range, res)
+            xs, _ = compute_coords(self.plot_width, self.plot_height,
+                                   self.x_range, self.y_range, res)
+
+        if np.isclose(bottom, self.y_range[0]) and \
+                np.isclose(top, self.y_range[1]) and \
+                np.size(yvals) == self.plot_height:
+            # To avoid floating point representation error,
+            # do not recompute y coords if same y_range and same plot_height
+            ys = yvals
+        else:
+            _, ys = compute_coords(self.plot_width, self.plot_height,
+                                   self.x_range, self.y_range, res)
 
         coords = {xdim: xs, ydim: ys}
         dims = [ydim, xdim]
--- a/datashader/core.py
+++ b/datashader/core.py
@@ -1128,25 +1128,19 @@ x- and y-coordinate arrays must have 1 or 2 dimensions.
             data = data.astype(dtype)
 
         # Compute DataArray metadata
-        if np.isclose(left, self.x_range[0]) and \
-                np.isclose(right, self.x_range[1]) and \
-                np.size(xvals) == self.plot_width:
-            # To avoid floating point representation error,
-            # do not recompute x coords if same x_range and same plot_width
-            xs = xvals
-        else:
-            xs, _ = compute_coords(self.plot_width, self.plot_height,
-                                   self.x_range, self.y_range, res)
 
-        if np.isclose(bottom, self.y_range[0]) and \
-                np.isclose(top, self.y_range[1]) and \
-                np.size(yvals) == self.plot_height:
-            # To avoid floating point representation error,
-            # do not recompute y coords if same y_range and same plot_height
-            ys = yvals
+        # To avoid floating point representation error,
+        # do not recompute x coords if same x_range and same plot_width,
+        # do not recompute y coords if same y_range and same plot_height
+        close_x = np.isclose([left, right], self.x_range).all() and np.size(xvals) == self.plot_width
+        close_y = np.isclose([bottom, top], self.y_range).all() and np.size(yvals) == self.plot_height
+        if close_x and close_y:
+            xs, ys = xvals, yvals
         else:
-            _, ys = compute_coords(self.plot_width, self.plot_height,
-                                   self.x_range, self.y_range, res)
+            xs, ys = compute_coords(self.plot_width, self.plot_height,
+                                    self.x_range, self.y_range, res)
+            xs = xvals if close_x else xs
+            ys = yvals if close_y else ys
 
         coords = {xdim: xs, ydim: ys}
         dims = [ydim, xdim]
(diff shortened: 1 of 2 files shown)


brendancol:
@thuydotm awesome I'm going to test this branch out and will post back

philippjfr:
Looking fine to me, happy to defer to @brendancol to test, review and merge.

jbednar:
Seems ok to me, though whenever I do see small floating-point differences like that I wonder how they arose...

thuydotm:
There is a section from Python documentation that explains in details the errors in float operations:
https://docs.python.org/2/tutorial/floatingpoint.html#floating-point-arithmetic-issues-and-limitations

Cited from the documentation, most decimal fractions cannot be represented exactly as binary fractions. A consequence is that, in general, the decimal floating-point numbers you enter are only approximated by the binary floating-point numbers actually stored in the machine. The errors in Python float operations are inherited from the floating-point hardware.

jbednar:
Oh, to be sure; there are many intricacies in float handling on real hardware; we have all had many painful experiences with those! But float issues are deterministic, and underlying any difference in floating point values is an expression or a value that is different in some way, even if just in the order that expressions are applied. And in many cases we have been too quick to blame generic "floating point issues" when in fact there were actual off-by-1/2 or similar issues in our code, and we've had to come along years later and figure out that our fix wasn't correct at all. I've been burned by this often enough in the past 25 years of numerical coding that now I'm always suspicious. :-)  Here, I cannot spot any error, but I couldn't spot them in those previous cases either, and only figured them out much later after much painful debugging of edge cases.

philippjfr:
>But float issues are deterministic, and underlying any difference in floating point values is an expression or a value that is different in some way, even if just in the order that expressions are applied.

This doesn't apply here, the situation where this occurs here is when you feed in a DataArray and all the ranges and width/height are the same. Now thinking about it though I wonder in which scenarios this actually happens and shouldn't it just short-circuit in that case and return the original dataset? Alternatively shouldn't this also handle the case where only one axis changes in range or density?

jbednar:
> the situation where this occurs here is when you feed in a DataArray and all the ranges and width/height are the same. 

But they aren't the same (at the floating point level), or there would be no issue, right?  So here what I'm worried about is just whether we're doing something subtly wrong inside Datashader that makes things not quite match up.  But I have no evidence for that, only suspicion, which feeds on not knowing precisely where the differences actually come from.

philippjfr:
I'm reasonably confident the logic is correct, because I spent way too much time on it. A linspace can diverge a bit from something loaded from disk, potentially generated in a different language and totally different system so I'm not all that surprised.

jbednar:
Ok, that's the explanation I was looking for: truncation of precision due to saving in a file format.  The story is then that some bin edges were computed (in some language, on some hardware), saved to disk with some finite precision, then loaded back into Datashader for processing, slightly losing some precision in the process.  The goal here is to "snap" to the values saved in that file if we get close enough to them, even if it's not precisely the same, so that the overall behavior is less surprising.

That all sounds great, but then I still have a concern: shouldn't this "snapping" simply be an optimization that makes the behavior more reasonable in a certain situation, by avoiding aliasing or other artifacts caused by a misalignment of a grid?  That's not what was reported in #958, which was about an error, not about a suboptimal rendering.  

So I'm still worried not that this snapping is a bad idea (it does seem appropriate), but that it doesn't address the underlying issue. I.e., even if we were to deliberately resample a raster at a very, verly slightly different coordinate spacing, shouldn't it still work, with no errors?  Seems like it should work, but with one column or row being skipped or duplicated in the output.  Avoiding that skipped or duplicated column or row is great, but that's an optimization in behavior, not a bug fix, and here a bug fix seems needed.

philippjfr:
The problem isn't exactly in our code, rather it's due to the fact that for doing stacking or similar operations we are using xarrays aligning code which expects exact matches. We should look into whether we can make xr.align a little fuzzy on the alignment. Not on a computer right but I'll check whether there is some form of tolerance parameter there.

jbednar:
Makes sense. In that case our bug would be that the error message ("TypeError: ufunc 'over' not supported for the input types, and the inputs could not be safely coerced to any supported types according to the casting rule ''safe''") seems to have nothing to do with the problem. Trying to stacking unaligned xarrays is not uncommon and should have a useful error message. Thanks for looking into it.

brendancol:
@jbednar @philippjfr My general expectation from `ds.Canvas` is always returns aggregates with equal height / width and coords.  I didn't think persisting to disk had any bearing here, although I understand there could be other issues there. 

Regarding tf.stack, I agree a better error message would help.  Do you think it could be useful to be able to stack unaligned arrays by creating a new canvas "under the hood" based on the first input to stack, then calling canvas.raster on the others to align them?

jbednar:
> My general expectation from ds.Canvas is that it always returns aggregates with equal height / width and coords. I didn't think persisting to disk had any bearing here, although I understand there could be other issues there.

My expectation is the same, but the claim from the initial comment in this issue is that there is a difference between the input and the output rasters, not that various output rasters from the same Canvas differ in their height/width/coords.  So if you provide a raster generated outside of Datashader somehow, with coords stored internally, then Datashader generates its own coords based on height/width/range, then I'm not surprised those don't match; we have no control over any truncation or other inaccuracies that might be in the _input_ raster, only those we calculate for the _output_ raster. I'm still not sure exactly how to proceed, though, which is waiting on @philippjfr to investigate xarray tolerance parameters.

thuydotm:
@jbednar @philippjfr @brendancol Any chance we can get this PR merged? Would be great if we have this worked since it is needed for our new open-source library: https://github.com/makepath/mapshader

philippjfr:
I'd be okay with merging this but I'll make some minor code formatting comments.

datashader/core.py
```diff
@@ -1121,7 +1123,26 @@ def raster(self,
...
+        if np.isclose(bottom, self.y_range[0]) and \
+                np.isclose(top, self.y_range[1]) and \
+                np.size(yvals) == self.plot_height:
+            # To avoid floating point representation error,
+            # do not recompute y coords if same y_range and same plot_height
+            ys = yvals
+        else:
+            _, ys = compute_coords(self.plot_width, self.plot_height,
+                                   self.x_range, self.y_range, res)
+
```
philippjfr, line 1145: Just to avoid calling `compute_coords` twice:

```suggestion
        # To avoid floating point representation error,
        # do not recompute x coords if same x_range and same plot_width
        close_x = np.isclose([left, right], self.x_range).all() and np.size(xvals) == self.plot_width
        close_y = np.isclose([bottom, top], self.y_range).all() and np.size(yvals) == self.plot_height
        if close_x and close_y:
            xs, ys = xvals, yvals
        else:
            xs, ys = compute_coords(self.plot_width, self.plot_height,
                                    self.x_range, self.y_range, res)
            xs = xvals if close_x else xs
            ys = yvals if close_y else ys
```

thuydotm: Thanks @philippjfr. Changed as suggested.

thuydotm:
This has been rebased on master and all tests passed. Can we expect this to be merged? Would be great to have this fixed as we've been using `Canvas.raster()` a lot.

ianthomas23:
It looks like there is mostly agreement on merging this, with the concern about wider floating-point comparison issues for coordinates. I have come across the latter recently and am happy to contribute some time to help deal with it as it is a blocker for some of my work. It would be helpful for this to be merged first so that future work can build on top of it rather than reinventing it.

jbednar:
Ok, that sounds like a good bargain to me. :-)

ianthomas23:
Thanks @jbednar!

PR #960, opened 2020-09-06
Unpin pyct 

ceball:
0.4.6 is the latest release anyway
Bug in latest dev release (affecting test suite) is now fixed (https://github.com/pyviz-dev/pyct/issues/85).


ceball:
The technically correct thing to do would probably be to exclude 0.4.6, but that seems overly complicated (and might e.g. cause installation problems for people). 

As long as pyct 0.4.7 or above makes it to conda-forge and main before the next datashader release, unpinning seems ok to me.

Issue #970, opened 2020-11-26
ndarray TypeError when visualising Dask Array


peterroelants:
When trying to visualize a Dask array with `raster` and `shade` I get a `TypeError: data must be an ndarray` error in the `eq_hist` method. From the DataShader documentation I expected Dask arrays to be supported in DataShader.

I have provided a minimal notebook at https://gist.github.com/peterroelants/1d77e09bd05cc55c240bc11983e2a0c4 to reproduce the error. 

#### ALL software version info
```
Python implementation: CPython
Python version       : 3.8.6
IPython version      : 7.19.0

Compiler    : GCC 7.5.0
OS          : Linux
Release     : 5.4.0-54-generic
Machine     : x86_64
Processor   : x86_64
CPU cores   : 4
Architecture: 64bit

datashader: 0.11.1
sys       : 3.8.6 | packaged by conda-forge | (default, Oct  7 2020, 19:08:05) 
[GCC 7.5.0]
numpy     : 1.19.4
xarray    : 0.16.1
dask      : 2.30.0
```

#### Description of expected behavior and the observed behavior

I expect `shade` being able to visualise a Dask array.

#### Complete, minimal, self-contained example code that reproduces the issue

https://gist.github.com/peterroelants/1d77e09bd05cc55c240bc11983e2a0c4

#### Stack traceback and/or browser JavaScript console output

```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-6-a01bef0eda6c> in <module>
      1 canvas = datashader.Canvas(plot_width=900, plot_height=400)
      2 datashader.transfer_functions.Images(
----> 3     datashader.transfer_functions.shade(
      4         canvas.raster(data_da, agg=datashader.reductions.mean('z'))
      5     )

~/miniconda3/envs/msi/lib/python3.8/site-packages/datashader/transfer_functions/__init__.py in shade(agg, cmap, color_key, how, alpha, min_alpha, span, name, color_baseline)
    509 
    510     if agg.ndim == 2:
--> 511         return _interpolate(agg, cmap, how, alpha, span, min_alpha, name)
    512     elif agg.ndim == 3:
    513         return _colorize(agg, color_key, how, alpha, span, min_alpha, name, color_baseline)

~/miniconda3/envs/msi/lib/python3.8/site-packages/datashader/transfer_functions/__init__.py in _interpolate(agg, cmap, how, alpha, span, min_alpha, name)
[...]

PR #971, opened 2020-11-27
XArray + DaskArray working and tested for shade

peterroelants:
Using `tf.shade` was throwing a `TypeError` when using XArray + DaskArray as documented in https://github.com/holoviz/datashader/issues/970

I did a first stab at fixing this by first adapting the transfer functions tests by testing Xarray + DaskArray, and then fixing Dask support in `transfer_functions`. I think the key is to call `.compute()` on the Dask arrays to force them to a numpy array before any shading interpolation is done.

Tests in `datashader/tests/test_transfer_functions.py` pass with the same warnings as before. I also tested the XArray + DaskArray support for Datashader and Holoviews in https://gist.github.com/peterroelants/0624834713a1388c7f57d3cafd9b800b

I would love any feedback an pointers to get this fix in Datashader in a consistent way.

### DIFF
--- a/datashader/tests/test_transfer_functions.py
+++ b/datashader/tests/test_transfer_functions.py
@@ -4,6 +4,7 @@ from io import BytesIO
 
 import numpy as np
 import xarray as xr
+import dask.array as da
 import PIL
 import pytest
 from collections import OrderedDict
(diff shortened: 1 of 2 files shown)


datashader/tests/test_transfer_functions.py
```diff
@@ -28,14 +29,22 @@ def build_agg(array_module=np):
...
+
+
 try:
     import cupy
-    aggs = [build_agg(np), build_agg(cupy)]
-    arrays = [np.array, cupy.array]
-except ImportError:
+    aggs = [build_agg(np), build_agg(cupy), build_agg_dask()]
+    arrays = [np.array, cupy.array, create_dask_array]
+except (ImportError, ModuleNotFoundError):
```
jbednar, line 44:
```suggestion
except ImportError:
```

Thanks! This PR looks beautiful. It's failing a linting test ("./datashader/tests/test_transfer_functions.py:44:22: F821 undefined name 'ModuleNotFoundError'"), but that seems to be just that the linter is confused, because ordinarily exceptions don't need to be imported explicitly.

In any case, is there some reason for testing against ModuleNotFoundError specifically? According to https://docs.python.org/3/library/exceptions.html#ModuleNotFoundError, ModuleNotFoundError is a subclass of ImportError and thus should already be caught by checking for ImportError. So unless I'm confused, the PR should be ready to merge if you just apply the suggestion here to revert that change. But I'll leave it for you to apply the suggestion, in case I'm confused about it.

peterroelants: While debugging I noticed that a `ModuleNotFoundError` was thrown when importing cupy. However, you are right, it seems not be needed and I updated the PR to revert this.

jbednar (approved):

Issue #976, opened 2020-12-11
Dead kernel in 3_Interactive.ipynb


jbednar:
As originally reported in https://github.com/holoviz/datashader/pull/939, the notebook getting_started/3_Interactive.ipynb sometimes fails to execute using either current Datashader master or the 0.11.1 release, reporting that the kernel has died.  I have not yet been able to reduce it to a minimal example, but here's at least a cut-down one. First, create `segfault.ipynb`, containing:

```python
import pandas as pd, numpy as np, holoviews as hv
import holoviews.operation.datashader as hd
from holoviews.streams import RangeXY
hv.extension("bokeh", "matplotlib") 

num=100000
np.random.seed(1)

dists = {cat: pd.DataFrame(dict([('x',np.random.normal(x,s,num)), 
                                 ('y',np.random.normal(y,s,num)), 
                                 ('val',val), 
                                 ('cat',cat)]))      
         for x,  y,  s,  val, cat in 
         [(  2,  2, 0.03, 10, "d1"), 
          (  2, -2, 0.10, 20, "d2"), 
          ( -2, -2, 0.50, 30, "d3"), 
          ( -2,  2, 1.00, 40, "d4"), 
          (  0,  0, 3.00, 50, "d5")] }

df = pd.concat(dists,ignore_index=True)
df["cat"]=df["cat"].astype("category")

points = hv.Points(df.sample(10000))

pts = hd.datashade(points, width=400, height=400)

quadmesh =  hv.QuadMesh(hd.aggregate(points, width=10, height=10, dynamic=False)) \
            .opts(tools=['hover'], alpha=0, hover_alpha=0.2)

dynamic = hv.util.Dynamic(hd.aggregate(points, width=10, height=10, streams=[RangeXY]), 
                          operation=hv.QuadMesh) \
          .opts(tools=['hover'], alpha=0, hover_alpha=0.2)

(pts * quadmesh).relabel("Fixed hover") + (pts * dynamic).relabel("Dynamic hover")
```

Second, create `segfault.py`, containing:

```python
from nbconvert.preprocessors import ExecutePreprocessor
import nbformat
import io

for i in range(100):
    print(i)
    nb = io.open('./segfault.ipynb', encoding='utf-8')
    notebook = nbformat.read(nb, as_version=4)
    ep = ExecutePreprocessor(allow_errors=False)
    output = ep.preprocess(notebook, {})
```

[...]

PR #981, opened 2020-12-23
Fix out-of-bounds error on Points aggregation

philippjfr:
- [x] Fix for https://github.com/holoviz/datashader/issues/976

The problem here is floating point comparison issues, in particular on this line:

```python
xi, yi = (xx - 1 if x == xmax else xx,
          yy - 1 if y == ymax else yy)
```

where `y=9.921649019012754` and `ymax=9.921649019012756` this result in an out-of-bounds index (`yi`) being used during the aggregation step:

```python
agg[y, x] += 1
```

This PR isn't a particularly principled fix but simply allows for a 1e15 tolerance. Any other suggestions (without high computational cost welcome), one other option is to check against the actual shape of the array.

### DIFF
--- a/datashader/glyphs/points.py
+++ b/datashader/glyphs/points.py
@@ -163,7 +163,8 @@ class Point(_PointLike):
                 xx = int(x_mapper(x) * sx + tx)
                 yy = int(y_mapper(y) * sy + ty)
 
-                xi, yi = min(xx, xxmax-1), min(yy, yymax-1)
+                xi, yi = (xx-1 if xx == xxmax else xx,
+                          yy-1 if yy == yymax else yy)
                 append(i, xi, yi, *aggs_and_cols)
 
         @ngjit
--- a/datashader/glyphs/points.py
+++ b/datashader/glyphs/points.py
@@ -163,8 +163,8 @@ class Point(_PointLike):
                 xx = int(x_mapper(x) * sx + tx)
                 yy = int(y_mapper(y) * sy + ty)
 
-                xi, yi = (xx-1 if xx == xxmax else xx,
-                          yy-1 if yy == yymax else yy)
+                xi, yi = (xxmax-1 if xx >= xxmax else xx,
+                          yymax-1 if yy >= yymax else yy)
                 append(i, xi, yi, *aggs_and_cols)
 
         @ngjit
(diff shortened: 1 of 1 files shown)


datashader/glyphs/points.py
```diff
@@ -159,9 +159,9 @@ def _perform_extend_points(i, sx, tx, sy, ty, xmin, xmax, ymin, ymax, xs, ys, *a
             if (xmin <= x <= xmax) and (ymin <= y <= ymax):
                 xx = int(x_mapper(x) * sx + tx)
                 yy = int(y_mapper(y) * sy + ty)
-                xi, yi = (xx - 1 if x == xmax else xx,
-                          yy - 1 if y == ymax else yy)
 
+                xi, yi = (xx - 1 if x >= (xmax-1e-15) else xx,
+                          yy - 1 if y >= (ymax-1e-15) else yy)
```
jbednar, line 164: Fundamentally, we aren't meant to get to this location unless x<=xmax and y<=ymax, so can we be more direct about this?  E.g. can we calculate the integer xx and yy values that are out of bounds, and then test against xx and yy rather than x and y with an epsilon?  That seems like a more direct way to enforce that the results are in bounds, with no arbitrary constants.

jbednar: e.g. 

```
xi, yi = (xxmax - 1 if xx >= xxmax else xx,
          yymax - 1 if yy >= yymax else yy)
```

Seems like this will work as long as we can calculate xxmax and yymax.

jbednar: e.g. 

```
xi, yi = (xxmax - 1 if xx >= xxmax else xx,
          yymax - 1 if yy >= yymax else yy)
```

Seems like this will work as long as we can calculate xxmax and yymax.

philippjfr: Definitely slower which is why I didn't suggest it. We could compute that once and pass it in as arguments, just seems awkward all around. Alternatively we could check the shape of the first aggregate array in aggs_and_cols, but that would probably break with the argument expansion ast manipulation code.

jbednar: It does seem awkward, but an epsilon doesn't seem like a viable alternative unless it were a quantity expressed in terms of the fundamental floating-point precision, and not just an arbitrary number that will affect data ranges of different magnitudes differently.

philippjfr: 1e-15 is one order of magnitude larger than sys.float_info.epsilon. That said I'm not actually arguing this should be merged. I think we will just have to pass in the array bounds.

jbednar: Passing in the array bounds is ugly but I don't see any viable alternative either.

philippjfr: Done now.

jbednar (approved):
That wasn't so bad! :-) Thanks.

PR #986, opened 2021-02-04
dsshow: fix error with dictionary color_key when creating a legend

nvictus:
`get_legend_elements` was assuming that `color_key` is always a list and would fail if `color_key` were a dict. This PR fixes that.


jbednar:
Thanks!

Issue #990, opened 2021-03-02
error with `Canvas().raster` on xarray 0.17


darribas:
#### ALL software version info

The full list is available [here](https://github.com/darribas/gds_env/raw/master/gds_py/stack_py.txt), the gist is:

- Python 3.8
- `xarray` 0.17.0
- `datashader` 0.12.0

#### Description of expected behavior and the observed behavior

I expect `Canvas().raster` to return a `DataArray` object, instead it returns an error.

#### Complete, minimal, self-contained example code that reproduces the issue

Adapted from [your documentation](https://datashader.org/user_guide/Grids.html)

```python
from datashader import transfer_functions as tf, reductions as rd
import numpy as np, datashader as ds, xarray as xr

def f(x,y):
    return np.cos((x**2+y**2)**2)

def sample(fn, n=50, range_=(0.0,2.4)):
    xs = ys = np.linspace(range_[0], range_[1], n)
    x,y = np.meshgrid(xs, ys)
    z   = fn(x,y)
    return xr.DataArray(z, coords=[('y',ys), ('x',xs)])

da = sample(f)
ds.Canvas().raster(da, interpolate='linear')
```

#### Stack traceback and/or browser JavaScript console output

```python
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-3-7e8b8b432665> in <module>
     12 
     13 da = sample(f)
---> 14 ds.Canvas().raster(da, interpolate='linear')

/opt/conda/lib/python3.8/site-packages/datashader/core.py in raster(self, source, layer, upsample_method, downsample_method, nan_value, agg, interpolate, chunksize, max_mem)
   1131         dims = [ydim, xdim]
   1132         attrs = dict(res=res[0])
-> 1133         if source._file_obj is not None and hasattr(source._file_obj, 'nodata'):
   1134             attrs['nodata'] = source._file_obj.nodata
   1135 

/opt/conda/lib/python3.8/site-packages/xarray/core/common.py in __getattr__(self, name)
    237                 with suppress(KeyError):
    238                     return source[name]
--> 239         raise AttributeError(
    240             "{!r} object has no attribute {!r}".format(type(self).__name__, name)
    241         )

AttributeError: 'DataArray' object has no attribute '_file_obj'
```

I *think* this is related to [`#4809`](https://github.com/pydata/xarray/pull/4809).

jbednar:
[...]

PR #991, opened 2021-03-05
Fix nodata handling for xarray raster files

jbednar:
Fix for https://github.com/holoviz/datashader/issues/990 and https://github.com/holoviz/hvplot/issues/563. 

Previously, Datashader Canvas.raster was accessing the underlying `_file_obj` attribute of a provided Xarray DataArray raster to find a 'nodata' (to be treated as NaN) value. `_file_obj` was clearly marked private, so accessing it directly was presumably always a bad idea.  This usage broke with xarray 0.17 when the `_file_obj` attribute was hidden inside a closure.  Instead, what other libraries (e.g. [rioxarray](https://corteva.github.io/rioxarray/stable/getting_started/nodata_management.html)) seem to do is look for various "nodata" attributes on the xarray object, presumably each corresponding to some convention or the output of some file reader or file writer. 

With this PR, Datashader now checks the DataArray attrs for ['_FillValue', 'missing_value', 'fill_value', 'nodata', 'NODATA'] (in that order), taking a single scalar value from the first one it finds. If it fails to find any of them, it then looks for 'nodatavals', and if that's a list and has at least one element it takes the first in the list. Otherwise, there is no "nodata" value available.

These attr names were suggested by various people at the linked issues and in the rioxarray docs, and I do not have access to any files actually defining or respecting such conventions. So I haven't tested that these actually work "in the wild", but at least if I hack in those attribute values into an already-constructed DataArray it seems to work:

```python
from datashader import transfer_functions as tf, reductions as rd
import numpy as np, datashader as ds, xarray as xr

def f(x,y):
    return np.cos((x**2+y**2)**2)

def sample(fn, n=50, range_=(0.0,2.4)):
    xs = ys = np.linspace(range_[0], range_[1], n)
    x,y = np.meshgrid(xs, ys)
    z   = fn(x,y)
    return xr.DataArray(z, coords=[('y',ys), ('x',xs)])

da = sample(f)
da.attrs['missing_value'] = 1
out = ds.Canvas().raster(da, interpolate='linear')
out.attrs

{'res': 0.04897959183673469, 'nodata': 1}
```

### DIFF
--- a/datashader/core.py
+++ b/datashader/core.py
@@ -1130,8 +1130,19 @@ x- and y-coordinate arrays must have 1 or 2 dimensions.
         coords = {xdim: xs, ydim: ys}
         dims = [ydim, xdim]
         attrs = dict(res=res[0])
-        if source._file_obj is not None and hasattr(source._file_obj, 'nodata'):
-            attrs['nodata'] = source._file_obj.nodata
+
+        # Find nodata value if available in any of the common conventional locations
+        # See https://corteva.github.io/rioxarray/stable/getting_started/nodata_management.html
+        # and https://github.com/holoviz/datashader/issues/990
+        for a in ['_FillValue', 'missing_value', 'fill_value', 'nodata', 'NODATA']:
+            if a in source.attrs:
+                attrs['nodata'] = source.attrs[a]
+                break
+        if 'nodata' not in attrs:
+            try:
+                attrs['nodata'] = source.attrs['nodatavals'][0]
+            except:
+                pass
 
         # Handle DataArray with layers
         if data.ndim == 3:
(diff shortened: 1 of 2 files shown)


jbednar:
Merging now that tests pass due to @philippjfr's fix in #993. Should be able to make a new release now; will see if I can fix the broken docs today and release later today either way to help people avoid problems for now.

michaelaye:
it's missing `nodatavals` it seems?

philippjfr:
@michaelaye It's right here: https://github.com/holoviz/datashader/pull/991/files#diff-98ed557bc2b35ad5f79c51aa106a68ea68ce2aa129aa46e6d2f4da997c7f5908R1143

jbednar:
Right; `nodatavals` is the only one that's defined as a sequence that I know of, but if any of the others might be a sequence, then we'd need to handle that.

PR #1001, opened 2021-04-11
Fix dynspread not to count already-spread pixels and respect array bnds

### LINKED ISSUE #1000: Improving dynspread
Given Datashader's model of a point as a mathematically ideal, infinitesimally small location, Datashader relies on spreading (image dilation) for making isolated datapoints visible on high-resolution displays. However, spreading inherently destroys data for non-isolated datapoints, which means that the correct (useful) amount of spread (say, in the range 0 to `max_px`) depends on zoom level and the data distribution. When zoomed out so that data points are clustered near each other, we want zero spreading so that we can see the distribution accurately, while when zoomed in enough that data points are fully spread out, we want the full amount of spreading to make individual dots fully visible. 

The dynspread command was meant to achieve that, using a metric based on how many data points have  non-empty neighbors. dynspread iteratively increases the pixel spreading until the metric exceeds a fixed threshold, and that's how much spreading is done. See the docs for `max_px` and `threshold`:

image

However, dynspread has never worked very well, and it looks to me like the [logic in it](https://github.com/holoviz/datashader/blob/master/datashader/transfer_functions/__init__.py#L714) is fundamentally flawed, in that previous iterations of spreading fill in a widening circle around each point, and then _the_ _pixels_ _inside_ _the_ _circle_ erroneously count as neighboring non-empty pixels.  The result is that even widely separated actual data points are treated as if they have overlap, completely inappropriately. Here I've printed the level of `px` being tested, the specified threshold, and the computed metric for that `px` amount of spreading:

```
import pandas as pd, holoviews as hv, holoviews.operation.datashader as hd
hv.extension('bokeh')

df = pd.DataFrame(dict(x=[1,2,3,4,5], y=[5,0,8,2,3]))
img = hd.rasterize(hv.Points(df), height=50, width=50, dynamic=False)
img
```
image
image
image
image
image
image
image

Notice how there is in fact _zero_ adjacency between the rendered datapoints all the way up to a max_px of 4, and even 5 has no actual overlap even though it does have a couple of adjacent pixels. `px` above 5 does have overlap, but in fact I wouldn't have expected such overlap to reach the default threshold of 0.5, as intended by the docstring, and so for this very sparse dynspread plot I would have expected that it used `max_px` spreading with no adjustments to parameters by the user, so that the result is equivalent to using `spread()`:

image

Can anyone suggest an alternative dynspread metric where only the overlapping pixels count?  It still needs to correctly return `px=0` for a densely packed plot, while returning `px=max_px` for widely separated points like in this plot, and preferably smoothly varying in between in a reasonable way.

jbednar:
Fixes #1000. As described in #1000, dynspread was not working as documented or designed, because it was iteratively measuring the density of already-spread pixels, which produced large dense regions that were indistinguishable from dense regions of datapoints. Also, it turns out, the density calculations were reading outside the bounds of the array, which is dangerous and likely to give incorrect readings for density, even though it does not involve _writing_ outside the array. 

Fixed both problems by (1) respecting the bounds of the array when computing density, and (2) iteratively measuring density using an increasing neighborhood (as before) but no longer with actual spreading for each interative step (unlike before). I.e., instead of spreading then checking to see if the density metric is exceeded, simply tries increasingly large values of a neighborhood radius to count how many pixels have neighbors within this radius, and stopping when that normalized number reaches the specified threshold. The neighborhood search is now twice as large as the spreading amount being tested, since unlike actual spreading each datapoint has to check distance to the neighboring datapoint's center, not edge, but it no longer has to do actual spreading, so the speed may not decrease much. 

In any case, the behavior is now vastly more appropriate, and is approximately how I first envisioned dynspread working, smoothly switching to larger point sizes (even for large `max_px` values of e.g. 100) on zooming in. The default threshold of 0.5 actually still seems about right to me, so no need to update that. The `max_px` default of 3 now seems low; 5 seems more visible and clear once zooming in. But I'll leave the default value unchanged here since really it's the default value in the plotting library (e.g. HoloViews hv.operation.datashader.dynspread) that affects interactive use.

Here's an example using the datashader_dashboard, with `max_px=10` (fairly big!):

https://user-images.githubusercontent.com/1695496/114293108-0ca77500-9a59-11eb-86ed-9679603e7fd1.mp4

### DIFF
--- a/datashader/tests/test_transfer_functions.py
+++ b/datashader/tests/test_transfer_functions.py
@@ -948,7 +948,7 @@ def test_rgb_dynspread():
     coords = [np.arange(5), np.arange(5)]
     img = tf.Image(data, coords=coords, dims=dims)
     assert tf.dynspread(img).equals(tf.spread(img, 1))
-    assert tf.dynspread(img, threshold=0.9).equals(tf.spread(img, 2))
+    assert tf.dynspread(img, threshold=0.9).equals(tf.spread(img, 3))
     assert tf.dynspread(img, threshold=0).equals(img)
     assert tf.dynspread(img, max_px=0).equals(img)
 
@@ -964,7 +964,7 @@ def test_array_dynspread():
     coords = [np.arange(5), np.arange(5)]
     arr = xr.DataArray(data, coords=coords, dims=dims)
     assert tf.dynspread(arr).equals(tf.spread(arr, 1))
-    assert tf.dynspread(arr, threshold=0.9).equals(tf.spread(arr, 2))
+    assert tf.dynspread(arr, threshold=0.9).equals(tf.spread(arr, 3))
     assert tf.dynspread(arr, threshold=0).equals(arr)
     assert tf.dynspread(arr, max_px=0).equals(arr)
 
@@ -996,7 +996,7 @@ def test_categorical_dynspread():
     arr = xr.DataArray(data, coords=coords + [['a', 'b', 'c']],
                        dims=dims + ['cat'])
     assert tf.dynspread(arr).equals(tf.spread(arr, 1))
-    assert tf.dynspread(arr, threshold=0.9).equals(tf.spread(arr, 2))
+    assert tf.dynspread(arr, threshold=0.9).equals(tf.spread(arr, 3))
     assert tf.dynspread(arr, threshold=0).equals(arr)
     assert tf.dynspread(arr, max_px=0).equals(arr)
 
--- a/datashader/tests/test_transfer_functions.py
+++ b/datashader/tests/test_transfer_functions.py
@@ -912,60 +912,78 @@ def test_rgb_density():
     assert tf._rgb_density(data) == 1.0
     data = np.zeros((4, 4), dtype='uint32')
     assert tf._rgb_density(data) == np.inf
-    data[2, 2] = b
+    data[3, 3] = b
     assert tf._rgb_density(data) == 0
-    data[2, 1] = data[1, 2] = data[1, 1] = b
-    assert np.allclose(tf._rgb_density(data), 3./8.)
+    data[2, 0] = data[0, 2] = data[1, 1] = b
+    assert np.allclose(tf._rgb_density(data), 0.75)
+    assert np.allclose(tf._rgb_density(data, 3), 1)
 
 def test_int_array_density():
     data = np.ones((4, 4), dtype='uint32')
     assert tf._array_density(data, float_type=False) == 1.0
     data = np.zeros((4, 4), dtype='uint32')
     assert tf._array_density(data, float_type=False) == np.inf
-    data[2, 2] = 1
+    data[3, 3] = 1
     assert tf._array_density(data, float_type=False) == 0
-    data[2, 1] = data[1, 2] = data[1, 1] = 1
-    assert np.allclose(tf._array_density(data, float_type=False), 3./8.)
+    data[2, 0] = data[0, 2] = data[1, 1] = 1
+    assert np.allclose(tf._array_density(data, float_type=False), 0.75)
+    assert np.allclose(tf._array_density(data, float_type=False, px=3), 1)
 
+    
 def test_float_array_density():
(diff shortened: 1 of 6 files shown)


philippjfr:
Nice! My main question I have is if you looked at possible performance implications?

>Also, it turns out, the density calculations were reading outside the bounds of the array, which is dangerous and likely to give incorrect readings for density

😬

jbednar:
I have not done any benchmarking, but I wasn't able to spot performance differences in my informal testing.  It gets slower as the image size increases and as max_px increases (but only for images sparse enough to approach max_px).

jlstevens:
Very nice! I'll give this PR a go tomorrow morning and review.

jlstevens:
I've always felt that a single dynamic spreading approach for multiple glyphs (e.g points versus lines)  would have such tradeoffs. Imho what lines really need is a dynamic thickness instead of a convolution type approach.

At the holoviews level we could know which glyph is being rasterized and choose different code paths that way. At the datashader level it would be up to the user to choose a dynspread approach appropriate to what is being rasterized.

jbednar:
To-do items: Ok, after discussion with @jlstevens and @philippjfr, we agreed:

1. For points plots, this new algorithm is a clear improvement and behaves very reasonably, and should go in. There might be other algorithms that would work as well, but this one is very simple to reason about and matches what we generally want dynspread to do, when operating under its current constraints.
2. For line plots, it's more complicated:
   a. The new algorithm does not ever do any spreading for normal line/curve/area plots, because every point in such plots of spatially extended items always has a neighbor.
   b. In nearly every case we could think of, `spread(px=1)` is actually what we'd want to see for a line plot, and that has always behaved well.
   c. `dynspread` previously typically ended up spreading a line plot a little, nearly being equivalent to spread(px=1), whereas now line plots will no longer have spreading with `dynspread`. To re-enable spreading (previously only happening as a side effect of a broken algorithm), those examples will need updating.
   d. It's conceivable that we could write a neighborhood-based dynspread function for use with lines that looks at the fraction of points filled out of the neighborhood radius, and this could let a line plot use no spreading when on balance it's a dense tangle, and spreading when zoomed in. Making such an algorithm would be a research project (probably not too hard?), and normalizing it to behave similarly with respect to the threshold value as the current algorithm would be difficult, so it seems like it could be a fully separate function if we do decide we need it. I propose that we not do that work, and just see if a fixed spread value is good enough; seems more of a priority to make an antialiased line with thickness than to have a special dynspread for lines.
3. Should we offer the old dynspread function somehow for backwards compatibility? Our conclusion is that preserving the old behavior is not worth the complexity, but we'll need some text about this issue in the release notes regardless.
   a. Cons:
      1. The old one was buggy in several ways and just did not work well.
      2. Supporting both is more complex and messy.
      3. Supporting both would e.g. require explaining to users to do something like `datashader.transfer_functions.dynspread = datashader.transfer_functions.dynspread_old`, which is not substantially easier than telling them to change `dynspread` to `spread(px=1)`. 
   b. Pros:
      1. People who were relying on the previous buggy behavior would more easily be able to return to it.
4. The new `dynspread` appears to work as well as it can given the underlying assumptions, but see #1012 for a proposal for how to eliminate the sharp jumps in spreading sizes currently required.


To-do list before merging: (Jim)
- [ ] Update the Datashader tests that were verifying the old buggy behavior.
- [ ] Search the datashader docs for `dynspread` used on line plots, and update those to use `spread(px=1)` instead. 
- [ ] Rename `density` to something more accurate. neighbor_fraction?
- [ ] Write a release notes entry about dynspread updates.

To-do after merging:
- [ ] Search related sites like holoviews for `dynspread` used on line plots, and update those to use `spread(px=1)` instead.

To-do after releasing datashader 0.13:
- [ ] Update examples.pyviz and holoviews websites to use Datashader 0.13 so that their dynspread behavior improves.

jbednar:
Tests are passing on python 3.6, but failing on later versions because of https://github.com/holoviz/spatialpandas/issues/68 and possibly other issues with related libraries. Merging this now, but still need to address the downstream issues.

PR #1008, opened 2021-05-21
Fix nan assignment to integer array in RaggedArray

philippjfr:
Fixes tests

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -26,6 +26,7 @@ install_requires = [
 ]
 
 examples = [
+    'distributed',
     'holoviews >=1.10.0',
     'scikit-image',
     'bokeh',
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -45,13 +43,13 @@
...
         'fastparquet >=0.1.6',  # optional dependency
         'holoviews >=1.10.0',
+        'bokeh',
         'pyarrow',
         'netcdf4',
         'twine',   # required for pip packaging
         'rfc3986', # required by twine
         'keyring', # required by twine
-        'spatialpandas',
-        'ipympl'
```
jbednar, line 54: Why was this removed? Isn't it needed for the datashader examples now?

philippjfr: ipympl is not enabled by default no and I was trying to improve solve time.

PR #1016, opened 2021-08-02
Use cupy.interp when available

philippjfr:
For now we still provide backward compatibility for older versions of cupy. This also eliminates a warning about transferring an numpy array to the GPU by explicitly casting it with cupy.asarray.

### DIFF
--- a/datashader/transfer_functions/_cuda_utils.py
+++ b/datashader/transfer_functions/_cuda_utils.py
@@ -111,14 +111,16 @@ def masked_clip_2d_kernel(data, mask, lower, upper):
         cuda_atomic_nanmin(data, (i, j), upper)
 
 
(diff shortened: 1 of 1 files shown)


maximlt:
@philippjfr could you check out the failing tests? Thanks!

philippjfr:
Logs appear to have expired and it **very** doubtful the failures were related to this PR since we don't/can't test GPU codepaths atm.

Hoxbro:
I didn't see this PR but have been working on some more cuda support here https://github.com/Hoxbro/datashader/pull/1, also including interp. 

I didn't want to add it before CI was fixed, but I can see a new release is on the horizon, and because I'm not at my computer until next week, I will just “announce” it here.

PR #1020, opened 2021-09-06
Categorical color_key for 2D (not stacked) aggregates

thuydotm:
This PR adds color key support for 2D categorical data arrays. The approach is similar to the 3D case. The RGB channels are computed by the colors in the provided `color_key`. The A channel is fixed as the given `alpha` value for all categories specified in the `color_key`, and to zero otherwise.


ianthomas23:
@thuydotm Now that the CI is fixed, you can rebase this on master and hopefully the tests will pass.

thuydotm:
Thanks @ianthomas23, I've rebased on master. Let's see if the tests pass.

jbednar (approved):
Thanks, and sorry for the long delay in getting the tests running! The code here looks reasonable to me and does seem useful. I'll need to expand the descriptions of 2D and 3D categorical aggregates in the docs after merging this, but because of this being a fork, I won't do it in place. I think people will need more explanation of what a 3D categorical is (i.e., a stack of separate values by category that can be mixed to get a final color) and what a 2D categorical is (a single category per pixel, with no mixing). 

Can you please provide an example in your "Colormapping 2D categorical data" section that demonstrates the support for non-1.0 alpha?  I see that in the code, but am not sure how to get that when the aggregate can have only a single value to denote the category.

thuydotm:
Thanks for the comments, I'll update the PR and let you know!

thuydotm:
Thanks for mentioning how alpha channel is computed. I updated the docs and made changes to the alpha calculation. There's no mixing in a 2D categorical array so the A channel should always be fixed as the given `alpha` value for all categories specified in the `color_key`, and to zero otherwise.

PR #1024, opened 2021-10-16
MapboxTilesRenderer

hokieg3n1us:
Implement MapboxTilesRenderer, which adds support for outputting TMS tilesets as a mbtiles file (a SQLite database following the [MBTiles 1.3 schema](https://github.com/mapbox/mbtiles-spec/blob/master/1.3/spec.md). This file can then be easily used in GIS clients, such as QGIS, though the primary motivation was creating an output format that could easily be used by GeoServer (using the MBTiles community extension). 

Example using data from ACLED (Armed Conflict Location & Event Data):

```
import dask.dataframe
import datashader
from datashader.tiles import render_tiles
from datashader.utils import lnglat_to_meters

import colorcet


def _get_extents():
    return df.x.min().compute(), df.y.min().compute(), df.x.max().compute(), df.y.max().compute()


def _load_data_func(x_range, y_range):
    return df.loc[df.x.between(*x_range) & df.y.between(*y_range)]


def _rasterize_func(df, x_range, y_range, height, width):
    cvs = datashader.Canvas(x_range=x_range, y_range=y_range,
                            plot_height=height, plot_width=width)
    return agg


def _shader_func(agg, span=None):
    img = datashader.tf.dynspread(datashader.tf.shade(agg, cmap=colorcet.fire))
    return img


# Can be utilized to customize image with watermark, etc.
def _post_render_func(img, **kwargs):
    return img


if __name__ == '__main__':
    df = dask.dataframe.read_csv('ACLED.csv', usecols=['longitude', 'latitude']).persist()

    # Coordinates should be in Mercator format.
    df['x'], df['y'] = lnglat_to_meters(df['longitude'], df['latitude'])

    df.drop(['longitude', 'latitude'], axis=1)

    min_zoom = 0
    max_zoom = 8
    output_path = 'output/ACLED.mbtiles'

    render_tiles(_get_extents(),
                 range(min_zoom, max_zoom + 1),
                 load_data_func=_load_data_func,
                 rasterize_func=_rasterize_func,
                 shader_func=_shader_func,
                 post_render_func=_post_render_func,
                 output_path=output_path, num_workers=4)
```

[ACLED.mbtiles](https://github.com/hokieg3n1us/datashader_output/blob/main/ACLED.mbtiles) loaded into QGIS:

image

The render_tiles function now has two big changes to it's behavior. 

1. It validates the output_path immediately. If the output_path is a directory, it will create it. If it's a file path that ends with mbtiles, it'll create the directory structure, then setup the tables & metadata. This should only be done a single time, so it's functionality is exposed by a static method in the MapboxTilesRenderer.
2. It exposes the num_workers used by the Dask Bag, so that the SQLite file doesn't lock due to extremely high concurrency.


hokieg3n1us:
Included an additional capability for out-of-core processing. An optional parameter, local_cache_path, can be provided. If provided, the aggregates generated by Datashader for the super tiles will be persisted to that local cache as a NetCDF file, instead of keeping all these intermediate results in memory. These aggregates will then be individually loaded by Dask workers to generate the individual tiles.  This allows tile generation for larger datasets at much higher zoom levels, being only dependent on your disk space.

PR #1033, opened 2021-12-15
Drop Python 2 compat code and bump support to >=3.7

maximlt:
* Run coverage in CI
* Add ability to manually trigger the tests and docs workflows
* Attempt to run the CI on Python 3.9

TODOs
- [x] Update the versions supported by datashader in the docs
- [x] Figure out which was the last version that supported Python 2.7
- [x] Add a badge like on param's readme to show the supported versions of Python


maximlt:
@jbednar I've tried to run the CI on Python 2.7 with a simple config (no examples test) but couldn't conda to solve during  the `develop_install` step (on ubuntu, on the other platforms it failed before that with some obscure reasons). With Philipp we've spent some time looking for the last version 'officially' supporting Python 2. We've found that the CI was running on Python 2 (to be accurate, just linting was being run) in the 0.11.x series. So this is the version you could mention in the release note if anyone is willing to run Datashader on Python 2.

In this PR I went ahead and removed the compatibility code that was there to support Python 2, and pinned datashader to Python >= 3.6 in its `setup.py`.

maximlt:
@jbednar this is ready for review.

jbednar:
Thanks, @maximlt ! Nice to get that cleaned up.

PR #1037, opened 2021-12-25
Switch to the pydata sphinx theme

maximlt:
WIP


maximlt:
Converting to Draft since the docs building is failing, the environment solving just hangs on.

maximlt:
Feedback from @jbednar 

- [ ] No TOC on the first page (how does that work on [[holoviews.org?](http://holoviews.org?)](http://holoviews.org/?](http://holoviews.org?))?)
- [ ] User guide has a redundant copy of the section links at the end of the page: https://pyviz-dev.github.io/datashader/user_guide/index.html
- [ ] The footer has two periods in a row in "Anaconda, Inc..".
- [ ] It would be good to put "© Copyright 2016-2022, Anaconda, Inc. Last updated on 2022-01-12." all on a single line to reduce the size of the footer.
- [ ] I think the Discourse link in the header should go directly to https://discourse.holoviz.org/c/datashader, not just to https://discourse.holoviz.org/ (and similarly for all our sites that have a corresponding section in the Discourse)
- [ ] Switching the theme like this loses the link to HoloViz (from the top right header). All the other links like that moved to the icons in the top right of the menu bar, so can we add such an icon for HoloViz there on all our sites to help link them with each other?

There are also a variety of issues with how we convey that this page is a Jupyter notebook and how to run it for full interactivity:
* The "Open this page in Binder" appears in the right sidebar, far from the interactivity warning that motivates why it would be important to open it in Binder.
* The interactivity warning appears at the bottom, and doesn't mention that Binder is an option.
* Immediately after the interactivity warning, the page says to "Right click to download this notebook from GitHub", which seems redundant and confusing.

Can we consolidate all that into something in the right sidebar, formatted something like:
* Page automatically generated from a Jupyter notebook. For full interactivity:
  * Open in Binder, or
  * Download to run locally

maximlt:
I've partially addressed the comments above.

> No TOC on the first page

This is by design in the pydata-sphinx-theme (for some reason we can't explain the TOC is on the landing page of HoloViews). The theme is evolving fast though, so let's keep track of what they're up to and see if it fits our needs. Also, in an ideal world, each project would have a proper landing page.

> Switching the theme like this loses the link to HoloViz

We're onto it. @jlstevens will create an SVG icon. One annoyance to note is that the icon file path must be declared as a relative path from the root, which makes it impossible to be shared by `nbsite` in its share `_static` folder. So it seems to be that as it stands the icon file will have to be added to each repo.

> There are also a variety of issues with how we convey that this page is a Jupyter notebook and how to run it for full interactivity

Yes this is annoying. I've found out that the interactivity warning could have both the download and the binder options (https://github.com/pyviz-dev/nbsite/blob/faead4763535e1a6a30e62ec1181b74814f0357e/nbsite/nbbuild.py#L47). The problem though is that it's logic is implemented in the notebook preprocessor of nbsite while I'd much rather have this logic implemented within a template. The thing is that I don't know if within a template we can know if the file source is actually a notebook or a rst/md file.

So either I leave things as they are, or I remove the `Open in Binder` link in favor of the interactivity warning that would also include a binder link.

An issue not mentioned above is the integration of the twitter timeline:
image
It works well on HoloViews because HoloViews index page is written in rst and the timeline is embedded within the rst, with hacks to embed some HTML to control the display. While Datashader's index page is totally written in a notebook and as such it's more difficult to embed anything within that page. I'd vote for dropping the twitter timeline here.

Last issue I noted, I've just tried rebuilding the docs and they failed with this cryptic error. I also saw a few other errors that were raised while running the notebooks (which don't seem to terminate the docs build).
```
writing output... [  4%] FAQ
writing output... [  9%] about
writing output... [ 13%] api
writing output... [ 18%] getting_started/Interactivity

Exception occurred:
  File "/usr/share/miniconda3/envs/test-environment/lib/python3.8/site-packages/pkg_resources/__init__.py", line 783, in resolve
    raise DistributionNotFound(req, requirers)
pkg_resources.DistributionNotFound: The 'black' distribution was not found and is required by the application
```

jbednar:
Cool! Looks great. Apart from the tests being broken, looks like https://pyviz-dev.github.io/datashader/topics/index.html has all the topics listed twice. The Time Series user guide page also shows a blue background on some plots right now, which appears to be a regression, but is presumably not related to the docs build. Not sure when that would have appeared.

maximlt:
https://pyviz-dev.github.io/datashader/topics/index.html no longer has the duplicate list of topics.

PR #1040, opened 2022-01-20
Update dask_xarray.py

zjzh:
refactoring code with var unpack which is more pythonic, concise, readable and efficient; how do think this change which has practical value？


jbednar:
Thanks for the suggestion, but I don't think it's worth making the code more obscure to save this one line of code.

Issue #1038, opened 2022-01-04
Differences in coordinates calculated in raster() and other Canvas functions.


ianthomas23:
This is another example of when x, y coordinates are calculated slightly differently and give the error `TypeError: ufunc 'over' not supported for the input types, and the inputs could not be safely coerced to any supported types according to the casting rule ''safe''`. This is related to, but not the same as, issue #958 and PR #959. Simple reproducer:

```python
import datashader as ds
import datashader.transfer_functions as tf
import pandas as pd
import xarray as xr

points = pd.DataFrame(data=dict(x=[2, 6, 8], y=[8, 6, 2]))

raster = xr.DataArray(data=[[0.0, 1.0], [2.0, 3.0]], dims=("y", "x"),
    coords=dict(x=[0, 10], y=[0, 10]), name="my name")

#w = h = 20  # This always works.
w = h = 25
canvas = ds.Canvas(w, h, x_range=(0, 10), y_range=(0, 10))

agg_points = canvas.points(points, x="x", y="y")

agg_raster = canvas.raster(raster)
#agg_raster = canvas.quadmesh(raster)  # This always works.

im_points = tf.shade(agg_points)
im_raster = tf.shade(agg_raster)

# xdiff values should all be zero.
print("xdiff", im_points.coords["x"].data - im_raster.coords["x"].data)

tf.stack(im_raster, im_points)
```
On my dev machine this gives
```
xdiff [ 0.00000000e+00 -1.11022302e-16 -2.22044605e-16 -2.22044605e-16
 -2.22044605e-16 -4.44089210e-16 -4.44089210e-16 -8.88178420e-16
 -8.88178420e-16 -8.88178420e-16 -8.88178420e-16 -1.77635684e-15
 -8.88178420e-16 -8.88178420e-16 -1.77635684e-15 -8.88178420e-16
 -1.77635684e-15 -1.77635684e-15 -8.88178420e-16 -1.77635684e-15
 -1.77635684e-15 -1.77635684e-15 -1.77635684e-15  0.00000000e+00
  0.00000000e+00]
Traceback (most recent call last):
  File "bug_simple2.py", line 27, in <module>
    tf.stack(im_raster, im_points)
  File "/home/iant/github/datashader/datashader/transfer_functions/__init__.py", line 137, in stack
    out = tz.reduce(tz.flip(op), [i.data for i in imgs])
[...]

PR #1042, opened 2022-02-01
Allow floating-point tolerances in tf.stack coords

ianthomas23:
Candidate fix for issue #1038. There is a wider discussion in issue #959.

The underlying problem is that `xr.DataArray`/`tf.Image` objects passed to `tf.stack` may have slightly different coordinates (e.g. 1e-15 for reproducer in #1038). When the images are stacked using
https://github.com/holoviz/datashader/blob/25578abde75c7fa28c6633b33cb8d8a1e433da67/datashader/transfer_functions/__init__.py#L135
xarray takes the union of these coordinates, making the coordinate shapes larger, and eventually we get the somewhat cryptic error
```
TypeError: ufunc 'over' not supported for the input types
```
I have opted for a really simple one line fix which is easy to understand but might be overly smart.  It is to use `xr.align(*imgs, ..., join="override")` which uses the coordinates of the first `img` and ignores the others provides they all have the same shape (which they always do, it is checked a few lines earlier). This kwarg option has been in xarray since August 2019 (v0.13.0) so it is well tested.

This change fixes the reproducer in issue #1038 and does not cause any existing tests to fail.

But it may be a bit general. Maybe it also needs a prior check that the coordinates are not only the same shape but are also close enough. We would have to put in a numerical tolerance for this, with the possibility of annoying those using very large or very small coordinates. And what would we do with coordinates that fail this "close enough" test? We shouldn't use the previous code as this gives the cryptic error. So the least bad options are
  1. Always use `join="override"` and state in the docs this is what we are going to do.
  2. Put in a numerical tolerance and raise some sort of "not close enough" exception.

### DIFF
--- a/datashader/tests/test_transfer_functions.py
+++ b/datashader/tests/test_transfer_functions.py
@@ -520,6 +520,28 @@ def test_stack():
     np.testing.assert_equal(img.data, out)
 
 
+def test_stack_tolerance():
+    # Floating-point coords
+    coords = [c.astype(np.float64) for c in coords2]
+
+    for offsets in ([1e-10, 0.0], [0.0, 1e-10]):
+        img2_float = tf.Image(
+            img2.data, dims=img2.dims,
+            coords=[coords[0] + offsets[0], coords[1] + offsets[1]])
+
+        # Prior to fix for issue #1038 the next line gives:
+        # TypeError: ufunc 'over' not supported for the input types, and the
+        # inputs could not be safely coerced to any supported types according
+        # to the casting rule ''safe''
+        img = tf.stack(img2_float, img1)
+        np.testing.assert_array_equal(img.x_axis, img2_float.x_axis)
+        np.testing.assert_array_equal(img.y_axis, img2_float.y_axis)
+
+        img = tf.stack(img1, img2_float)
+        np.testing.assert_array_equal(img.x_axis, img1.x_axis)
+        np.testing.assert_array_equal(img.y_axis, img1.y_axis)
+
+
 def test_masks():
     # Square
     mask = tf._square_mask(2)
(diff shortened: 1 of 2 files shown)


jbednar:
If we override the coords, then yes, I'd only want to do so for a small enough tolerance. We have sample code in the HoloViews Image class testing whether the given grid is close enough to a fixed grid to be rendered that way (otherwise suggesting QuadMesh), and we could use similar tolerances here. But I'd be a lot more comfortable if the underlying issue were addressed. I.e., _why_ does the raster code produce different coordinates? I assume it's because we brought that code in from a different project that does things differently,  but what I don't know is how those differences occur, whether it's feasible to make them go away, and whether they are important in some cases.

ianthomas23:
> _why_ does the raster code produce different coordinates? I assume it's because we brought that code in from a different project that does things differently, but what I don't know is how those differences occur, whether it's feasible to make them go away, and whether they are important in some cases.

I have made more progress here. Coordinates for `Canvas.raster()` calls are calculated here
https://github.com/holoviz/datashader/blob/25578abde75c7fa28c6633b33cb8d8a1e433da67/datashader/core.py#L1140-L1141
This is the only use of `utils.compute_coords()` which is:
https://github.com/holoviz/datashader/blob/25578abde75c7fa28c6633b33cb8d8a1e433da67/datashader/utils.py#L314-L324

`Canvas.points()` and the other non-`raster` calls go via the `bypixel` pipeline Dispatcher magic to `Axis.compute_index()`:
https://github.com/holoviz/datashader/blob/25578abde75c7fa28c6633b33cb8d8a1e433da67/datashader/core.py#L84-L86

It looks entirely tenable to replace `raster`'s use of `compute_coords` with `Axis.compute_index` and so on.

So I suggest that we halt this PR here and I will write another one with these changes to see what that throws up.

ianthomas23:
Superseded by #1046.

PR #1044, opened 2022-02-04
Fix/cudf api updates

AjayThorve:
This PR replaces deprecated methods in cudf, related to conversion to a cupy array from cudf.DataFrame and cudf.Series. The new methods also provide performance improvements as demonstrated here: https://docs.rapids.ai/api/cudf/nightly/user_guide/10min-cudf-cupy.html#Converting-a-cuDF-Series-to-a-CuPy-Array

Finally, also updating the minimum version of cudf's last stable release (22.02)

### DIFF
--- a/datashader/glyphs/area.py
+++ b/datashader/glyphs/area.py
@@ -490,8 +490,8 @@ class AreaToZeroAxis1(_PointLike):
             aggs_and_cols = aggs + info(df)
 
             if cudf and isinstance(df, cudf.DataFrame):
-                xs = df[list(x_names)].as_gpu_matrix()
-                ys = df[list(y_names)].as_gpu_matrix()
+                xs = df[list(x_names)].to_cupy()
+                ys = df[list(y_names)].to_cupy()
                 do_extend = extend_cuda[cuda_args(xs.shape)]
             else:
                 xs = df[list(x_names)].values
@@ -597,9 +597,9 @@ class AreaToLineAxis1(_AreaToLineLike):
             aggs_and_cols = aggs + info(df)
 
             if cudf and isinstance(df, cudf.DataFrame):
-                xs = df[list(x_names)].as_gpu_matrix()
-                ys = df[list(y_names)].as_gpu_matrix()
-                y_stacks = df[list(y_stack_names)].as_gpu_matrix()
+                xs = df[list(x_names)].to_cupy()
+                ys = df[list(y_names)].to_cupy()
+                y_stacks = df[list(y_stack_names)].to_cupy()
                 do_extend = extend_cuda[cuda_args(xs.shape)]
             else:
                 xs = df[list(x_names)].values
@@ -666,7 +666,7 @@ class AreaToZeroAxis1XConstant(AreaToZeroAxis1):
             aggs_and_cols = aggs + info(df)
 
             if cudf and isinstance(df, cudf.DataFrame):
-                ys = df[list(y_names)].as_gpu_matrix()
+                ys = df[list(y_names)].to_cupy()
                 do_extend = extend_cuda[cuda_args(ys.shape)]
             else:
                 ys = df[list(y_names)].values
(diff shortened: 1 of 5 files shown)


jbednar:
Thanks!  However, it doesn't seem safe to merge this, because it requires cudf version 22.02.00 released only five days ago, and thus releasing datashader with these changes would mean people could use only the very latest cudf, which isn't always practical given the driver versions people have installed.

Since the updates are to address deprecated API, is there a minimum version of cudf supporting this API that isn't quite so new? Otherwise we'll have to park this PR and revisit it in the following rather than the upcoming Datashader release.

AjayThorve:
@jbednar, that makes sense. 

So the minimum version the new api supports is 21.12, which was released a couple of months ago (Dec 3, 2021). How old a version of cudf do you think would be practical to merge? Another option could be using api's best on the current installed version of cudf, for e.g: 22.04 to be released will have removed the older version, so we can just check the user installed version and use the corresponding api:

```python
if cudf_version >= Version("22.02"):
    return df.to_cupy()
else:
    return df.as_gpu_matrix()
```

Does the above work?

AjayThorve:
I have pushed some changes, and reverted the minimum cudf version update (although the recommended version would always be the latest stable release)

philippjfr:
Thanks @AjayThorve! The December release is still pretty recent, what's the driver compatibility for cuDF 21.12? I guess all of the recent versions require CUDA 11.x + Driver v450.80.02+? In that case I'm okay with this. Do we need to make similar updates in HoloViews?

datashader/glyphs/points.py
```diff
@@ -1,4 +1,5 @@
 from __future__ import absolute_import, division
+from distutils.version import Version
```
jbednar, line 2: I think this needs to be packaging.version rather than distutils.version, because of [distutils now being deprecated ](https://www.python.org/dev/peps/pep-0632/) and slated for removal in an upcoming Python release (see https://github.com/holoviz/holoviews/pull/5180).

datashader/reductions.py
```diff
@@ -193,7 +199,10 @@ def apply(self, df):
             else:
                 nullval = 0
             a = cupy.asarray(a)
-            b = cupy.asarray(df[self.column].to_gpu_array(fillna=nullval))
+            if cudf.__version__ >= "22.02":
```
jbednar, line 202: I think this needs to be something like `packaging.version.Version("22.02")`.

datashader/glyphs/area.py
```diff
@@ -115,8 +115,8 @@ def extend(aggs, df, vt, bounds, plot_start=True):
             aggs_and_cols = aggs + info(df)
 
             if cudf and isinstance(df, cudf.DataFrame):
-                xs = self.to_gpu_matrix(df, x_name)
-                ys = self.to_gpu_matrix(df, y_name)
+                xs = self.to_cupy_array(df, x_name)
```
jbednar, line 118: Is it important to change the name of the method like this? It seems like the name itself wouldn't matter, but maybe this is matching cudf's name in a way that you think will be clearer?

AjayThorve: the method actually returns a cupy array, (2d if its multiple columns, but a 1d array if its a single column - cudf.Series). Though behind that was, `gpu_matrix` might sound confusing in case it's a Series behind returned. Recent changes to cudf api (https://github.com/rapidsai/cudf/pull/9236), also aim at standardizing the `to_cupy()` and `to_numpy()` methods to avoid any ambiguity.

datashader/glyphs/points.py
AjayThorve: thanks, fixed this

datashader/reductions.py
AjayThorve: fixed it, thanks :)

datashader/glyphs/area.py
jbednar: Ok, sounds good! Thanks.

philippjfr:
Okay great, I'm happy to see this merged!

hvPlot indeed uses HoloViews to generate the plots, help updating it would be hugely appreciated!

PR #1045, opened 2022-02-07
CI: add recent Python versions in the matrix

maximlt:
This PR is ready in itself even if 4 tests are failing on Py 3.8, 3.9 and 3.10 due to the latest version of pandas (1.4.0) being pulled on these versions. 
```
=========================== short test summary info ============================
FAILED datashader/tests/test_datatypes.py::TestRaggedGetitem::test_getitem_invalid
FAILED datashader/tests/test_datatypes.py::TestRaggedInterface::test_tolist
FAILED datashader/tests/test_datatypes.py::TestRaggedMethods::test_where_series[True]
FAILED datashader/tests/test_datatypes.py::TestRaggedMethods::test_where_series[False]
= 4 failed, 765 passed, 48 skipped, 2 xfailed, 658 warnings in 300.81s (0:05:00) =
```

PR #1046, opened 2022-02-08
Calculate raster coords identically to other primitives

ianthomas23:
Fixes issue #1038, supersedes PR #1042.

Calculation of x, y coords for rasters was using a different code path to that used by other geometric primitives, leading to very small differences (~1e-15) that can cause problems when using `tf.stack`.

This PR removes the `utils.compute_coords` function that was only used by `Canvas.raster` and instead uses the same combination of `Axis.compute_scale_and_translation` and `Axis.compute_index` as the other primitives, so that now the coordinates are identical.

I have slightly refactored where this code is called in `Canvas.raster` as the x and y coords are now independent, and replaced use of `np.isclose(...).all()` with `np.allclose(...)` which is simpler.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -16,7 +16,7 @@ jobs:
       matrix:
         os: ['ubuntu-latest', 'macos-latest', 'windows-latest']
         python-version: [3.6, 3.7, 3.8]
-    timeout-minutes: 40
+    timeout-minutes: 60
     defaults:
       run:
         shell: bash -l {0} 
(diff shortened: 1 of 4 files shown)


philippjfr:
Much better, thank you!

jbednar:
Fabulous! *This* is what I was looking for all along. :-) 

I keep seeing test jobs being cancelled, but I haven't been cancelling them, and am not sure what is doing that.

maximlt:
Seems like the timeout of 40 minutes is not enough on MacOs. I'll increase it in https://github.com/holoviz/datashader/pull/1045.

jbednar (approved):

PR #1052, opened 2022-03-15
fix incorrect version comparison introduced in #1044

AjayThorve:
#1044 introdced a syntax error, which compared string version with `packaging.version.Version`. Minor PR to fix it.


jbednar (approved):
Thanks!

Issue #357, opened 2017-05-17
Improve colors for small numbers of discrete values


jbednar:
The eq_hist code works very well when there are large numbers of discrete values, such as a range of different counts per pixel, which can be mapped onto the full range of color values:

![image](https://cloud.githubusercontent.com/assets/1695496/26156380/beacca8e-3adb-11e7-8408-4db4b095a593.png)

However, when zooming into a plot like that, there is often a stage where there are only two or three distinct different numbers of counts, once things overlap only very rarely.  In these cases, the colors chosen are not very helpful -- typically nearly all points get mapped to a very dimly visible color, with a few randomly overlapping points showing up clearly: 

![image](https://cloud.githubusercontent.com/assets/1695496/26156194/2c868618-3adb-11e7-85f5-654efe86cb5e.png)

Similar issues occur with count or sum aggregations using timeseries, which often double back on themselves leaving strange artifacts at peaks:

![image](https://cloud.githubusercontent.com/assets/1695496/26156759/f7426df8-3adc-11e7-86d4-8540b9f4d413.png)

In each of these cases, the issues occur when there are only a small number of total colors being used in the image, in which case using the full dynamic range is not providing any more information, and it's actually making the data less visible, because the lower bound on the color range is usually chosen to be just barely saturated enough to be reliably visible.  The result is that nearly all data points in these cases end up just barely reliably visible, rather than being the most visible colors available (which is what a person would choose in this scenario).  

I'm not sure how best to address this, since the information how many discrete bins were non-zero is in the eq_hist function, while the colors are chosen in the shade function, but presumably something can be done here.

fzhang00:
I'm also having this issue with my hv.Curve plots, where in some cases there are very little bins with high data points, and most pixel got mapped to lower intensity color making them not nearly not visible. 
![image](https://user-images.githubusercontent.com/4636221/27195008-dcc4f7e0-51d2-11e7-96cc-6dc5e057774d.png)

ijstokes:
Since the shading is already dynamic, would it be feasible to establish the dynamic range and bin occupancy for a discrete and relatively low-numbered color map, then re-balance the colormap for a particular view to achieve a more uniform bin distribution, and "grounding" the bottom of the color map in something with good visibility?  I've used that kind of technique for this problem in the past, albeit under conditions where I knew exactly my dataset and the visualization I was trying to generate.

e.g.

original:

value | # of points | color
------|------------|------
1        |  100            | light blue
2        | 10               | blue
[...]

PR #1050, opened 2022-03-03
Fix unittest

Hoxbro:
Properly needs to be merged after #1045

### General notes
- The pin of `numpy` is because of https://github.com/numpy/numpy/pull/21113.
- To get `pandas` 1.4 to work (commit [df59da1](https://github.com/holoviz/datashader/pull/1050/commits./df59da1a4e3e8efeea2454b8e3bdf26457ddff3f) and [30a5d5b](https://github.com/holoviz/datashader/pull/1050/commits/30a5d5b84599cf08e37b63804c6edac981f1dbf4)) the new tests added in `pandas` 1.4 had to be updated to work with `RaggedArray`. Similar updates have done elsewhere in the tests already.   
-  Changed `open_rasterio` from `xarray` to `rioxarray` (`rasterio` is a dependency of `rioxarray`) in https://github.com/holoviz/datashader/pull/1050/commits/9e729dae8055b4202ee8050c4a8b96802c373dfb as `xarray.open_rasterio` is deprecated. Added `rioxarray` as an test dependency in `setup.py`.
- Ignored `All-NaN slice encountered` in https://github.com/holoviz/datashader/pull/1050/commits/b8989a5cd57fe14080c7d07b0bad44093a88d949, this only seem to be happening for `cupy.array`. The warning and comment was copied from:  https://github.com/holoviz/datashader/blob/11a192439b9f59c50508ebd00c70a410c1d51fd2/datashader/core.py#L1234-L1237
- Changed `df[list(x_names)].values` to `df.loc[:, list(x_names)].to_numpy()` in https://github.com/holoviz/datashader/pull/1050/commits/4a1ee28c79a19faacc80cae61f193b7faec7c86b to remove the warning `Using a non-tuple sequence for indexing is deprecated`. I changed `.values` to `.to_numpy` as this is what advised by [`pandas`](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.values.html?highlight=values#pandas.DataFrame.values), but I'm afraid that I could have overlooked something.
- I have ignored the warnings related to `datashader/tests/test_bokeh_ext.py::test_interactive_image_update` as `InteractiveImage has been deprecated as of datashader 0.8.0`.

### RaggedArray
I have updated the low hanging fruits of the `VisibleDeprecationWarning: Creating an ndarray from ragged nested sequences is deprecated` in https://github.com/holoviz/datashader/pull/1050/commits/6fa2989f4761edcafe42a224f8109a31fed1d255. 
[...]

PR #1055, opened 2022-03-28
Rescale discrete levels for how='eq_hist'

ianthomas23:
This is a candidate fix for issue #357.

Under normal circumstances if `how='eq_hist'` then the `span` is calculated as `(np.nanmin, np.nanmax)` of the masked data, hence the data is colored using the full `cmap` range. If there are a small number of discrete values in the data then this can lead to low values, which are often in the majority, being rendered at the low end of `cmap` with low `alpha`.

This PR adds a new `rescale_small_values` boolean kwarg to `shade()` with a default of `False`. If it is `True`, `how='eq_hist'` and there are a small number of discrete values in the masked data then the lower `span` limit is reduced so that the data is rendered more towards the top end of the `cmap` range.

The exact form of the equation relating the `lower_span` limit to the max value of the masked data `max_data` is to be determined. Currently trialling (line 259):
```python
lower_span = 1.0 - 0.6*np.log10(max_data)
```

### DIFF
--- a/datashader/tests/test_transfer_functions.py
+++ b/datashader/tests/test_transfer_functions.py
@@ -1034,13 +1034,13 @@ def test_eq_hist():
     data[np.random.randint(300**2, size=100)] = np.nan
     data = (data - np.nanmin(data)).reshape((300, 300))
     mask = np.isnan(data)
-    eq = tf.eq_hist(data, mask)
+    eq, _ = tf.eq_hist(data, mask)
     check_eq_hist_cdf_slope(eq)
     assert (np.isnan(eq) == mask).all()
     # Integer
     data = np.random.normal(scale=100, size=(300, 300)).astype('i8')
     data = data - data.min()
-    eq = tf.eq_hist(data)
+    eq, _ = tf.eq_hist(data)
     check_eq_hist_cdf_slope(eq)
 
 
--- a/datashader/tests/test_transfer_functions.py
+++ b/datashader/tests/test_transfer_functions.py
@@ -82,6 +82,15 @@ eq_hist_sol = {'a': np.array([[0, 4291543295, 4288846335],
                               [4281281791, 4278190335, 0]], dtype='u4')}
 eq_hist_sol['c'] = eq_hist_sol['b']
 
+eq_hist_sol_rescale_discrete_levels = {
+    'a': np.array([[0, 4289306879, 4287070463],
+                   [4284834047, 0, 4282597631],
+                   [4280361215, 4278190335, 0]], dtype='u4'),
+    'b': np.array([[0, 4289306879, 4287070463],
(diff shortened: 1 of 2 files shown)


ianthomas23:
Here is a video of it using HoloViews for interactive zooming.

https://user-images.githubusercontent.com/580326/160407840-a05dc299-828a-437a-b0ba-854c4103d6ac.mov

ianthomas23:
This now uses a different equation that gives better results. If the number of discrete values is less than 100 it increases the span range by a factor that depends on the number of discrete levels; this factor is 1 for 100 discrete values rising linearly to 1.5 for 2 discrete values. This span range factor is applied by lowering the lower span limit. An equivalent way of thinking about this is that if there are 100 or more discrete values then the data fills the colormap as usual, but as the number of discrete values is reduced the data fills a smaller proportion of the top end of the colormap, at 2 discrete levels this is the top 2/3 of the colormap.

The equation is linear in the number of discrete values. I have also tried logarithmic variation here but the results are not noticeably different. I have left the maths of the equation in the code so that there are no magic numbers.

Here is a video of it in action:

https://user-images.githubusercontent.com/580326/160814964-17ab0050-4a45-4ee0-a7de-c5ccc309beb1.mp4

ianthomas23:
Most of the test failures are Ragged array related, which are think are fixed by https://github.com/holoviz/datashader/pull/1050 but I have not explicitly checked this.
```
=========================== short test summary info ============================
FAILED datashader/tests/test_datatypes.py::TestRaggedGetitem::test_getitem_invalid
FAILED datashader/tests/test_datatypes.py::TestRaggedInterface::test_tolist
FAILED datashader/tests/test_datatypes.py::TestRaggedMethods::test_where_series[True]
FAILED datashader/tests/test_datatypes.py::TestRaggedMethods::test_where_series[False]
= 4 failed, 765 passed, 48 skipped, 2 xfailed, 80 warnings in 308.55s (0:05:08) =
```

Hoxbro:
The four failing tests should be fixed by #1050.

jbednar (approved):
Looks great! I'll rebase it now so that it can be tested, then separately rename `rescale_small_values` to `rescale_discrete_levels` since the values themselves may not actually be small; e.g. two discrete values of 10 and 100000 would still be covered by this rescaling. I'll also rename `max_data` to `num_levels`, again because the value is about the number of levels not the value of the data. I may also add control over those numeric values; not sure yet. In any case, definitely looks like a big improvement in eq_hist behavior!

jbednar:
Looks like there are still failing tests after rebasing. Also, can you please temporarily make "rescale_discrete_values" be True and make sure the tests pass in that case as well? We'll want to turn it on by default soon enough, and need to be able to use it everywhere in the meantime to test it.

Issue #1051, opened 2022-03-03
Background no longer transparent


jbednar:
Datashader 0.13 and master both show a background from the colormap, rather than letting the page background show through, in certain cases that used to work fine. E.g. the [current Datashader docs]( https://datashader.org/user_guide/Timeseries.html#trends-and-outliers) show:

![image](https://user-images.githubusercontent.com/1695496/156641173-ca733d69-025e-41e8-b579-40894e714755.png)

Here the background should have been (and used to be) NaN or zero in a way that renders to an alpha value of zero in the background, letting the page show through.

ianthomas23:
This might be a red herring, but that looks suspiciously like the output from my currently open PR. Is there any way the docs build could be using the wrong branch?

ianthomas23:
Following on from @Hoxbro's comment, if you change the line
```python
        if data.dtype.kind == 'u':
```
to
```python
        if data.dtype.kind in ('u', 'i'):
```
then this renders correctly and `pytest` runs OK. I don't know if there are wider ramifications of this change though.

Hoxbro:
Based on https://github.com/holoviz/datashader/pull/910#issuecomment-625019138, I will say it was chosen to work this way - if this is still the case, I don't know.

If it is still the case, I suggest adding an argument to `tf.shade` like `zero_as_nan`, which can switch on the desired behavior.

jbednar:
It was definitely not intended that the background ever show up blue like that. :-) So that's an unintentional implication of an intentional change.  For floats, we are using NaN to represent missing data, i.e., the absence of any values. For an integer count, 0 is the absence of values, so 0 as NaN makes sense for count. But for a sum, there could be negative numbers where the result is zero even though there is data, and in that case we can't treat 0 as no data. So I don't think it's safe to do `if data.dtype.kind in ('u', 'i'):`. But if we are summing values that we know are nonnegative, as in the example shown, then 0 as NaN should be safe.

The two cases where I can see this behavior in the docs are:

1. `.sum()` (`tf.shade(merged.sum())` in Timeseries.ipynb and shown above): .sum() returns a signed integer array even though all its arguments are nonnegative. I think this is an xarray sum method, and thus nothing we could change. We could just update the docs to force the output to be unsigned and explain why. Alternatively, we could have a sum function that checks for uint inputs and promotes them to floats with zeros as NaNs before summing, but then people would have to know to use that, or we'd have to return an xarray subclass with sum redefined, which seems like overkill.
[...]

PR #1065, opened 2022-04-09
Fix uint handling in eq_hist and xarray examples

jbednar:
Fixes https://github.com/holoviz/datashader/issues/1051 .

Xarray operations `sum` and `where` currently lose track of the input datatype, returning floats for unsigned int input. Usually that wouldn't be a problem, but preserving the type is crucial for `shade` to know that zero is a missing-data value, so patched the affected examples to make the types unsigned after those operations.

Also renamed the `normalization` option for HoloViews's `datashade` function to `cnorm`; apparently `normalization` is no longer accepted.

### DIFF
--- a/datashader/transfer_functions/__init__.py
+++ b/datashader/transfer_functions/__init__.py
@@ -170,6 +170,8 @@ def eq_hist(data, mask=None, nbins=256*256):
 
     data2 = data if mask is None else data[~mask]
     if data2.dtype == bool or (np.issubdtype(data2.dtype, np.integer) and data2.itemsize <= 4):
+        if data2.dtype == 'uint32':
+             data2 = data2.astype('int32')
         flattened = data2.reshape(-1)
         assert(flattened.itemsize <= 4)
         hist = np.bincount(flattened)
--- a/datashader/transfer_functions/__init__.py
+++ b/datashader/transfer_functions/__init__.py
@@ -170,7 +170,8 @@ def eq_hist(data, mask=None, nbins=256*256):
 
     data2 = data if mask is None else data[~mask]
     if data2.dtype == bool or (np.issubdtype(data2.dtype, np.integer) and data2.itemsize <= 4):
-        hist = np.bincount(data2.ravel())
+        flattened = data2.reshape(-1)
+        hist = np.bincount(flattened)
         bin_centers = np.arange(len(hist))
         idx = int(np.nonzero(hist)[0][0])
         hist, bin_centers = hist[idx:], bin_centers[idx:]
--- a/datashader/transfer_functions/__init__.py
+++ b/datashader/transfer_functions/__init__.py
@@ -171,6 +171,7 @@ def eq_hist(data, mask=None, nbins=256*256):
     data2 = data if mask is None else data[~mask]
     if data2.dtype == bool or (np.issubdtype(data2.dtype, np.integer) and data2.itemsize <= 4):
         flattened = data2.reshape(-1)
+        assert(flattened.itemsize <= 4)
         hist = np.bincount(flattened)
         bin_centers = np.arange(len(hist))
         idx = int(np.nonzero(hist)[0][0])
--- a/datashader/transfer_functions/__init__.py
+++ b/datashader/transfer_functions/__init__.py
@@ -151,7 +151,8 @@ def eq_hist(data, mask=None, nbins=256*256):
        Boolean array of missing points. Where True, the output will be `NaN`.
     nbins : int, optional
         Number of bins to use. Note that this argument is ignored for integer
-        arrays, which bin by the integer values directly.
+        and boolean arrays with dtypes 32 bits or smaller, which bin by the
+        integer values directly.
 
     Notes
     -----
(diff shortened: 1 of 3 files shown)


examples/getting_started/2_Pipeline.ipynb
```diff
@@ -315,14 +315,19 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "tf.Images(tf.shade(agg_d3_d5.where(aggc.sel(cat='d3') == aggc.sel(cat='d5')), name=\"d3+d5 where d3==d5\"),\n",
-    "          tf.shade(      agg.where(aggc.sel(cat='d3') == aggc.sel(cat='d5')), name=\"d1+d2+d3+d4+d5 where d3==d5\"))"
+    "sel1 = agg_d3_d5.where(aggc.sel(cat='d3') == aggc.sel(cat='d5')).astype('uint64')\n",
+    "sel2 =       agg.where(aggc.sel(cat='d3') == aggc.sel(cat='d5')).astype('uint64')\n",
```
jbednar, line 319:
```suggestion
    "sel1 = agg_d3_d5.where(aggc.sel(cat='d3') == aggc.sel(cat='d5')).astype('uint32')\n",
    "sel2 =       agg.where(aggc.sel(cat='d3') == aggc.sel(cat='d5')).astype('uint32')\n",
```

uint64 was causing issues on Linux, presumably because of eq_hist allocating space for all possible integer values

datashader/transfer_functions/__init__.py
```diff
@@ -169,8 +170,10 @@ def eq_hist(data, mask=None, nbins=256*256):
 
     data2 = data if mask is None else data[~mask]
     if data2.dtype == bool or np.issubdtype(data2.dtype, np.integer):
-        if data2.dtype.kind == 'u':
-             data2 = data2.astype('i8')
+        # workaround for bincount not supporting uint64 (https://github.com/numpy/numpy/issues/17760)
```
Hoxbro, line 173: Out of curiosity, why not use `histogram` for `uint64` instead of downcasting it?

examples/getting_started/2_Pipeline.ipynb
```diff
@@ -315,14 +315,19 @@
...
+    "\n",
+    "tf.Images(tf.shade(sel1, name=\"d3+d5 where d3==d5\"),\n",
+    "          tf.shade(sel2, name=\"d1+d2+d3+d4+d5 where d3==d5\"))"
    ]
   },
   {
    "cell_type": "markdown",
    "metadata": {},
    "source": [
+    "Here we're specifying `uint64` because at least current versions of xarray's `where` method convert to floating point, and keeping the type unsigned is important for `shade` to know that zero is a missing-data value here, to be rendered with a transparent background.\n",
```
Hoxbro, line 329: Noticed that `uint64` should be `uint32` here to match with the code cell.

jbednar: Good catch!

```suggestion
    "Here we're specifying `uint32` because at least current versions of xarray's `where` method convert to floating point, and keeping the type unsigned is important for `shade` to know that zero is a missing-data value here, to be rendered with a transparent background.\n",
```

datashader/transfer_functions/__init__.py
jbednar: It's a good question. Mainly I don't want to change too many things, as we're trying to get a release out by last week. But there are other factors, such as that if a user did use uint64, it would allocate an enormous array to hold the results, and I want to give users feedback not to do that. I also considered automatically switching to the binning approach (instead of one bin per possible integer value) for any 64-bit integer type, and on balance, maybe I should just do that.

jbednar: Philipp and I worked out a better approach using `np.unique` and putting guards on the potential number of unique values to avoid these issues.

jbednar:
In the process of testing these examples, it became clear that the handling of ints and uints in eq_hist was revealing issues with the underlying numpy bincount function, which [doesn't handle uint64](https://github.com/numpy/numpy/issues/17760) and was behaving strangely on Linux for code that worked on a similar Mac machine (see the CI history for this PR). Eventually, Philipp and I worked out a different way to handle the integer-count case in eq_hist.

For background, integers have always been handled in a separate code path that attempts to measure exact counts for each value, which previously required allocating an array as large as the maximum integer value for computing those counts. Thus in practice, if an aggregate contained a pixel with a very large count, a very large array could be required, reaching memory limitations of the underlying system. Under Linux in some cases this limit was reached by bincount, with a request for 16GB, but the "max count" in involved was 2^31-1, strongly suggesting an underflow somewhere we couldn't track down (presumably in the NumPy C code from what we can tell -- feel free to try to debug the issue from the CI results!). 

Anyway, to avoid this issue entirely, we decided to:

1. Only attempt the exact-count approach for a maximum aggregate value less than `nbins`, which means allocating a data structure at most `nbins` long in all cases. That way we can have exact computations for feasible (and typical) situations, and fall back to the histogramming/binning approach otherwise.
2. Instead of using bincount, use `np.unique`, which should behave better for outliers, since it keeps a running list of values encountered rather than allocating based on the maximum.

This seems to avoid problems on Linux and should give better behavior for the largest datasets as well (avoiding potential memory issues and integer type casting issues).

PR #1072, opened 2022-04-21
Fix antialiased line issues

ianthomas23:
This is a WIP to fix various issues with antialiased lines for the impending release.

The major change is support for both 1 and 2-stage aggregations. Where possible (e.g. `ds.any` and `ds.max`) a single-stage agg is used on an individual line segment basis. `ds.min` always uses a 2-stage agg on an individual line segment basis. `ds.count` and `ds.sum` now have 2 options controlled by `self_intersect` boolean kwarg. If `False` the usual 2-stage agg is used which is very accurate but slower and ignores line self-intersections. If `True`, the default, a 1-stage agg is used that considers the previous line segment so that pixels are not overwritten at line segment joins; this therefore deals with self-intersections and is faster but there will be rendering artifacts in some situations (some are fixable, others are unavoidable).

I am open to suggestions for a better name for the kwarg than `self_intersect`.

Antialiased self-intersecting path using `ds.count` now looks like this (from 15-Large_Data.ipynb): 
paths

Antialiased multiple timeseries using `ds.count` now looks like this (from 3_Timeseries.ipynb):
overlap_ts

### DIFF
--- a/datashader/core.py
+++ b/datashader/core.py
@@ -261,7 +261,7 @@ class Canvas(object):
             approximating the visual appearance of a subpixel line
             width.
         antialias : bool, optional
-            This option is kept for backward compatibility only. 
+            This option is kept for backward compatibility only.
             ``True`` is equivalent to ``line_width=1`` and
             ``False`` (the default) to ``line_width=0``. Do not specify
             both ``antialias`` and ``line_width`` in the same call as a
@@ -335,7 +335,8 @@ class Canvas(object):
         """
         from .glyphs import (LineAxis0, LinesAxis1, LinesAxis1XConstant,
                              LinesAxis1YConstant, LineAxis0Multi,
-                             LinesAxis1Ragged, LineAxis1Geometry)
+                             LinesAxis1Ragged, LineAxis1Geometry,
+                             AntialiasCombination)
 
         validate_xy_or_geometry('Line', x, y, geometry)
 
@@ -427,7 +428,11 @@ The axis argument to Canvas.line must be 0 or 1
         glyph.set_line_width(line_width)
         if line_width > 0:
             if isinstance(agg, (rd.any, rd.max)):
-                glyph.set_antialias_combination_max()
+                glyph.set_antialias_combination(AntialiasCombination.MAX)
+            elif isinstance(agg, rd.min):
+                glyph.set_antialias_combination(AntialiasCombination.MIN)
+            else:
+                glyph.set_antialias_combination(AntialiasCombination.SUM)
 
             # Switch agg to floating point.
             if isinstance(agg, rd.count):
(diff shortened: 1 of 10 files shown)


datashader/glyphs/line.py
```diff
@@ -17,23 +19,41 @@
     cuda_args = None
 
 
+# Eventually this will be replaced with attributes and/or
```
jbednar, line 22: Making promises in comments isn't usually a good idea, unless it's a promise to be fulfilled in this PR. Otherwise those stay around in the code confusing people! Here e.g. it could say "It would be cleaner to replace these constants with ...".

ianthomas23: Corrected.

ianthomas23:
The above multiple timeseries plot has artifacts to the left of y-extrema. This is now fixed:

lw1

Here is the equivalent `line_width=0` plot:

lw0

ianthomas23:
Performance of antialiased lines is slower than non-antialiased for 2 reasons:

  1. Low-level pixel-setting code is more complicated.
  2. Two-stage agg code has multiple agg-combining calls, each of which is a NaN-aware combination of 2 arrays of same shape as canvas plus a reset of one of those arrays to zero (or NaN). This does not occur for one-stage agg.

Timings using code based on last plot in Timeseries jupyter notebook that combines `nlines` lines of 10-points each onto a 1000x400 pixel canvas. Timed `line_width=1` for both `ds.count(self_intersect=True)` (1-stage agg) and `ds.count(self_intersect=False)` (2-stage agg) compared to `line_width=0`.

| nlines | Slowdown factor for 1-stage agg | Slowdown factor for 2-stage agg |
| --- | --- | --- |
| 100 | 1.3 | 2.5 |
| 1k | 1.6 | 4.7 |
| 10k | 3.3 | 7.9 |
| 100k | 5.4 | 13.0 |
| 1M | 5.7 | 13.7 |

ianthomas23:
Example output for multiple timeseries (1000 here) where x-values are spaced apart, for `line_width=1`:
ts_lw1

Compare with `line_width=0`:
ts_lw0

datashader/core.py
```diff
@@ -426,12 +427,25 @@ def line(self, source, x=None, y=None, agg=None, axis=0, geometry=None,
 
         glyph.set_line_width(line_width)
         if line_width > 0:
+            # Eventually this will be replaced with attributes and/or
```
jbednar, line 430:
```suggestion
            # Eventually this should be replaced with attributes and/or
```

datashader/glyphs/line.py
```diff
@@ -17,23 +19,41 @@
     cuda_args = None
 
 
+# It would be preferable to replace this with attributes and/or
+# member functions of Reduction classes.
```
jbednar, line 23:
```suggestion
# This Enum should eventually be replaced with attributes
# and/or member functions of Reduction classes.
```

jbednar (approved):
I don't know any clearer option name, so it looks ready to merge to me!

PR #1075, opened 2022-04-24
Fix the failing example on Windows / Python 3.7

maximlt:
The test workflow was failing on Windows / Python 3.7 with an error coming from `fiona` (`ImportError: the 'read_file' function requires the 'fiona' package, but it is not installed or does not import correctly.`). An [issue](https://github.com/conda-forge/fiona-feedstock/issues/196) has been opened on the repo of the conda-forge recipe of `fiona`. In the meantime, this PR pins gdal and fiona to older versions on Windows / Python 3.7, which fixes the issue.


jbednar:
Thanks!!

PR #1077, opened 2022-04-25
Fix polygon rendering in docs and simplify code

philippjfr:
Fixes an issue with a plot in the 8_Polygons user guide not rendering initially.

### DIFF
--- a/examples/user_guide/8_Polygons.ipynb
+++ b/examples/user_guide/8_Polygons.ipynb
@@ -553,7 +553,7 @@
    "outputs": [],
    "source": [
     "def compute_partitions(el):\n",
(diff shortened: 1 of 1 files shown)


examples/user_guide/8_Polygons.ipynb
```diff
@@ -552,22 +552,11 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "polys = hv.Polygons(ddf, vdims=['pop_est'])\n",
+    "def compute_partitions(el):\n",
+    "    n = ddf.cx_partitions[slice(*el.range(0)), slice(*el.range(1))].npartitions\n",
```
philippjfr, line 556:
```suggestion
    "    n = ddf.cx_partitions[slice(*el.range('x')), slice(*el.range('y'))].npartitions\n",
```

PR #1086, opened 2022-05-06
Skip rescaling when not applicable

jbednar:
`rescale_discrete_values` is only valid for some cnorm types, and having it error is inconvenient for external tools that want to set default values appropriate for any cnorm type, so we should catch such cases and avoid errors.


ianthomas23:
This is a duplicate of #1085.

PR #1092, opened 2022-06-10
Do not snap trimesh vertices to pixel grid

### LINKED ISSUE #806: Trimesh rasterization seems slightly off

ianthomas23:
Closes #806.

Previously `trimesh` snapped each vertex to the nearest pixel center before deciding which pixels were inside each triangle. This worked mostly OK except that it never rendered pixels in the right-hand column or bottom row, and small triangles (just a few pixels wide/high) rendered badly. This fix does not snap vertices to pixel centers to avoid these problems.

Here are before and after images for the three test cases in issue #806, blue lines show the boundary passed to `trimesh` and the last example should completely fill the agg:
before1 after1

before2after2

before3after3

Most of the work here is changing and checking all of the trimesh tests. I have passed a human eyeball over all of the integration test results (`test_pandas` and `test_dask`), and had to rewrite all of the unit tests (`test_glyphs`) as the internal `draw_triangle` functions now take floating-point vertex locations (unsnapped).


datashader/glyphs/trimesh.py
```diff
@@ -214,10 +227,10 @@ def map_onto_pixel(vt, bounds, x, y):
 
         Points falling on upper bound are mapped into previous bin.
```
jbednar, line 228: Is this statement still accurate?

jbednar (approved):
Looks good; thanks so much! Happy for you to merge it after addressing one comment about the docstring.

datashader/glyphs/trimesh.py
jbednar: (Maybe that same point should be made still, but elsewhere, since it's no longer enforced here?)

ianthomas23: No, it is no longer accurate so I have removed it.

ianthomas23:
I have made another small change so that subpixel rendering (whole triangle within a single pixel) remains the same.

I am as happy with this as I can be. Longer term I think it will be necessary to check the results are consistent with `canvas.polygons` plus also that we are correctly dealing with special situations like long thin axis-aligned meshes (e.g. 1 or 2 pixels in y dimension, many pixels in x dimension) as they are not handled particularly well, both before and after this PR.

PR #1093, opened 2022-06-12
Implementation of first and last reduction

tselea:
For my current work, it is useful for me to make use of the first([column]) and last([column]) reductions, which at the moment are implemented only for rasters.

For both reductions, I've initialised a numpy array with numpy.nan values ( create() method). The append() method is implemented as follows:

- first([column]): check if field is not null and if agg[y,x] is null (has not yet been filled), then set agg[y,x] = field
- last([column]): check if field is not null, then set agg[y,x] = field

For both reductions, the finalize() method wraps the numpy array to a xarray DataArray (similar to the count() reduction).

Please let me know if you think this may be a good fit for the project.

### DIFF
--- a/datashader/reductions.py
+++ b/datashader/reductions.py
@@ -876,7 +876,7 @@ class first(Reduction):
 
     @staticmethod
     @ngjit
-    def _append(x, y, agg,field):
+    def _append(x, y, agg, field):
         if not isnull(field) and isnull(agg[y, x]):
             agg[y, x] = field
 
--- a/datashader/reductions.py
+++ b/datashader/reductions.py
@@ -875,20 +875,21 @@ class first(Reduction):
     _dshape = dshape(Option(ct.float64))
 
     @staticmethod
-    def _append(x, y, agg):
-        raise NotImplementedError("first is currently implemented only for rasters")
+    def _append(x, y, agg,field):
+        if not isnull(field) and isnull(agg[y, x]):
+            agg[y, x] = field
 
     @staticmethod
     def _create(shape, array_module):
-        raise NotImplementedError("first is currently implemented only for rasters")
+        return array_module.full(shape, np.nan)
 
     @staticmethod
     def _combine(aggs):
         raise NotImplementedError("first is currently implemented only for rasters")
 
     @staticmethod
-    def _finalize(bases, **kwargs):
-        raise NotImplementedError("first is currently implemented only for rasters")
+    def _finalize(bases,cuda=False, **kwargs):
+        return xr.DataArray(bases[0], **kwargs)
 
 
 
--- a/datashader/reductions.py
+++ b/datashader/reductions.py
@@ -886,7 +886,7 @@ class first(Reduction):
 
     @staticmethod
     def _combine(aggs):
-        raise NotImplementedError("first is currently implemented only for rasters")
+        raise NotImplementedError("first is not implemented for dask DataFrames")
 
     @staticmethod
     def _finalize(bases, cuda=False, **kwargs):
@@ -922,7 +922,7 @@ class last(Reduction):
 
     @staticmethod
     def _combine(aggs):
-        raise NotImplementedError("last is currently implemented only for rasters")
+        raise NotImplementedError("last is not implemented for dask DataFrames")
 
     @staticmethod
     def _finalize(bases, cuda=False, **kwargs):
--- a/datashader/reductions.py
+++ b/datashader/reductions.py
@@ -925,7 +925,7 @@ class last(Reduction):
         raise NotImplementedError("last is currently implemented only for rasters")
 
     @staticmethod
-    def _finalize(bases, **kwargs):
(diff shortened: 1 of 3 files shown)


jbednar:
That's super helpful, thanks! I approve adding these to Datashader and at a quick glance the code looks good, but I only had a minute to look at it. @ianthomas23 , can you look a little closer just in case?

tselea:
Thank you @jbednar for the reply! Glad to contribute. Please let me know if further changes are needed.

datashader/reductions.py
```diff
@@ -875,20 +875,22 @@ class first(Reduction):
...
+            agg[y, x] = field
 
     @staticmethod
     def _create(shape, array_module):
-        raise NotImplementedError("first is currently implemented only for rasters")
+        return array_module.full(shape, np.nan)
 
     @staticmethod
     def _combine(aggs):
         raise NotImplementedError("first is currently implemented only for rasters")
```
ianthomas23, line 889: This `NotImplementedError` needs a different explanation now, perhaps "first is not implemented for dask DataFrames"?

datashader/reductions.py
```diff
@@ -909,20 +911,22 @@ class last(Reduction):
     _dshape = dshape(Option(ct.float64))
 
     @staticmethod
-    def _append(x, y, agg):
-        raise NotImplementedError("last is currently implemented only for rasters")
```
ianthomas23, line 913: As above.

ianthomas23 (changes requested):
This is looking good so far, I've left a few review comments.

The `NotImplementedError` needs the mention of dask because the `_combine` function is called for dask `DataFrames` to combine the results of the multiple dask partitions. We don't need to support this method of working because we will be unable to decide which was the overall `first`/`last`, so we just need to explicitly state that use of dask is not supported here.

We also need a test here, which will remove the Code Coverage complaints. I'd probably opt for a new test in `datashader/tests/test_pandas.py` using a small (5x5?) canvas and a single line with `x`, `y` and `z` values that crosses itself. Then use `canvas.lines(..., agg=ds.first("z"), ...)` and check the return against a known result, then do the same for `ds.last("z")`. If this isn't enough information I can help more!

tselea:
Thank you @ianthomas23 for the review! Sorry for the late response.

Your review comments and suggestions for test implementation were really helpful! Thanks! Hope I understood corectly. 
The line exemple that crosses itself that I used for testing is depicted below.
line_example

Please let me know if further changes are needed.

datashader/reductions.py
```diff
@@ -875,20 +875,22 @@ class first(Reduction):
     _dshape = dshape(Option(ct.float64))
 
     @staticmethod
-    def _append(x, y, agg):
-        raise NotImplementedError("first is currently implemented only for rasters")
+    @ngjit
+    def _append(x, y, agg,field):
```
ianthomas23, line 879:
```suggestion
    def _append(x, y, agg, field):
```

datashader/reductions.py
```diff
@@ -909,20 +911,22 @@ class last(Reduction):
     _dshape = dshape(Option(ct.float64))
 
     @staticmethod
-    def _append(x, y, agg):
-        raise NotImplementedError("last is currently implemented only for rasters")
+    @ngjit
+    def _append(x, y, agg,field):
```
ianthomas23, line 915:
```suggestion
    def _append(x, y, agg, field):
```

datashader/tests/test_pandas.py
```diff
@@ -546,6 +546,47 @@ def test_categorical_std(df):
...
+    sol = np.array([
+        [np.nan, np.nan, np.nan, np.nan, 100.],
+        [np.nan, np.nan, 102., 100., 103.],
+        [np.nan, np.nan, 100., 104., np.nan],
+        [np.nan, 100., 102., np.nan, np.nan],
+        [100., 101., 101., np.nan, np.nan]], dtype='float64')
+
+    out = xr.DataArray(sol, coords=[lincoords, lincoords],
+                       dims=['y', 'x'])
+    assert_eq_xr(agg, out)
```
ianthomas23, line 567:
```suggestion
    assert_eq_ndarray(agg, sol)
```

ianthomas23 (changes requested):
This is very nearly there. The tests are very clear, but you could make them simpler by only comparing the numpy arrays `agg.data` and `sol` rather than the xarray DataArrays. Then the `axis` and `lincoords` aren't needed at the top of the functions. If you can change this in both test functions and the extra couple of whitespace characters that I just found, it will be ready to merge.

tselea:
I’ve updated the code based on your last review. Please let me know if further changes should be implemented.

I’m grateful for you taking the time to review the code and for all your suggestions. Thanks, @ianthomas23!

ianthomas23 (approved):
Thanks @tselea for your first PR. We hope to see you again soon!

PR #1094, opened 2022-06-14
Remove empty histogram bins in eq_hist

ianthomas23:
This removes empty histogram bins and implements `rescale_discrete_levels=True` when using `shade(..., how="eq_hist")` with floating-point aggs. This was previously only implemented for integer aggs, and brings it into line with `EqHistColorMapper` in Bokeh which does not special case integer datasets as JavaScript does not have integer types.

Test changes in the antialiased line tests are because they  use `how="eq_hist"` (they should probably be `how="linear"` but that is a separate change). Most of them look identical to the naked eye and are just small floating-point ~changes~ improvements in the interpolation of the CDF.  `test_007_noaa_normal.png` looks worse to the naked eye as the pixels where the lines cross over are no longer visible, but this is actually a more accurate result numerically as there are 194 pixels with a value of 5 and 4 pixels with a value of 10 so the value 5 pixels are at 98% of the max CDF so almost the same RGB as the value 10 pixels.


jbednar (approved):
Thanks!

Issue #1054, opened 2021-01-13
Using datashade=True flips the image up-down


bidhya:
#### Description of expected behavior and the observed behavior
Using datashade=True option the plotted image is flipped up-down. Is there some bug due to datashade or I am doing something wrong. I also see the dynamic range of color changed as well. 

See the two images below side-by-side, with only change being 'datashade=True'

Here dem is any 1-band DEM
#### Complete, minimal, self-contained example code that reproduces the issue

```
dem = dems[0] 
ds =rioxarray.open_rasterio(f'{base_folder}/{dem}')
print(f'DEM: {dem}')
print(ds.shape)

asp = ds.shape[-1]/ds.shape[1]

ds.hvplot.image(x='x', y='y', aspect=asp, frame_width=250, cmap='gist_earth') + ds.hvplot.image(datashade=True, x='x', y='y', aspect=asp, frame_width=250, cmap='gist_earth')

```
#### Screenshots or screencasts of the bug in action
![hvplot_flip_datashade](https://user-images.githubusercontent.com/6469101/104520894-dd07de80-55c9-11eb-9126-b37e717580fc.png)

#### ALL software version info
all these libraies from conda-forge
hvplot:  0.7.0  
datashader 0.12.0  
holoviews 1.14.1  
xarray 0.16.2  
rioxarray 0.1.1

ianthomas23:
This is indeed a bug in datashader. It is simple to understand, but is going to be awkward to fix without breaking other use cases.

The clue is that GeoTIFF files usually (always?) have their dimensions in the order `(y, x)` and the y-coordinates are decreasing. Other uses still use the numpy standard of `(y, x)` but coordinates increasing, plus it is entirely possible to use `(x, y)` dimensions instead. In `datashader/utils.py` we have inconsistent handling of the dimension order:

- `calc_res` assumes `(y, x)` and flips the y-axis.
- `calc_bbox` assumes `(x, y)`.
- `orient_array` assumes `(y, x)` as most of the work is done by `calc_res`.
- `downsample_aggregate` uses `(x, y)`.

We could do a localised fix in `calc_res` to check the dimension order and behave correctly, but then I suspect that (possibly incorrect) assumptions made elsewhere in datashader and possibly even in HoloViews/hvPlot might be affected.

Various functions calculate and/or pass around array properties named similar to `(xprop, yprop)`. In some cases these might actually be referring to properties in the `(x, y)` order and in others they might be referring to properties in the same order as the array indices.

[...]

PR #1095, opened 2022-06-15
Correctly orient (y, x) arrays

ianthomas23:
Candidate fix for #1054. Surprisingly small amount of code changed in the end.

`Canvas.raster` and `tf.shade` both accept `xr.DataArray`s with `dims[-2:] == ('y', 'x')` and the direction of each of `x` and `y` `coords` may be increasing or decreasing. The orientation of the `DataArray` returned from these functions should be the same as the input, so if e.g. the input to `Canvas.raster` is flipped in the x-direction then so will the output be. This keeps the data and coordinates in sync. This was not previously the case, the data was flipped within datashader but the coordinates weren't.

Because the data and its coordinates are synchronised, you can plot the `DataArray`s in `holoviews` or `hvplot` and they are correctly rendered with increasing x and y.

Here are two examples of this working. Firstly the OP's test case of issue #1054 using a GeoTIFF obtained from https://github.com/mommermi/geotiff_sample. The land is on the southeast, the x-coords are increasing and the y-coords are decreasing:

two

At the top the datashader `Image` has y-flipped as the input data does because it just draws the image data and ignores the coordinates. Below this it is plotted directly by `hvplot` and also using `datashader` which use the coordinates correctly.

Second example to be completely explicit.

one

There are 4 `DataArray`s, one for each of the 4 x/y flipping combinations. `hv.Image` and `shade(image)` draw them all correctly as the coordinates remain correctly oriented with respect to the data regardless of if they are passed through datashader or not. At the bottom are `datashader.Image`s that are returned from `ds.Canvas.raster` followed by `ds.tf.shade`. The images are shown in their flipped orientation as the coordinates are ignored for rendering.

One thing I have not done is consider `DataArray`s with `dims == ('x', 'y')`. These are allowed but always interpreted as `dims == ('y', 'x')` so strange things can result. We will need to do something about this as even some of the reproducers in issue #1054 use `dims == ('x', 'y')` which made them not very helpful in diagnosing the cause of the problem. I don't think we can forceably transpose them to get `('y', 'x')` but we could do some checking and either raise an exception or print a warning. But we need to be careful as we cannot just look for `x` and `y` but also need to consider `lat`, `lon`, `latitude` and so on.

I have kept the `calc_res` function as it is. This returns the x and y resolutions (i.e. spacings) of the specified `DataArray`, but it chooses to reverse the yresolution. This annoys me a lot as it is unnecessary, but I have left it in case of external libraries that are  using this function as it is considered public.

### DIFF
--- a/datashader/tests/test_utils.py
+++ b/datashader/tests/test_utils.py
@@ -1,7 +1,9 @@
 from __future__ import absolute_import
 from datashape import dshape
+import numpy as np
+from xarray import DataArray
 
-from datashader.utils import Dispatcher, apply, isreal
+from datashader.utils import Dispatcher, apply, calc_res, isreal, orient_array
 
 
 def test_Dispatcher():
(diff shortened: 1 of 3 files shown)


ianthomas23:
I think we can scratch my concerns about `dims=('x', 'y')` as the changes here deal with it in a reasonable way. Now that datashader returns `DataArray`s with the correct coordinates, flipped or not, holoviews renders them correctly regardless of dim order. Here is the second example above but using `dims=('x', 'y')` rather than `('y', 'x')`:

xy

Holoviews output is correct. Datashader output as images ignores the attached coords so is displayed using the conventional increasing `('y', 'x')`. This can give you a headache if you are trying to interpret it as shown, but if you pass it on to any other code that cares about the dims then everything turns out fine.

ianthomas23:
Self merging so we can try it out in dev release.

PR #1097, opened 2022-06-20
Correctly handle NaNs in antialiased line coordinates

ianthomas23:
This PR fixes antialiased line handling of NaNs in point coordinates. NaNs are already correctly handled for deciding which line segments to draw (for antialiased and non-antialiased lines) but they also need to be used to decide when end caps need to be drawn on the ends of line segments. Some of the logic was already present for this, but because end caps are not drawn on non-antialiased lines it wasn't fully complete.

I have added tests for all the possible line types, included ragged and geometry.

### DIFF
--- a/datashader/glyphs/line.py
+++ b/datashader/glyphs/line.py
@@ -1008,7 +1008,7 @@ def _build_extend_line_axis0(draw_segment, expand_aggs_and_cols, antialias_combi
         segment_start = (plot_start if i == 0 else
                          (isnull(xs[i - 1]) or isnull(ys[i - 1])))
 
-        segment_end = (i == len(xs)-2)
+        segment_end = (i == len(xs)-2) or isnull(xs[i+2]) or isnull(ys[i+2])
 
         if segment_start or antialias_combination != AntialiasCombination.SUM_1AGG:
             xm = 0.0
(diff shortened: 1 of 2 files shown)


datashader/glyphs/line.py
```diff
@@ -1437,10 +1437,14 @@ def extend_cpu_numba(
                 segment_start = (
                         (j == 0) or
                         isnull(x_flat[x_start_index + j - 1]) or
-                        isnull(y_flat[y_start_index + j] - 1)
```
ianthomas23, line 1440: Note that the location of the closing square bracket has been wrong here for quite a long time.

ianthomas23:
Self-merging prior to 0.14.1 release tomorrow.

datashader/glyphs/line.py
jbednar: Yikes! With what effect?

ianthomas23: Fortunately the incorrect array index was always in bounds, so wouldn't have caused a segmentation violation! 

The `isnull()` is essentially a `nan` check, so subtracting the `1` makes no difference. Hence it was looking one index out to identify if a line segment between 2 (x, y) points is the start of a new line or not, and could have made the wrong decision. Before antialiased lines the `segment_start` flag was only used to decide whether to render the start pixel, so this error could have double-rendered some pixels.

You could probably design a pathological case to cause a line to be completely double rendered. But it would only ever have affected those using `nan` coords in ragged arrays, and if you are using ragged arrays you don't ever need to use `nan`s anyway. So it probably didn't affect any real user such that they would notice.

PR #1100, opened 2022-07-05
Improve graph line rendering

sagfox:
Add line functionality to accept input in form (x_src, y_src, x_dst, y_dst), and run on gpu

### DIFF
--- a/datashader/core.py
+++ b/datashader/core.py
@@ -374,6 +374,8 @@ class Canvas(object):
             orig_x, orig_y = x, y
             x, y = _broadcast_column_specifications(x, y)
 
+            # (x_src, y_src, x_dst, y_dst)
+
             if axis == 0:
                 if (isinstance(x, (Number, str)) and
                         isinstance(y, (Number, str))):


ianthomas23:
@sagfox I have been thinking about this since we talked, and I think the first task is to identify exactly what your preferred data format is. If that is a DataFrame with 4 columns `x_src`, `y_src`, `x_dst`, `y_dst` then Datashader already supports that via the `LineAxis1` class:

```python
import pandas as pd
import datashader as ds

cvs = ds.Canvas(plot_width=100, plot_height=100)
df = pd.DataFrame(dict(
    x_src=[2, 9, 5, 3], y_src=[1, 9, 1, 9],
    x_dst=[1, 3, 7, 9], y_dst=[5, 6, 5, 2],
))
agg = cvs.line(source=df, x=["x_src", "x_dst"], y=["y_src", "y_dst"], axis=1, agg=ds.count())
im = tf.shade(agg)
ds.utils.export_image(im, "lines_2pts", background="white")
```

But if your motivation is graphs with nodes and edges and each node can potentially have many incident edges then this format duplicates your node coordinates. Maybe a better approach is support for indexed lines, i.e. the nodes as a sequence of `(x, y)` coordinates and the edges a sequence of `(start_index, end_index)` which index into the nodes sequence. This would seem to imply two separate DataFrames of different lengths, which Datashader doesn't yet support, but that should not be insurmountable.

AjayThorve:
> @sagfox I have been thinking about this since we talked, and I think the first task is to identify exactly what your preferred data format is. If that is a DataFrame with 4 columns `x_src`, `y_src`, `x_dst`, `y_dst` then Datashader already supports that via the `LineAxis1` class:
> 
> ```python
> import pandas as pd
> import datashader as ds
> import datashader.transfer_functions as tf
> 
> cvs = ds.Canvas(plot_width=100, plot_height=100)
> df = pd.DataFrame(dict(
>     x_src=[2, 9, 5, 3], y_src=[1, 9, 1, 9],
>     x_dst=[1, 3, 7, 9], y_dst=[5, 6, 5, 2],
> ))
> agg = cvs.line(source=df, x=["x_src", "x_dst"], y=["y_src", "y_dst"], axis=1, agg=ds.count())
> im = tf.shade(agg)
> ds.utils.export_image(im, "lines_2pts", background="white")
> ```

@ianthomas23 that is interesting, I have never come across examples which demonstrate that. This could be all we need, but I was wondering why the graph edge implementation in datashader [here](https://github.com/holoviz/datashader/blob/master/datashader/bundling.py#L335). It seems it's using a path, with NaN after every 2 rows, to break the path to render it as a separate line?

PR #1104, opened 2022-08-03
Correct dask and bokeh dependencies

### LINKED ISSUE #1053: datashader requires `bokeh` on conda-forge ?
#### Description of expected behavior and the observed behavior
I've noticed that datashader on `conda-forge` has a dependency on `bokeh` which I think is quite odd.
(It's one of the biggest dependencies and if one intends to use `matplotlib`, it is obsolete)

using the following in a fresh environment 
```
conda install -c conda-forge datashader
```
adds the following dependency:
```
bokeh                                 2.4.2  py39hcbf5309_0      conda-forge/win-64      14 MB
```

... also with `mamba` a query of the dependencies shows:
`mamba repoquery whoneeds --tree bokeh`
```
bokeh[2.4.2]
  └─ dask[2022.1.1]
     └─ datashader[0.13.0]
```

ianthomas23:
Fixes #1053.

`install_requires` in `setup.py` now contains `dask`, which is what `pip` requires. `conda` requires `dask-core` only which is what the `pyctdev` mapping in `setup.cfg` produces. These changes ensure that `bokeh` is not an install dependency for `datashader`.

The `conda-forge` `datashader-feedstock` will need fixing in line with this, but I will do that when 0.14.2 is released and the `conda-forge` bot identifies the new release and raises a PR for the updates.

I have also changed `license_file` to `license_files` as the former is deprecated, and added `spatialpandas` to the `examples` dependencies as example 8 (polygons) uses `spatialpandas`. There is no requirement to add the full `dask` to the `examples` dependencies as this is automatically pulled in by the `spatialpandas` requirements.

`spatialpandas` needs to have similar changes to this PR, and we will have to check how the `dask` dependencies chain through the libraries following that. But there will be a `datashader` release before the next `spatialpandas` release and this PR is sufficient until the latter occurs.

I have check local installs using `pip` and `conda` (via `pyctdev`) and also built both `pip` and `conda` packages locally to check that the changes here are correct.

### DIFF
--- a/conda.recipe/meta.yaml
+++ b/conda.recipe/meta.yaml
@@ -26,7 +26,7 @@ requirements:
   run:
     - python {{ sdata['python_requires'] }}
     {% for dep in sdata.get('install_requires',{}) %}
-    - {{ dep }}
+    - {{ dep if dep != 'dask' else 'dask-core'}}
     {% endfor %}
     # adding pyct here is temporary (this conda recipe template will disappear when new pyctdev is released)
     - pyct
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,5 +1,9 @@
 [metadata]
-license_file = LICENSE.txt
+license_files = LICENSE.txt
 
 [wheel]
 universal = 1
+
+[tool:pyctdev.conda]
+namespace_map =
+    dask=dask-core
--- a/setup.py
+++ b/setup.py
@@ -9,13 +9,9 @@ import pyct.build
 ########## dependencies ##########
 
 install_requires = [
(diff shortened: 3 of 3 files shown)


jbednar:
Sounds good, thanks!

ianthomas23:
Confirmed that `conda install -c pyviz/label/dev datashader` today gives you `datashader 0.14.2` with `dask-core` and without `bokeh`.

ianthomas23:
Also confirmed that conda install -c conda-forge datashader today gives you datashader 0.14.2 with dask-core and without bokeh.

PR #1111, opened 2022-09-01
Fix dask dependencies

ianthomas23:
PR #1107 erroneously (my fault) reintroduced a heavy dask dependency that we have been trying to remove. This PR corrects the change.


ianthomas23:
I've confirmed locally that the dependencies are correct after installing via:
  * `pip install -ve .`
  * `pyctdev` using `pip`
  * `pyctdev` using `conda`

jbednar:
Thanks!

Issue #1112, opened 2022-09-01
Fully curate dependencies


ianthomas23:
The recent changes to remove unwanted dependencies has not gone well, e.g. #1111.

We need to fully curate all of the dependencies, for install, test and examples. This includes checking that we are not including a dependency that we don't use, and we are not excluding a dependency that we need. In some situations we may be relying on a dependency being present when we don't really need to.

The big one was `dask` leading to `bokeh` (#1104). Another that comes to mind is `spatialpandas` which used to be hard dependency and now isn't, but we still import it in core code and do not gracefully handle the situation if it is not installed.

This needs to be checked from both the `pip` and `conda` point of views via `pyctdev`, and ideally manually as well. All the CI tests are currently using `conda` so we should add at least one CI test run using `pip` on each of the 3 major platforms so that `pip` dependencies are covered by CI.

PR #1115, opened 2022-09-05
Remove unneeded future imports

ianthomas23:
This PR removes unneeded `__future__` imports of `absolute_import`, `division` and `print_function`. Support for Python 2.* required them and they are no longer necessary. Table of minimum versions at https://docs.python.org/3/library/__future__.html shows they have all been available since Python 3.0.

This is part of issue #1112.

### DIFF
--- a/datashader/__init__.py
+++ b/datashader/__init__.py
@@ -1,4 +1,4 @@
-from __future__ import absolute_import
+from __future__ import annotations
 
 from packaging.version import Version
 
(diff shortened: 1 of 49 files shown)


datashader/__init__.py
```diff
@@ -1,5 +1,3 @@
-from __future__ import absolute_import
```
mattpap, line 1: Perhaps instead of removing `__future__` imports altogether, it would be better to replace them with:
```py
from __future__ import annotations
```
for the sake of cohesion between holoviz projects.

ianthomas23: I was going to say that this is out of scope of the current changes as I am looking back fixing historic problems rather than looking forward making improvements, but I may as well do this now as it affects exactly the same lines of code. So yes, I will.

PR #1116, opened 2022-09-05
Curate dependencies

### LINKED ISSUE #1112: Fully curate dependencies

ianthomas23:
Closes #1112.

Significant changes here are:

1. `spatialpandas` is not a compulsory dependency (it hasn't been for a while) so I have modified the `import spatialpandas` code to gracefully handle the situation of it not being available.
2. I've removed `colorcet` as a compulsory dependency as it is not used anywhere in the code other than the tests and examples. I think this is the correct approach, but it may impact downstream projects which depend on it being included with `datashader`.
3. I've added a github action to test using `pip` (via the `pyctdev` `doit ecosystem=pip` approach) so that we are automatically testing both `pip` and `conda` installations.

I also removed `bcolz` as it is no longer maintained (https://github.com/Blosc/bcolz) but it is only used in `filetimes.py`. There are other possible removals here but I intend to replace the benchmarking code with a full `asv` benchmarking suite so I will address the other `filetimes.py` changes then.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -99,17 +99,20 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: ['ubuntu-latest', 'macos-latest', 'windows-latest']
+        os: ['ubuntu-latest', 'macos-latest']
         python-version: ["3.9"]
     steps:
       - name: Checkout source
         uses: actions/checkout@v2
         with:
-          fetch-depth: "100"
+          fetch-depth: 0
       - name: Install Python ${{ matrix.python-version }}
         uses: actions/setup-python@v4
         with:
           python-version: ${{ matrix.python-version }}
+      - name: Update setuptools
+        run: |
+          pip install --upgrade setuptools
       - name: Install pyctdev
         run: |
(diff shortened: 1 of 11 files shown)


ianthomas23:
Yes, your argument makes a lot of sense. It is not as if `colorcet` has a complicated set of dependencies anyway. I'll put it back in as a compulsory dependency.

PR #1122, opened 2022-09-20
Ensure spatial index _sindex is retained on dataframe copy

### LINKED ISSUE #1121: Datashader discards spatialpandas spatial indexer
If you are datashading a `spatialpandas` geometry such as a collection of polygons, it is usual to enable spatial indexing so that geometries that are outside of the `Canvas` range are rejected early, for performance reasons. The spatial indexer is stored on the `GeoDataFrame`'s extension array as a `_sindex` attribute. On the current `master` branch, if your `DataFrame` contains any columns that are not needed for datashading then the shallow copy of columns of the `DataFrame` that occurs in Datashader discards the `_sindex`.

The problem occurs in the third line here:
https://github.com/holoviz/datashader/blob/f1fbae82e7b30daad91f9b5d9205aa69dcdb3951/datashader/core.py#L1267-L1269

I assume the change of behaviour was caused at some time in the past when `pandas` tightened up this sort of shallow copying. Probably there is a recommended solution that needs to be in `spatialpandas` for the extension arrays. In the meantime there is the following workaround:
```python
cols_to_keep = _cols_to_keep(source.columns, glyph, agg)
if len(cols_to_keep) < len(source.columns):
    sindex = getattr(source.geometry.array, "_sindex", None)
    source = source[cols_to_keep]   # Warning: this drops the source.geometry.array._sindex
    if sindex is not None and getattr(source.geometry.array, "_sindex", None) is None:
        source.geometry.array._sindex = sindex
```

This is of course difficult to test as the output is identical with and without spatial indexing turned on, and the change only occurs internally in the `bypixel` function. I have been using the following test code to measure the performance:
```python
import datashader as ds
import spatialpandas.io as spdio
import time

directory = "/home/data/hv/nyc_buildings.parq"
x_range = (-8239401.166656778, -8231737.932910569)
y_range = (4971194.047138389, 4978243.635301908)

cvs = ds.Canvas(plot_width=800, plot_height=800, x_range=x_range, y_range=y_range)
df = spdio.read_parquet(directory)
df.build_sindex()

for repeat in range(3):
    s = time.perf_counter()
    agg = cvs.polygons(source=df, geometry="geometry", agg=ds.count())
    if repeat > 0:
        print("Time", time.perf_counter()-s)
```
If I run this against `master` it takes approximately 1.0 seconds for each `cvs.polygons` call. With the workaround above it is 0.07 seconds.

ianthomas23:
Fixes #1121.

The fix is to copy the `_sindex` if it is not copied automatically when we slice a `DataFrame`. We should probably consider this a temporary solution as the full solution would occur in `spatialpandas`, but with possible changes due to use of `awkward-pandas` we should hold off on that for the moment.

It wasn't possible to directly access the problem code as it occurred within a `bypixel` call. So I have pulled out the conversion code into a `_bypixel_sanitise` function which is tested on its own in the new `test_spatial_index_not_dropped` test.


jbednar (commented):
Looks like a reasonable approach, thanks.

PR #1124, opened 2022-09-22
Implement GPU-accelerated edge bundling

scottagfox:
FFT bundling is an alternative edge bundling method to Datashader's existing Edgehammer bundling implementation. It runs magnitudes faster than Edgehammer on larger graphs (10k+ edges) thanks to GPU-acceleration and cuFFT.

FFT bundling takes input in the form of two dataframes (cuDF or Pandas) containing normalized node coordinates and an edge list, and outputs a cuDF dataframe containing line segments.

Performance and the end result depend greatly on several parameters:

    Iterations: The maximum number of passes over the edge segments (bundling will end if kernel grows too small).
    Accuracy: Resolution of possible edge positions.
    Move Distance: How far an edge segment will be moved along the gradient.
    Kernel Size: Initial size of Gaussian kernel.
    Kernel Decay: Rate of Gaussian kernel decay.
    Resample Frequency: How often edges will be resampled.
    Resample Distance: How far apart edge segments can be from each other; resolution of an edge.
    Final Smooth: Number of smoothing passes to perform on the final graph.

The FFT edge bundling pipeline is similar to Edgehammer's with a few key differences.
- Edge segments are stored on GPU in a CuPy array (which is resized when resampling occurs).
- The gradient map that the edge segments are moved along is generated by convoluting the Fourier transforms of the density map and a Gaussian kernel.

    Pipeline:

    read edge segments from df
    initial sample
        generate gaussian kernel
        resample edges
        generate density map
        generate gradient map
        move edge segments
        smooth
    final smooth

FFT bundling is still missing some features supported by the Edgehammer bundling method and could use some additional features such as support for weighted graphs, and 'batching' edges to avoid memory constraints. @ianthomas23 mentioned introducing it as an "experimental" feature for now.

@exactlyallan


exactlyallan:
@ianthomas23 Would love it if you reviewed this and suggested next steps.

jbednar:
Looks cool, thanks! Before we could release it we'd need some documentation, preferably including comparisons with the existing edge bundling, and some tests.  I don't suppose NVidia has ever worked out a way we could get CUDA-enabled workers for our Github Actions so we can keep this working? :-)

exactlyallan:
@jbednar we are in the process on updating our DevOps to enable more testing with external libraries using GitHub Actions, though that would still only be accessible within NVIDIA. We'll work on getting some more docs for this!

Issue #1131, opened 2022-10-07
cudf test failure in test_log_axis_line with numpy 1.23


ianthomas23:
Running `DATASHADER_TEST_GPU=1 pytest datashader/tests/test_dask.py::test_log_axis_line` gives
```
datashader/tests/test_dask.py::test_log_axis_line[ddf0] FAILED           [100%]

<snip> 
../../.miniconda/envs/rapids/lib/python3.9/site-packages/cudf/core/scalar.py:156: in _preprocess_host_value
    value = to_cudf_compatible_scalar(value, dtype=dtype)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

val = NotImplemented, dtype = <class 'numpy.bool_'>

    def to_cudf_compatible_scalar(val, dtype=None):
        """
        Converts the value `val` to a numpy/Pandas scalar,
        optionally casting to `dtype`.
    
        If `val` is None, returns None.
        """
    
        if cudf._lib.scalar._is_null_host_scalar(val) or isinstance(
            val, cudf.Scalar
        ):
            return val
    
        if not cudf.api.types._is_scalar_or_zero_d_array(val):
>           raise ValueError(
                f"Cannot convert value of type {type(val).__name__} "
                "to cudf scalar"
            )
E           ValueError: Cannot convert value of type NotImplementedType to cudf scalar

../../.miniconda/envs/rapids/lib/python3.9/site-packages/cudf/utils/dtypes.py:247: ValueError
------------
```
this is with:
```
cuda-python               11.7.0           py39h3fd9d12_0    nvidia
cudatoolkit               11.5.1           hcf5317a_9        nvidia
cudf                      22.08.00         cuda_11_py39_gb71873c701_0    rapidsai
cupy                      10.6.0           py39hc3c280e_0    conda-forge
dask                      2022.7.1         pyhd8ed1ab_0      conda-forge
dask-core                 2022.7.1         pyhd8ed1ab_0      conda-forge
dask-cudf                 22.08.00         cuda_11_py39_gb71873c701_0    rapidsai
datashader                0.14.2.post11+gb0e120c.dirty       pypi_0      pypi
numba                     0.56.2           py39h61ddf18_0    conda-forge
numpy                     1.23.1           py39hf838250_0  
numpy-base                1.23.1           py39h1e6e340_0
```

[...]

PR #1132, opened 2022-10-07
Consistent test_gpu flag

### LINKED ISSUE #1130: GPU test problems if run without DATASHADER_TEST_GPU being set
If you have the required GPU libraries installed (cupy, cudf) and run the test suite with either `DATASHADER_TEST_GPU=0` or `DATASHADER_TEST_GPU=1` then the correct tests are run. But if you not specify `DATASHADER_TEST_GPU` at all then some GPU tests run but fail. This is not a problem with the tests themselves but is the code used to identify if GPU tests should be run which is different in different test code files, so we have tests which run because they determine that GPU support is available, but the test comparisons fail because they think that GPU support is not available.

ianthomas23:
Fixes #1130.

This simplifies the `test_gpu` flag so that it is boolean. Previously it could be `True`, `False` or `None` (meaning unspecified) and the `None` option was supposed to behave the same as `False` but with different test skipped messages. But it wasn't handled consistently, now it is. It still defaults to not running GPU tests. I am not sure this is the best default, but it can be dealt with in a later PR.

On a system with a GPU and with `cupy` and `cudf` installed, running the tests in either of these ways runs exactly the same tests (i.e. ignores the all GPU tests):
```
pytest datashader/tests
DATASHADER_TEST_GPU=0 pytest datashader/tests
```
and to run all tests including GPU ones use
```
DATASHADER_TEST_GPU=1 pytest datashader/tests
```
 
Also tested all 3 options in an environment without `cupy` and `cudf` installed, and if you set `DATASHADER_TEST_GPU=1` then the `test_gpu_dependencies` test fails as expected.

After this PR, all tests run on CPU and GPU on my dev machine (Ubuntu 22.04 x86_64, Nvidia Quadro T1000) with the exception of that identified in issue #1131, which can be manually disabled using the `-k "not log_axis_line"` command-line argument.

### DIFF
--- a/datashader/tests/benchmarks/test_canvas.py
+++ b/datashader/tests/benchmarks/test_canvas.py
@@ -5,10 +5,7 @@ import pandas as pd
 
 import datashader as ds
 
-if "DATASHADER_TEST_GPU" in os.environ:
-    test_gpu = bool(int(os.environ["DATASHADER_TEST_GPU"]))
-else:
-    test_gpu = None
+test_gpu = bool(int(os.getenv("DATASHADER_TEST_GPU", 0)))
 
 
 @pytest.fixture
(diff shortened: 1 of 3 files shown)


jbednar (approved):
Looks good, thanks!

PR #1125, opened 2022-09-26
Pin nbformat to 5.4.0

ianthomas23:
CI is failing on linting the example notebooks with `nbformat` 5.5.0 and 5.6.0. Pinning to 5.4.0 temporarily to pass CI.

PR #1139, opened 2022-10-19
Remove nbformat version pin

ianthomas23:
The `nbformat` version pin introduced in PR #1125 is no longer needed as the underlying problem has been fixed in https://github.com/pyviz-dev/nbsmoke/pull/63 (thanks @maximlt).

I am removing the `nbformat` dependency completely as that is how it was before the version pin. The dependency is picked up via `nbconvert` and `nbsmoke`.


ianthomas23:
Superseded by #1170.

Issue #1133, opened 2022-10-13
Antialiasing fails on uint32 with a mean reduction


jlstevens:
To reproduce, note that the following code works with `line_width=0`:

```python
import datashader as ds
import pandas as pd
cvs = ds.Canvas()
cvs.line(pd.DataFrame({'x': [1, 2, 3], 'y': [2, 3, 4], 'z': np.array([True, False, True]).astype('uint32')},
                      columns=['x', 'y', 'z']), x='x', y='y', agg=ds.mean('z'), line_width=0)
```

But fails with a numba error with a non-zero `line_width`:

<details>
---------------------------------------------------------------------------
TypingError                               Traceback (most recent call last)
Input In [20], in <cell line: 5>()
      1 import datashader as ds
      3 cvs = ds.Canvas()
----> 5 cvs.line(pd.DataFrame({'x': [1, 2, 3], 'y': [2, 3, 4], 'z': np.array([True, False, True]).astype('uint32')},
      6                       columns=['x', 'y', 'z']), x='x', y='y', agg=ds.mean('z'), line_width=1)

File ~/Desktop/development/knauf_panel/stocks_example/envs/default/lib/python3.8/site-packages/datashader/core.py:449, in Canvas.line(self, source, x, y, agg, axis, geometry, line_width, antialias)
    446     # Switch agg to floating point.
    447     agg = rd._reduction_to_floating_point(agg)
--> 449 return bypixel(source, self, glyph, agg)

File ~/Desktop/development/knauf_panel/stocks_example/envs/default/lib/python3.8/site-packages/datashader/core.py:1281, in bypixel(source, canvas, glyph, agg)
   1279 with np.warnings.catch_warnings():
   1280     np.warnings.filterwarnings('ignore', r'All-NaN (slice|axis) encountered')
-> 1281     return bypixel.pipeline(source, schema, canvas, glyph, agg)

File ~/Desktop/development/knauf_panel/stocks_example/envs/default/lib/python3.8/site-packages/datashader/utils.py:109, in Dispatcher.__call__(self, head, *rest, **kwargs)
    107 typ = type(head)
    108 if typ in lk:
--> 109     return lk[typ](head, *rest, **kwargs)
    110 for cls in getmro(typ)[1:]:
    111     if cls in lk:

[...]

PR #1142, opened 2022-11-01
Fit antialiased line code within usual numba/dask framework

ianthomas23:
This PR modifies both the antialiased line implementation and the low level numba/dask framework code so that the former now works within the latter rather than having its own specific code path. This is necessary before:

1. Support for compound `Reduction`s with antialiased lines (issue #1133)
2. Antialiased lines on the GPU. This is part-implemented herein, but it turns out that compound `Reduction` support is needed first.

There is no functional change from the end user's point of view.

Catalogue of changes:

1. New `antialiased` kwarg for `compile_components()`, the various `make_*` functions, and the `pipeline` functions. The latter now have two optional bool args `antialias` and `cuda` so I have forced them to be kwargs to avoid potential ambiguity.
2. `Glyph` class stores a bool `antialias` attribute. This is only used within `_expand_aggs_and_cols` which is the function called when we use the `@expand_aggs_and_cols` decorator.
3. `antialias` flag passes through the various `Line` classes and line rendering functions in `line.py`.
4. `_full_antialias` now works with the `@expand_aggs_and_cols` decorator which is necessary for it to work on the GPU. 
5. `_full_antialias` takes a `workspace` which is an array (`numpy` or `cupy`) of length 8. CUDA cannot create arrays so it has to be passed in from outside the function, this should eventually (after other changes) give increased performance for all antialiased lines due to many fewer array allocations and deallocations.
6. Remove the `any_f32` and `count_f32` reductions completely.
7. Reduction classes modified so that their dtypes are not static. Now the same reduction can be used for antialiased and non-antialiased lines, and it will create the appropriately dtyped aggs and use the correct `append` functions. There are some refactors here that can be done in the future to reduce the duplication of the various `append` and `create` functions, but this PR is large enough as it is.

### DIFF
--- a/datashader/compiler.py
+++ b/datashader/compiler.py
@@ -14,7 +14,7 @@ __all__ = ['compile_components']
 
 
 @memoize
-def compile_components(agg, schema, glyph, cuda=False):
+def compile_components(agg, schema, glyph, cuda=False, antialias=False):
     """Given a ``Aggregation`` object and a schema, return 5 sub-functions.
 
     Parameters
@@ -53,7 +53,7 @@ def compile_components(agg, schema, glyph, cuda=False):
     bases = list(unique(concat(r._build_bases(cuda) for r in reds)))
     dshapes = [b.out_dshape(schema) for b in bases]
     # List of tuples of (append, base, input columns, temps)
-    calls = [_get_call_tuples(b, d, schema, cuda) for (b, d) in zip(bases, dshapes)]
+    calls = [_get_call_tuples(b, d, schema, cuda, antialias) for (b, d) in zip(bases, dshapes)]
     # List of unique column names needed
     cols = list(unique(concat(pluck(2, calls))))
     # List of temps needed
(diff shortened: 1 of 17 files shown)


ianthomas23:
Rebased to include #1142 to avoid unrelated CI failures.

jbednar:
Ian and I went over this PR in a shared meeting, and I'm ok with these changes after a trivial suggestion to rename "workspace" to something like "buffer" so that it's clearer that it's simply a block of memory for the callee to use as it likes. Thanks, @ianthomas23 !

ianthomas23:
`workspace` renamed to `buffer`, and added some more explanation as comments in the code.

For clarity, a `buffer` is used entirely within the lifetime of a `_full_antialias` call, the values are set once and then read multiple times. It doesn't matter what the buffer contains before or after `_full_antialias` is called. `buffer`s are created at a higher level so that there is one per CPU/GPU thread to avoid cross-talk.

PR #1143, opened 2022-11-03
Fix CI test running on macOS

ianthomas23:
CI is failing on macOS with the same symptoms as in https://github.com/holoviz/holoviews/pull/5498.

Trying out the same (temporary) solution here.


ianthomas23:
I am merging this as it fixes the CI, albeit temporarily.

PR #1145, opened 2022-11-08
Refactor stage 2 aggregation for antialiased lines

ianthomas23:
This PR refactors the second stage aggregation for antialiased lines so that the mathematical aggregation operations are passed into the line rendering functions from the `Reduction` classes, and in such a way that multiple reductions can be supported. There is no functional change from the end users point of view.

This is the final set of changes required before issue #1133 can be implemented for antialiased lines on the CPU.

Implementation details

- An `antialias_stage_2` tuple is created from the supplied `Reductions` and passed into the line rendering functions. This tuple contains one item per `Reduction`, and each item is a 2-tuple of `AntialiasCombination` (an enum specifying the mathematical combination operator) and a `zero_value` which is the initial value of the aggregation (currently only `np.nan` but will be `0` for e.g. `_sum_zero`).

### DIFF
--- a/datashader/compiler.py
+++ b/datashader/compiler.py
@@ -15,7 +15,7 @@ __all__ = ['compile_components']
 
 @memoize
 def compile_components(agg, schema, glyph, *, antialias=False, cuda=False):
-    """Given a ``Aggregation`` object and a schema, return 5 sub-functions
+    """Given an ``Aggregation`` object and a schema, return 5 sub-functions
     and information on how to perform the second stage aggregation if
     antialiasing is requested,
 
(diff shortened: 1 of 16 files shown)


datashader/compiler.py
```diff
@@ -15,7 +15,9 @@
 
 @memoize
 def compile_components(agg, schema, glyph, *, antialias=False, cuda=False):
-    """Given a ``Aggregation`` object and a schema, return 5 sub-functions.
+    """Given a ``Aggregation`` object and a schema, return 5 sub-functions
```
jbednar, line 18:
```suggestion
    """Given an ``Aggregation`` object and a schema, return 5 sub-functions
```

jbednar (approved):
It seems like this is necessary, and it doesn't make things appreciably worse, so go for it!

PR #1146, opened 2022-11-14
Support compound reductions for antialiased lines on the CPU

### LINKED ISSUE #1133: Antialiasing fails on uint32 with a mean reduction

ianthomas23:
This implements compound reductions such as `mean` and `summary` for antialiased lines on the CPU only. The only reductions not yet covered are `std` and `var`, but there is a possible solution for these being investigated. There is no support for antialiased lines on the GPU, and this will not occur soon as it requires significant changes.

Demonstration of antialiased lines and various reductions:
```python
import datashader as ds
import numpy as np
import pandas as pd

df = pd.DataFrame(dict(ystart=[0.2, 0, 1, 0.8], yend=[0.2, 1, 0, 0.8], value=[1, 2, 3, 4]))

cvs = ds.Canvas(plot_width=200, plot_height=150, x_range=(-0.1, 1.1), y_range=(-0.1, 1.1))
kwargs = dict(source=df, x=np.asarray([0, 1]), y=["ystart", "yend"], axis=1, line_width=15)

for i, reduction in enumerate([
    ds.any(), ds.count(), ds.sum("value"), ds.min("value"), ds.max("value"),
    ds.first("value"), ds.last("value"), ds.mean("value")
]):
    agg = cvs.line(agg=reduction, **kwargs)
    im = tf.shade(agg, how="linear")
    ds.utils.export_image(im, f"temp{i}")
```

Output images for `any`, `count` and `sum`
temp0 temp1 temp2

`min` and `max`
temp3 temp4

`first`, `last`, `mean`
temp5 temp6 temp7

Two observations:

- Note the halos around crossing edges of `min`, `first` and `last`.
- `mean` loses the antialiased edges as it essentially performs a block fill of the pixels that would be rendered partially transparent. This is understandable from the maths of `mean = sum / count`. For a single antialiased line of a constant `value` the pixel values for `sum` equal `count * value`, hence `mean = value`. Note that this is also subject to numerical rounding (not seen here) as typically an antialiased `count` is calculated using `float32` whereas a `sum` uses `float64`.

Closes #1133.

### DIFF
--- a/datashader/compiler.py
+++ b/datashader/compiler.py
@@ -59,8 +59,26 @@ def compile_components(agg, schema, glyph, *, antialias=False, cuda=False):
     # List of base reductions (actually computed)
     bases = list(unique(concat(r._build_bases(cuda) for r in reds)))
     dshapes = [b.out_dshape(schema, antialias) for b in bases]
+
+    # Information on how to perform second stage aggregation of antialiased lines,
+    # including whether antialiased lines self-intersect or not as need a single
+    # value for this even for a compound reduction. This is by default True, but
+    # is False if a single constituent reduction requests it.
+    if antialias:
+        self_intersect, antialias_stage_2 = make_antialias_stage_2(reds, bases)
+        if cuda:
+            import cupy
+            array_module = cupy
+        else:
+            array_module = np
+        antialias_stage_2 = antialias_stage_2(array_module)
+    else:
+        self_intersect = False
+        antialias_stage_2 = False
+
     # List of tuples of (append, base, input columns, temps)
-    calls = [_get_call_tuples(b, d, schema, cuda, antialias) for (b, d) in zip(bases, dshapes)]
+    calls = [_get_call_tuples(b, d, schema, cuda, antialias, self_intersect)
+             for (b, d) in zip(bases, dshapes)]
     # List of unique column names needed
     cols = list(unique(concat(pluck(2, calls))))
     # List of temps needed
@@ -71,17 +89,6 @@ def compile_components(agg, schema, glyph, *, antialias=False, cuda=False):
     append = make_append(bases, cols, calls, glyph, isinstance(agg, by), antialias)
     combine = make_combine(bases, dshapes, temps, antialias)
     finalize = make_finalize(bases, agg, schema, cuda)
-
-    if antialias:
-        antialias_stage_2 = make_antialias_stage_2(reds, bases)
-        if cuda:
-            import cupy
-            array_module = cupy
-        else:
-            array_module = np
-        antialias_stage_2 = antialias_stage_2(array_module)
-    else:
-        antialias_stage_2 = False
 
     return create, info, append, combine, finalize, antialias_stage_2
 
@@ -96,10 +103,10 @@ def traverse_aggregation(agg):
         yield agg
 
 
-def _get_call_tuples(base, dshape, schema, cuda, antialias):
+def _get_call_tuples(base, dshape, schema, cuda, antialias, self_intersect):
     # Comments refer to usage in make_append()
     return (
-        base._build_append(dshape, schema, cuda, antialias),  # func
+        base._build_append(dshape, schema, cuda, antialias, self_intersect),  # func
         (base,),  # bases
         base.inputs,  # cols
         base._build_temps(cuda),  # temps
@@ -217,6 +224,9 @@ def make_finalize(bases, agg, schema, cuda):
 
 def make_antialias_stage_2(reds, bases):
     # Only called if antialias is True.
+
+    # Prefer a single-stage antialiased aggregation, but if any requested
+    # reduction requires two stages then force use of two for all reductions.
     self_intersect = True
     for red in reds:
         if red._antialias_requires_2_stages():
@@ -226,4 +236,4 @@ def make_antialias_stage_2(reds, bases):
     def antialias_stage_2(array_module):
         return tuple(zip(*concat(b._antialias_stage_2(self_intersect, array_module) for b in bases)))
 
-    return antialias_stage_2
+    return self_intersect, antialias_stage_2
--- a/datashader/compiler.py
+++ b/datashader/compiler.py
@@ -61,7 +61,7 @@ def compile_components(agg, schema, glyph, *, antialias=False, cuda=False):
     dshapes = [b.out_dshape(schema, antialias) for b in bases]
 
     # Information on how to perform second stage aggregation of antialiased lines,
-    # including whether antialiased lines self-intersect or not as need a single
+    # including whether antialiased lines self-intersect or not as we need a single
     # value for this even for a compound reduction. This is by default True, but
     # is False if a single constituent reduction requests it.
     if antialias:
--- a/datashader/glyphs/line.py
+++ b/datashader/glyphs/line.py
@@ -5,9 +5,10 @@ from toolz import memoize
 
 from datashader.enums import AntialiasCombination
 from datashader.glyphs.points import _PointLike, _GeometryLike
-from datashader.utils import (isnull, isreal, ngjit, nanmax_in_place,
-                              nanmin_in_place, nansum_in_place, parallel_fill)
-from numba import cuda
+from datashader.utils import (
+    isnull, isreal, nanfirst_in_place, nanlast_in_place, nanmax_in_place,
+    nanmin_in_place, nansum_in_place, ngjit, parallel_fill)
+from numba import cuda, literal_unroll
 import numba.types as nb_types
 
 
@@ -31,12 +32,26 @@ def _two_stage_agg(antialias_stage_2):
         # Not using antialiased lines, doesn't matter what is returned.
         return False, False
 
-    if len(antialias_stage_2[0]) > 1:
-        raise NotImplementedError("Currently only support single antialiased reductions")
-    comb = antialias_stage_2[0][0]
+    # A single combination in (SUM_2AGG, FIRST, LAST, MIN) means that a 2-stage
+    # aggregation will be used, otherwise use a 1-stage aggregation that is
+    # faster.
+    use_2_stage_agg = False
+    for comb in antialias_stage_2[0]:
+        if comb in (AntialiasCombination.SUM_2AGG, AntialiasCombination.MIN,
+                    AntialiasCombination.FIRST, AntialiasCombination.LAST):
+            use_2_stage_agg = True
+            break
 
-    overwrite = comb != AntialiasCombination.SUM_1AGG
-    use_2_stage_agg = comb in (AntialiasCombination.SUM_2AGG, AntialiasCombination.MIN)
+    # Boolean overwrite flag is used in _full_antialias() is True to overwrite
+    # pixel values (using max of previous and new values) or False for the more
+    # complicated correction algorithm. Prefer overwrite=True for speed, but
+    # any SUM_1AGG implies overwrite=False.
+    overwrite = True
+    for comb in antialias_stage_2[0]:
+        if comb == AntialiasCombination.SUM_1AGG:
+            overwrite = False
+            break
+
     return overwrite, use_2_stage_agg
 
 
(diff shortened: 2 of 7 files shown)


datashader/glyphs/line.py
```diff
@@ -1107,18 +1152,13 @@ def cpu_antialias_2agg_impl(sx, tx, sy, ty, xmin, xmax, ymin, ymax, xs, ys,
             if ncols == 1:
                 return
 
-            if j == 0:
-                accum_aggs = [aggs[k].copy() for k in range(n_aggs)]
-            else:
-                for k in range(n_aggs):
-                    _combine_in_place(accum_aggs[k], aggs[k], antialias_combinations[k])
+            _aa_stage_2_accumulate(aggs_and_accums, antialias_combinations)
```
ianthomas23, line 1155: Note the increase in readability of this function and others like it by using separate jit-able functions like `_aa_stage_2_accumulate`. This also allows functions like this to have their `@expand_aggs_and_cols` decorators reinstated, in line with the other various low-level line rendering functions.

datashader/compiler.py
```diff
@@ -59,8 +59,26 @@ def compile_components(agg, schema, glyph, *, antialias=False, cuda=False):
     # List of base reductions (actually computed)
     bases = list(unique(concat(r._build_bases(cuda) for r in reds)))
     dshapes = [b.out_dshape(schema, antialias) for b in bases]
+
+    # Information on how to perform second stage aggregation of antialiased lines,
+    # including whether antialiased lines self-intersect or not as need a single
+    # value for this even for a compound reduction. This is by default True, but
```
jbednar, line 65:
```suggestion
    # including whether antialiased lines self-intersect or not as we need a single
    # value for this even for a compound reduction. This is by default True, but
```

?

jbednar (approved):
Thank you for taking this on. My brain wants to crawl into a corner and die when I read this code (before and after this PR, so not your fault!). 

> Note the halos around crossing edges of min, first and last.

I can see those, but can't quite get ahold of the mathematics involved to see if there is any way around that.

> mean loses the antialiased edges as it essentially performs a block fill of the pixels that would be rendered partially transparent. This is understandable from the maths of mean = sum / count. For a single antialiased line of a constant value the pixel values for sum equal count * value, hence mean = value.

Again, I'm not sure if there is a way around this?  Seems like value should be e.g. half the actual value, for the smoothed bits of the edge?

ianthomas23:
> Again, I'm not sure if there is a way around this? Seems like value should be e.g. half the actual value, for the smoothed bits of the edge?

There is no workaround for this, it would need a more complicated implementation as we don't store the information required to do this. We would need to separate out the antialias `weighting` (what would conventionally be called the `alpha` but trying not to use that term here) from the normal aggregate numbers. For the `count` of a single line segment the `count` agg would be 1 for every pixel in the line right up to the edges, and the `weighting` would be 1 along the middle of the line, down to just about 0 at the extreme edges. The 2D agg returned to the user would be the product of the two, giving the same number as we currently return (for a single line).

For a compound reduction of a `count` and `sum` we can imaging 3 aggs, `count`, `sum` and `weighting`. For the `count` we'd return `count*weighting` to the user, for the `sum` we'd return `sum*weighting`, and for the `mean` we'd return `(sum/count)*weighing`, giving you exactly what you want.

In general, we couldn't just have a single `weighting` per `canvas.line` call, we'd need a `weighing` per agg. So maybe we'd think of changing each of the current agg's shapes from `(height, width)` to `(height, width, 2)`, so the agg is attached to its `weighting`.

We would need to decide on appropriate mathematics to combine say `(value1, weighting1)` with `(value2, weighting2)` for a particular pixel. Let's assume we want a linear combination, so the combined `value` must be `value1*weighting1 + value2*weighting2`. Or is it? It might be more sensible to say we want the combined `value` and `weighting` so that `value*weighting = value1*weighting1 + value2*weighting2` because the `value` needs to be stored unweighted (the whole purpose of this approach is to keep the weighting separate from the value). This maths is not dissimilar to combined rendering of RGB and A separately, but there we have the concept of rendering on top of existing colors in a non-commutative way. So it is more complicated than that!

In summary, to obtain a `mean` that more in line with what is expected would require:

1. Twice as much memory for the 3D aggs.
2. Slightly more complicated code, but only at the very highest and lowest levels as we'd still be scanning the `source` dataframe just once, only with 3D aggs.
3. Thought about appropriate maths to combine values and weights.

jbednar:
You say there is no workaround, and then proceed to lay out a workaround. :-) Please open a separate issue proposing that work, emphasizing what limitations antialiasing will have until it is done, but then I don't think there's a particular reason to undertake it at this time. Do mention the limitation in the docs, though, and I guess link to the issue. Thanks!

ianthomas23:
On rereading my previous comment it sounds more positive than I intended it to be! It is not a workaround because it doesn't work; I don't believe there is any maths that can combine `values` and `weights` in a way that works for anything other than the simplest situations. But yes, it is worth writing it up as an issue for further consideration.

PR #1154, opened 2022-12-15
Validate calculated log canvas range

### LINKED ISSUE #340: Bug when using x_axis_type='log' and y_axis_type='log'
I am attempting to plot about 430,000 datapoints using datashader. 

The data looks like:

screen shot 2017-05-03 at 9 52 27 am

I plot using:

    plot_width=600
    plot_height=600
    x_range=(0,100)
    y_range=(0,100)

    cvs = ds.Canvas(
        plot_width=plot_width, 
        plot_height=plot_height, 
        x_range=x_range, 
        y_range=y_range, 
        x_axis_type='log', 
        y_axis_type='log')
    agg = cvs.points(df, 'x', 'y', ds.sum('count'))
    img = tf.shade(agg, cmap=viridis, how='eq_hist')

Note that I'm using x_range=(0,100) and y_range=(0,100) with x_axis_type='log' and y_axis_type='log' . This results in the following image:

download 64

If I use a linear scale for both axis, it renders correctly:

download 65

However, to make sense of the data, it needs to be log. If I set x_range and y_range to (0.01, 100), I get the image I need:

download 66

ianthomas23:
Fixes #340, producing a sensible error message rather than an empty plot.

Previously a `Canvas` validated its `x_range` and `y_range` if they were explicitly specified by the user. But if they were calculated from the data limits, no validation occurred.

### DIFF
--- a/datashader/core.py
+++ b/datashader/core.py
@@ -131,8 +131,7 @@ class LogAxis(Axis):
         if range is None:
             # Nothing to check if no range
             return
-        low, high = map(self.mapper, range)
-        if not (np.isfinite(low) and np.isfinite(high)):
+        if range[0] <= 0 or range[1] <= 0:
             raise ValueError('Range values must be >0 for logarithmic axes')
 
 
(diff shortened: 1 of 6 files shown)


jbednar (approved):
Looks good, thanks!

Issue #1126, opened 2022-09-26
Directly support HoloViews-style inspect operations


jbednar:
When Datashader renders a large dataset, a human being is usually able to see patterns and interesting datapoints that merit further investigation. Unfortunately, the rendered image does not provide any easy means of doing so, as the original datapoints have all been reduced to pixels (or more accurately, to scalar accumulated values in bins of a 2d histogram). To support investigation of interesting features, HoloViews implements a series of "inspect" operations that query the original dataset after a selection or hover event on the rasterized data. E.g. `inspect_points` in https://examples.pyviz.org/ship_traffic will query the original dataset to show hover and other information about the original datapoints being visualized. However, going back to the original dataset is quite slow, because it requires traversing either the entire dataset or (for a spatially indexed data structure) at least a chunk of the dataset, which makes the interface unpleasant and awkward and thus eliminating certain types of interactivity.

Datashader can collect multiple aggregations on a single pass through the data, so I suggest that we support an accumulation mode that gathers datapoint indexes rather than datapoints, so that hover and drilldown information can be supported instantaneously. Of course, arbitrarily many datapoints can be aggregated into a single pixel, while any practical aggregation can only accumulate a fixed number of indexes per pixel. Still, that's already how the `inspect_` operations work; they discard all but a configurable number of results, which is fine for linking to one or two examples per pixel, and allows single-datapoint precision with enough zooming in. By default I'd suggest accumulating the index of the minimum and the maximum value per pixel, but even just keeping the first or last datapoint for that pixel would be useful. 

If we keep at least three datapoints per pixel (e.g. min, max, and one other) we'd be able to distinguish between complete and incomplete inspection data for that pixel (i.e. are these the only points? Yes, if there are 2 or fewer; unclear otherwise). Seems to me that we should be able to have a fully responsive, fully inspectable rendering of a dataset at low computational and memory cost using this method.

jlstevens:
After some discussion, I think we agreed that a `first` and `last` index (per bin) aggregate makes sense and that a `where` aggregator (e.g. 'show the ship with the highest tonnage value that contributed to this pixel') would be nice too.

The only other point that I think is important is that you need to know the count because that can give you the context to know whether the 'first' or 'last' index is unique or a random sample that contributed to the pixel value (ignoring the possibility that it is meaningful due to sorting e.g. by time)

jbednar:
[...]

PR #1155, opened 2022-12-16
Add new where reduction

ianthomas23:
This partially implements issue #1126, adding a new `where` reduction that accepts either a `max` or `min` reduction. Best illustrated via an example:
```python
import datashader as ds
import numpy as np
import pandas as pd

x = np.arange(2)
df = pd.DataFrame(dict(
    y_from = [0.0, 1.0, 0.0, 1.0, 0.0],
    y_to   = [0.0, 1.0, 1.0, 0.0, 0.5],
    value  = [1.1, 3.3, 5.5, 2.2, 4.4],
    other  = [-55, -77, -99, -66, -88],
))

canvas = ds.Canvas(plot_height=3, plot_width=5)
agg = canvas.line(
    source=df, x=x, y=["y_from", "y_to"], axis=1,
    agg=ds.where(ds.max("value"), "other"),
)

print(agg)
```
which outputs
```
<xarray.DataArray (y: 3, x: 5)>
array([[-99., -88., -55., -66., -66.],
       [ nan, -99., -99., -88., -88.],
       [-77., -77., -77., -99., -99.]])
Coordinates:
  * x        (x) float64 0.1 0.3 0.5 0.7 0.9
  * y        (y) float64 0.1667 0.5 0.8333
```
You can think of this using the `max('value')` reduction as normal, but then returning the corresponding values from the `'other'` column rather that the `value` column.

What it currently supports:

- `where` takes either a `min` or `max` selector reduction.
- Works on CPU (not GPU), with or without dask.
- Works with antialiased lines.
- Cannot be used within a `summary` or categorical `by` reduction.

Note that there is no support for use of `first` and `last` within a `where` because there is no advantage in doing this, you can just use the `first` or `last` directly on their own.

Future improvements:

- Support within categorical reductions.
- Support for GPU.
- If `lookup_column` is not specified, use the index of the row in the supplied DataFrame.
- New `max_n`, `min_n`, `first_n`, `last_n` reductions.

All of these are possible but fiddly to implement, so I would rather have partial functionality available for users to experiment with and I can add these improvements over time.

Currently some combinations of lines and dask give different results depending on the number of dask partitions, but this has always been the situation and is no worse here.


datashader/reductions.py
```diff
@@ -1159,6 +1223,75 @@ def _finalize(bases, **kwargs):
         raise NotImplementedError("mode is currently implemented only for rasters")
 
 
+class where(FloatingReduction):
+    def __init__(self, selector: Reduction, lookup_column: str):
```
ianthomas23, line 1227: Needs a docstring with a good example.

jbednar:
Thanks! Can you clarify the current status of types? I.e. can you return an integer aggregate when testing on a float condition?

ianthomas23:
`where` always returns a `float64` with `nan`s to represent no data, just as `min`, `max`, `first`, `last` etc reductions.

jbednar:
Ok, I guess we'll need to deal with datatype issues when we support using the Pandas index as the "column" (actually just imputed values that act like a column, hence needing special support).

ianthomas23:
Rebased on top of main to pick up the CI fixes.

datashader/reductions.py
ianthomas23: I have added a docstring. It is somewhat tortuous English when written in a generic form so I have added a concrete example that should help.

I am using the argument name `selector` because long-term I would like to divide the `Reduction` class hierarchy into two:

- `Selection` reductions that use values from a column without modifying them, e.g. `first` and `max`.
- `Combination` reductions that do some form of mathematical combination of the values from a column, e.g. `mean`, `std`, `count`.

Given that division, a `where` reduction will accept a `Selection` but not a `Combination` to select values from the `lookup_column`.

ianthomas23:
The reduction in coverage is mostly due to changes to the CUDA `append` functions and such CUDA code is not run in github actions.

ianthomas23:
Pinging @jbednar. I'd like to merge this and add the extra functionality (such as use of a virtual integer row index) as separate PRs.

datashader/reductions.py
```diff
@@ -507,6 +531,10 @@ def __init__(self, cat_column, reduction=count()):
             self.categorizer = category_codes(cat_column)
         else:
             raise TypeError("first argument must be a column name or a CategoryPreprocess instance")
+
+        if isinstance(reduction, where):
+            raise TypeError("'by' reduction cannot use a 'where' reduction")
```
jbednar, line 536: Can you make "use" more specific? Cannot accept, does not support, etc.? It's confusing to read about reductions using reductions.

datashader/reductions.py
jbednar: Ok, sounds good. Agreed that it's tortuous, but the example does help a lot.

jbednar (approved):
Those look like some painful changes, and it's great to have them behind us! Thanks, and it looks good to me!

datashader/reductions.py
ianthomas23: Done.

PR #1162, opened 2023-01-06
nopython=True everywhere

ianthomas23:
This PR forces all use of `numba.jit` to be in `nopython=True` mode, as recommended by `numba` maintainer @stuartarchibald. The changes only affect the edge bundling code.

Test suite passes locally on CPU and GPU.

### DIFF
--- a/datashader/bundling.py
+++ b/datashader/bundling.py
@@ -24,7 +24,6 @@ try:
 except Exception:
     skimage = None
 
-import numba as nb
 import numpy as np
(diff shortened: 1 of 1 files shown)


stuartarchibald (approved):
Looks good, thanks for supplying the patch. This should future-proof `datashader` against changes being made for https://github.com/numba/numba/issues/3907.

jbednar (approved):

PR #1163, opened 2023-01-06
Use holoviz_task install action

ianthomas23:
Trying out use of holoviz_task install github action to try to solve CI build and test issues.


philippjfr:
Did this really work on the first try? 🤯

ianthomas23:
> Did this really work on the first try? 🤯

It did!

PR #1168, opened 2023-01-19
Return early in eq_hist() if all data masked out

### LINKED ISSUE #1166: shade(..., how="eq_hist") with no valid data raises index out of bounds error
First reported in holoviews as holoviz/holoviews#5595.

A simpler datashader-only reproducer is:
```python
import numpy as np
import xarray as xr

agg = xr.DataArray(
    data=np.zeros((2, 2, 2), dtype=np.uint32),
    coords=dict(y=[0, 1], x=[0, 1], cat=['a', 'b']),
)
im = tf.shade(agg, how='eq_hist')
print(im)
```
which produces:
```
  File "/Users/iant/github_temp/datashader_temp/empty_cdf.py", line 9, in <module>
    im = tf.shade(agg, how='eq_hist')
  File "/Users/iant/github/datashader/datashader/transfer_functions/__init__.py", line 721, in shade
    return _colorize(agg, color_key, how, alpha, span, min_alpha, name, color_baseline, rescale_discrete_levels)
  File "/Users/iant/github/datashader/datashader/transfer_functions/__init__.py", line 436, in _colorize
    a = _interpolate_alpha(data, total, mask, how, alpha, span, min_alpha, rescale_discrete_levels)
  File "/Users/iant/github/datashader/datashader/transfer_functions/__init__.py", line 474, in _interpolate_alpha
    a_scaled = _normalize_interpolate_how(how)(total - offset, mask)
  File "/Users/iant/github/datashader/datashader/transfer_functions/__init__.py", line 220, in eq_hist
    cdf = cdf / float(cdf[-1])
IndexError: index -1 is out of bounds for axis 0 with size 0
```
The problem is caused by passing an `xr.DataArray` that contains all values that will be masked out (i.e. zeros or nans depending on context) to `transfer_functions.image(..., how="eq_hist")`. This calls the `eq_hist()` function that is unable to determine a cumulative distribution function as there is no valid data.

The fix is to detect if all the supplied data to `eq_hist()` is masked out and return to the caller early.

ianthomas23:
Fixes #1166.

If pass an fully-masked array to `eq_hist()` function, return early with an array of `NaN`s. This avoids calculating the cumulative distribution function of no data and the resultant `IndexError` because the CDF has no contents.

Passes new test locally on both CPU and GPU.

### DIFF
--- a/datashader/tests/test_transfer_functions.py
+++ b/datashader/tests/test_transfer_functions.py
@@ -501,7 +501,6 @@ def test_shade_zeros(array):
 @pytest.mark.parametrize('agg', aggs)
 @pytest.mark.parametrize('attr', ['d'])
 @pytest.mark.parametrize('rescale', [False, True])
-#@pytest.mark.parametrize('rescale', [True])
 def test_shade_rescale_discrete_levels(agg, attr, rescale):
     x = getattr(agg, attr)
     cmap = ['pink', 'red']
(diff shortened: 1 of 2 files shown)


philippjfr (approved):
Looks good, thanks for the quick fix!

jbednar (approved):
Thanks!

PR #1170, opened 2023-01-24
Fix docs build

ianthomas23:
Docs build is failing installing the base environment using mamba 🤯 

Here removing mamba and making a few other tweaks.


ianthomas23:
Doc build now working again. There are a few problems in the rendered output, but I will deal with them in a separate PR.

PR #1171, opened 2023-01-25
More doc fixes

ianthomas23:
Another set of changes to remove errors and warnings in the built docs.


jbednar (commented):
Thanks! The streaming aggregation examples deserve some attention, but obviously we've never done anything with them...

PR #1172, opened 2023-01-27
User guide page for where reductions

ianthomas23:
This PR adds a page to the User Guide describing the use of `where` reductions. The page can be seen in the dev docs here: https://pyviz-dev.github.io/datashader/user_guide/Inspection_Reductions.html

It doesn't really fit in with the rest of the docs, it is evidently written by a different person. It will be easier to explain when we have support in `holoviews` for inspections using `where` as this will be most common way of using such reductions and the explanation can be interactive. But it is at least a step in the right direction.

### DIFF
--- a/doc/api.rst
+++ b/doc/api.rst
@@ -186,6 +186,7 @@ Definitions
 .. autoclass:: sum
 .. autoclass:: summary
 .. autoclass:: var
+.. autoclass:: where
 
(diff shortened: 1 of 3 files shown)


ianthomas23:
Oops, forgot to tidy my git history. Tis done now!

ianthomas23:
Merging so that it is in the upcoming release.

PR #1176, opened 2023-02-07
Do not use warnings from numpy

danigm:
numpy has removed the warnings symbol from the __init__.py file so np.warnings is not valid in newer versions of numpy. This patch justs imports the warnings module and use it directly.


ianthomas23 (approved):
Thanks @danigm

Issue #1178, opened 2023-02-07
Datashader categorical plot not rendering legend with cuDF


orozcojd:
#### ALL software version info
datashader=0.14.4
cudf=22.12.01
bokeh=2.4.3
panel=0.14.3
pandas=1.5.3
numpy=1.21.5
holoviews=1.15.4

#### Description of expected behavior and the observed behavior
After migrating Datashader code to use cudf from pandas.DataFrames, the legend for my categorical plot is no longer showing.

A working example could be found following [this documentation](conda update -n base -c conda-forge conda) and replacing pd.DataFrame with cudf.DataFrame, or refer to below snippet.

#### Complete, minimal, self-contained example code that reproduces the issue
```
df = pd.concat(dfs, ignore_index=True)
df.parameter = df.parameter.astype("category")
cuda_df = cudf.DataFrame.from_pandas(df)

curve = hv.Curve(cuda_df).opts(**options) # aggregator=ds.by("parameter", ds.count())
shader = datashade(curve, aggregator=ds.count_cat("parameter"), cnorm="log", color_key=cc.glasbey).opts(**options)

```

ianthomas23:
Thanks for the bug report, I can reproduce it. Simple reproducer:
```python
import datashader as ds
import holoviews as hv
from holoviews.operation.datashader import datashade
import pandas as pd
hv.extension('bokeh')

df = pd.DataFrame(dict(
    x    = [0.0, 1.0, 0.0, 1.0, 0.0],
    y    = [0.0, 1.0, 1.0, 0.0, 0.5],
    cat  = ['a', 'b', 'a', 'b', 'a'],
))
df.cat = df.cat.astype("category")

if 1:
    import cudf
    df = cudf.DataFrame.from_pandas(df)

curve = hv.Curve(df)
datashade(curve, aggregator=ds.by("cat"))
```
This works fine without the contents of the `if`-statement, and with it gives:
```
<snip>
File ~/.miniconda/envs/rapids/lib/python3.9/site-packages/holoviews/operation/datashader.py:1423, in shade._process(self, element, key)
   1421         return RGB(img, **params)
   1422     else:
-> 1423         img = tf.shade(array, **shade_opts)
   1424 return RGB(self.uint32_to_uint8_xr(img), **params)

File ~/github/datashader/datashader/transfer_functions/__init__.py:701, in shade(agg, cmap, color_key, how, alpha, min_alpha, span, name, color_baseline, rescale_discrete_levels)
[...]

PR #1179, opened 2023-02-10
Fix conversion from cupy in categorical rescale_discrete_levels

ianthomas23:
See issue #1178.

When using `cudf`, a categorical aggregate, `how='eq_hist'` and `rescale_discrete_levels=True` then categorical shading raises an error when trying to convert a tuple that contains a `cupy` array and scalar to a `cupy` array.

The simplest possible fix would have been in `_rescale_discrete_levels()` here:
https://github.com/holoviz/datashader/blob/d8167b41577058abe01ae1b847a5178ae6ccf172/datashader/transfer_functions/__init__.py#L230-L232
by casting `lower_span` to a `float`. This is fine if `lower_span` ia a scalar or `numpy` array, but if it is a `cupy` array then this would transfer it from GPU to CPU memory, only to be pushed back to the GPU subsequently. The solution here is to convert the span from tuple to `numpy`/`cupy` array using `hstack` in `_interpolate_alpha`.

This fixes the error produced by the OP's example code, but there is still a problem in `holoviews` not displaying the categorical legend if using the GPU. I will create a separate `holoviews` issue for this.

### DIFF
--- a/datashader/tests/test_transfer_functions.py
+++ b/datashader/tests/test_transfer_functions.py
@@ -9,7 +9,7 @@ import PIL
 import pytest
 from collections import OrderedDict
 import datashader.transfer_functions as tf
-from datashader.tests.test_pandas import assert_eq_xr
+from datashader.tests.test_pandas import assert_eq_ndarray, assert_eq_xr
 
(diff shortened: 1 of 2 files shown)


jbednar (commented):
Looks good, thanks!

Issue #1141, opened 2022-10-20
Invalid pointer when running aggregation 


Jap8nted:
I am running fairly straight forward process of passing a Dask DataFrame containing many points (lat, lon) and other dimensions to render to a canvas. Afterwards I run mean of one of those dimensions I get the error: 
```munmap_chunk(): invalid pointer
qemu: uncaught target signal 6 (Aborted) - core dumped
```
Is there anything I can check to find out the reason for this to generate an invalid pointer? it is clearly the data inside which generates this because with other datasets I get no error. 

#### ALL software version info
python 3.9.7

numba==0.56.3
pandas==1.3.5
dask==2022.5.0
numpy==1.22.0
datashader==0.14.2

```
from datashader import Canvas
from datashader.reductions import mean, sum
ddf = dask.read_csv("file.csv")
cvs = Canvas(plot_width=50, plot_height=17)
cvs.points(ddf, "LON", "LAT", agg=mean("VALUE"))
```

<img width="887" alt="image" src="https://user-images.githubusercontent.com/5432705/196892834-014e36ad-c16f-4ecd-b9e0-4ea0c9266ea9.png">

jbednar:
It would be very difficult for us to reproduce a problem like that from that description. I realize that it may be difficult for you to provide a reproducible example for a big run like that, but could you answer some questions?

- Are you running this on an M1 Mac? If so it might be an issue with the Intel processor implementation unrelated to Datashader (see e.g. https://github.com/docker/for-mac/issues/5148#issuecomment-784992663). You could try running it in native environment instead, which should also be faster.
- Are you saying that you get the error only on some datasets, but when you do get it, it's always on the `cvs.points` call? If so that warning might indicate that the entire aggregate is zero, which I'm pretty sure we have tests for, but perhaps not for this specific case using a mean aggregator. If so, maybe you could make a simple test case that would reproduce the issue.

Jap8nted:
Hi,

Sorry for the late response, I was long time away. The bug actually pops up when a dask dataframe is passed to cvs.points aggregating when the canvas has either width=0 or height=0,  I updated the versions of datashader and dask and run on an M1 Mac and a a Ubuntu amd64, the problem appeared on the amd64. Previously we got the core dump on a pod running on an EKS cluster. 


```
from datashader import Canvas
from datashader.reductions import mean, sum, max, min
from dask import dataframe as ddf
def main():
[...]

PR #1183, opened 2023-02-15
1141 | Validate canvas width, height

Jap8nted:
See issue https://github.com/holoviz/datashader/issues/1141.

When a canvas is created with plot_width=0 or plot_height=0 and the data is coming from dask, there is a core dump error, up to now, the error could just be reproduced on a machines running amd and intel. On Arm the problem does not appear. 

```
from datashader import Canvas
from datashader.reductions import mean, sum, max, min
from dask import dataframe as ddf
def main():
    df = pd.DataFrame({"LON": [0, 1, 2, 3, 4], "LAT": [0, 1, 2, 3, 4], "VALUE": [0, 1, 2, 3, 4]})
    data = ddf.from_pandas(df, npartitions=1)
    cvs = Canvas(
        plot_width=0,
        plot_height=0
    )
    cvs.points(data, "LON", "LAT", agg=max("VALUE"))
if __name__ == "__main__":
    main()
```

### DIFF
--- a/datashader/core.py
+++ b/datashader/core.py
@@ -1227,14 +1227,14 @@ x- and y-coordinate arrays must have 1 or 2 dimensions.
         self.x_axis.validate(x_range)
         self.y_axis.validate(y_range)
 
-    def validate_extend(self, width, height):
-        if width == 0 or height == 0:
-            raise ValueError("Invalid extend plot_width and plot_height must be bigger than 0")
+    def validate_size(self, width, height):
+        if width <= 0 or height <= 0:
+            raise ValueError("Invalid size: plot_width and plot_height must be bigger than 0")
        
     def validate(self):
         """Check that parameter settings are valid for this object"""
         self.validate_ranges(self.x_range, self.y_range)
-        self.validate_extend(self.plot_width, self.plot_height)
+        self.validate_size(self.plot_width, self.plot_height)
 
 
 def bypixel(source, canvas, glyph, agg, *, antialias=False):
--- a/datashader/core.py
+++ b/datashader/core.py
@@ -1229,7 +1229,7 @@ x- and y-coordinate arrays must have 1 or 2 dimensions.
 
     def validate_extend(self, width, height):
         if width == 0 or height == 0:
-            raise ValueError(f"Invalid extend plot_width and plot_height must be bigger than 0")
+            raise ValueError("Invalid extend plot_width and plot_height must be bigger than 0")
        
     def validate(self):
         """Check that parameter settings are valid for this object"""
--- a/datashader/core.py
+++ b/datashader/core.py
@@ -1227,9 +1227,14 @@ x- and y-coordinate arrays must have 1 or 2 dimensions.
         self.x_axis.validate(x_range)
         self.y_axis.validate(y_range)
 
+    def validate_extend(self, width, height):
+        if width == 0 or height == 0:
+            raise ValueError(f"Invalid extend plot_width and plot_height must be bigger than 0")
+       
     def validate(self):
         """Check that parameter settings are valid for this object"""
         self.validate_ranges(self.x_range, self.y_range)
+        self.validate_extend(self.plot_width, self.plot_height)
 
 
 def bypixel(source, canvas, glyph, agg, *, antialias=False):
--- a/datashader/tests/test_dask.py
+++ b/datashader/tests/test_dask.py
@@ -1564,3 +1564,18 @@ def test_line_antialias_where(npartitions):
             sol_where_min[j, i] = agg_where_min[j, i] = nan
 
     assert_eq_ndarray(agg_where_min.data, sol_where_min)
+
+def test_canvas_extend():
+
+    cvs_list = [
+        ds.Canvas(plot_width=0, plot_height=6),
+        ds.Canvas(plot_width=5, plot_height=0),
+        ds.Canvas(plot_width=0, plot_height=0)
+    ]
+    msg = r'Invalid extend plot_width and plot_height must be bigger than 0'
+    df = pd.DataFrame(dict(x=[0, 0.2, 1], y=[0, 0.4, 1], z=[10, 20, 30]))
+    ddf = dd.from_pandas(df, 1)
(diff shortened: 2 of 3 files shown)


jbednar:
Thanks. I'll leave it to @ianthomas23 to think about whether it's appropriate to raise an error or to return an empty array in this case; not sure which is e.g. less problematic for HoloViews.

ianthomas23:
OK, so the options here are

1. Raise an error.
2. Continue as normal and return zero-sized DataArrays.

I'd be happy with either but my preference is for 1 as I'd rather detect the situation as soon as possible and report it immediately.

@Jap8nted Are you happy to continue with this PR?

If so, this fix is good, it is failing CI because the raised `ValueError` doesn't need to use an f-string. Also, it should have a test to confirm that the fix is good. A new test in `test_pandas.py` should suffice, perhaps just below `test_line_coordinate_lengths` and it can use similar `pytest.raises` logic to that test.

Jap8nted:
Hi, 

I will continue with it, @ianthomas23, a ok, got it, will fix the f-string as well.

datashader/core.py
```diff
@@ -1227,9 +1227,14 @@ def validate_ranges(self, x_range, y_range):
         self.x_axis.validate(x_range)
         self.y_axis.validate(y_range)
 
+    def validate_extend(self, width, height):
```
ianthomas23, line 1230:
```suggestion
    def validate_size(self, width, height):
```

datashader/core.py
```diff
@@ -1227,9 +1227,14 @@ def validate_ranges(self, x_range, y_range):
...
+        if width == 0 or height == 0:
+            raise ValueError("Invalid extend plot_width and plot_height must be bigger than 0")
```
ianthomas23, line 1232:
```suggestion
            raise ValueError("Invalid size: plot_width and plot_height must be bigger than 0")
```

datashader/tests/test_dask.py
```diff
@@ -1564,3 +1564,18 @@ def test_line_antialias_where(npartitions):
             sol_where_min[j, i] = agg_where_min[j, i] = nan
 
     assert_eq_ndarray(agg_where_min.data, sol_where_min)
+
```
ianthomas23, line 1567:
```suggestion


```
I'd like to keep the whitespace consistent.

datashader/tests/test_dask.py
```diff
@@ -1564,3 +1564,18 @@ def test_line_antialias_where(npartitions):
...
+def test_canvas_extend():
```
ianthomas23, line 1568:
```suggestion
def test_canvas_size():
```

datashader/tests/test_dask.py
```diff
@@ -1564,3 +1564,18 @@ def test_line_antialias_where(npartitions):
...
+
```
ianthomas23, line 1569:
```suggestion
```

datashader/tests/test_dask.py
```diff
@@ -1564,3 +1564,18 @@ def test_line_antialias_where(npartitions):
...
+    cvs_list = [
+        ds.Canvas(plot_width=0, plot_height=6),
+        ds.Canvas(plot_width=5, plot_height=0),
+        ds.Canvas(plot_width=0, plot_height=0)
+    ]
+    msg = r'Invalid extend plot_width and plot_height must be bigger than 0'
```
ianthomas23, line 1575:
```suggestion
    msg = r'Invalid size: plot_width and plot_height must be bigger than 0'
```

datashader/tests/test_pandas.py
```diff
@@ -2393,3 +2393,20 @@ def test_line_coordinate_lengths():
     for ny in (1, 3):
         with pytest.raises(ValueError, match=msg):
             cvs.line(source=df, x=["x0", "x1"], y=np.arange(ny), axis=1)
+
```
ianthomas23, line 2396:
```suggestion
```

datashader/tests/test_pandas.py
```diff
@@ -2393,3 +2393,20 @@ def test_line_coordinate_lengths():
...
+
+
+def test_canvas_extend():
```
ianthomas23, line 2399:
```suggestion
def test_canvas_size():
```

datashader/tests/test_pandas.py
```diff
@@ -2393,3 +2393,20 @@ def test_line_coordinate_lengths():
...
+
```
ianthomas23, line 2400:
```suggestion
```

datashader/tests/test_pandas.py
```diff
@@ -2393,3 +2393,20 @@ def test_line_coordinate_lengths():
...
+    cvs_list = [
+        ds.Canvas(plot_width=0, plot_height=6),
+        ds.Canvas(plot_width=5, plot_height=0),
+        ds.Canvas(plot_width=0, plot_height=0)
+    ]
+    msg = r'Invalid extend plot_width and plot_height must be bigger than 0'
```
ianthomas23, line 2406:
```suggestion
    msg = r'Invalid size: plot_width and plot_height must be bigger than 0'
```

datashader/core.py
```diff
@@ -1227,9 +1227,14 @@ def validate_ranges(self, x_range, y_range):
         self.x_axis.validate(x_range)
         self.y_axis.validate(y_range)
 
+    def validate_extend(self, width, height):
+        if width == 0 or height == 0:
```
ianthomas23, line 1231:
```suggestion
        if width <= 0 or height <= 0:
```
Just in case someone passes in a negative number.

ianthomas23 (changes requested):
@Jap8nted This is functionally fine. I've manually tested it using `cudf` and `cudf-dask` as our CI doesn't do this for us and they both fail correctly.

I would rather call it `validate_size` than `validate_extend` so I have made a few suggested changes along those lines, and also a few changes to whitespace. If you could make those amendments it should be good to merge.

Jap8nted:
Thanks for the comments and help @ianthomas23 and happy to help!

Issue #1182, opened 2023-02-14
Dask and CUDA support for first and last reductions


ianthomas23:
`first` and `last` reductions are not implemented for dask or CUDA. This is because both work in parallel and we have no control over the order of running, hence up until now we have had no way of determining what is `first` and `last` at the point when we need to combine data from different dask partitions or different CUDA threads.

However, there is a way forward by reusing the virtual row index of the recently added `where` reduction (#1164). This allows each row of a `DataFrame` to know its row index, so if this information carries through the pipeline until the point at which we, for example, combine the results from separate dask partitions then `first` and `last` can be calculated correctly.

In pseudocode, consider that
```python
ds.first("some_column")
```
is equivalent to
```python
ds.where(ds.min(virtual_row_index), "some_column")
```
It isn't quite as simple as that as we need to use the `where` reduction the other way round from normal, i.e. using the virtual row index to determine what values of `"some_column"` to return when in fact `where` so far can only use the values of `"some_column"` to determine what virtual row index to return. But the existing machinery can be generalised for reuse here.

This will need #1177 implemented first for CUDA support.

ianthomas23:
Dask support is provided in PR #1214.

Issue #1177, opened 2023-02-07
CUDA mutexes for complicated reductions


ianthomas23:
Currently datashader only has CUDA support for relatively simple `Reduction` classes. This is primarily due to the parallel nature of CUDA hardware which means we have to support multiple GPU threads writing to the same element (pixel) of an aggregation, so we only support reductions which have a corresponding CUDA atomic operation such as `numba.cuda.atomic.max`. For more complicated reductions such as the planned `max_n` reduction we will need to implement some form of mutex ourselves, to limit one thread at a time accessing the relevant shared data. The CUDA name for this is "atomicCAS" or "atomic compare and swap".

This will need some experimentation. The initial implementation could try to limit access to a single element (pixel) of one aggregation to a single thread at a time. Benchmarking will be important here to identify a workable and efficient solution.

Useful reference: https://towardsdatascience.com/cuda-by-numba-examples-c583474124b0

PR #1184, opened 2023-02-16
first_n, last_n, max_n and min_n reductions

ianthomas23:
This is further work on improved inspection reductions (issue #1126) to add `first_n`, `last_n`, `max_n` and `min_n` reductions. Each accepts a column name and value for `n`, the number of results to return for each pixel. For example, `max("value", n=3)` will return a DataArray of shape `(ny, nx, n)` containing the 3 highest values of column `"value"` for each pixel.

Demo code using these within a `where` reduction to return corresponding row indexes or values from a different column:
```python
import datashader as ds
import pandas as pd

df = pd.DataFrame(dict(
    x     = [ 0,  0,  1,  1,  0,  0,  2,  2],
    y     = [ 0,  0,  0,  0,  1,  1,  1,  1],
    value = [ 9,  8,  7,  6,  2,  3,  4,  5],
    other = [11, 12, 13, 14, 15, 16, 17, 18],
    #index    0   1   2   3   4   5   6   7
))

canvas = ds.Canvas(plot_height=2, plot_width=3)

reductions = [
    ("where first_n index", ds.where(ds.first_n("value", 3))),
    ("where first_n other", ds.where(ds.first_n("value", 3), "other")),
    ("where max_n index", ds.where(ds.max_n("value", 3))),
    ("where max_n other", ds.where(ds.max_n("value", 3), "other")),
]

for name, reduction in reductions:
    agg = canvas.points(df, 'x', 'y', agg=reduction)
    print(name, agg.data.dtype)
    print(agg.data)
```
which outputs
```
where first_n index int64
[[[ 0  1 -1]
  [ 2  3 -1]
  [-1 -1 -1]]

 [[ 4  5 -1]
  [-1 -1 -1]
  [ 6  7 -1]]]
where first_n other float64
[[[11. 12. nan]
  [13. 14. nan]
  [nan nan nan]]

 [[15. 16. nan]
  [nan nan nan]
  [17. 18. nan]]]
where max_n index int64
[[[ 0  1 -1]
  [ 2  3 -1]
  [-1 -1 -1]]

 [[ 5  4 -1]
  [-1 -1 -1]
  [ 7  6 -1]]]
where max_n other float64
[[[11. 12. nan]
  [13. 14. nan]
  [nan nan nan]]

 [[16. 15. nan]
  [nan nan nan]
  [18. 17. nan]]]
```
where, as usual, `-1` means no row index and `nan` means no data to return.

This allows us to do some complicated combinations such as
```python
ds.summary(
    count=ds.count(),
    min_n=ds.where(ds.min_n("value", n=3)),
    max_n=ds.where(ds.max_n("value", n=3)),
)
```
to return `count` plus `min_n` and `max_n` (or `first_n` and `last_n`) in a single datashader pass.

`max_n` and `min_n` work with dask but not CUDA (issue #1177 needs to be solved for that). `first_n` and `last_n` only work on the CPU and without dask, the same as `first` and `last` (#1177 and #1182 are needed to fix that).

Using antialiased lines the results looked OK in some situation and not others, so I am raising a `NotImplemented` error for all of these when using with antialiasing and I will separately consider what is reasonable behaviour here. This includes `where(first_n)` and so on as well.

There is one issue here that needs deciding. I've called the third dimension of the DataArray returned by such a reduction `"n"` to fit in with the names `first_n`, etc. You can put multiple `whatever_n` reductions in a single `summary` reduction as shown above. If they have the same `n` then it all works out as expected. But we need a policy on labelling the third dimension if the `whatever_n` have different `n` values. We could keep the first `n` as `n`, and if subsequent `n` values are different call them `n1`, `n2`, etc?

### DIFF
--- a/datashader/compiler.py
+++ b/datashader/compiler.py
@@ -164,14 +164,21 @@ def make_append(bases, cols, calls, glyph, categorical, antialias):
         if antialias:
             args.append("aa_factor")
 
-        body.append('{0}(x, y, {1})'.format(func_name, ', '.join(args)))
-
         where_reduction = len(bases) == 1 and isinstance(bases[0], where)
+        if where_reduction:
+            arg_name = 'xxx'
+            args.append(arg_name)
+
+        call  = '{0}(x, y, {1})'.format(func_name, ', '.join(args))
+
         if where_reduction:
             # where reduction needs access to the return of the contained
             # reduction, which is the preceding one here.
-            body[-2] = 'if ' + body[-2] + ' >= 0:'
-            body[-1] = '    ' + body[-1]
+            body[-1] = f'{arg_name} = {body[-1]}'
+            body.append(f'if {arg_name} >= 0:')
+            call = f'    {call}'
+
+        body.append(call)
 
     body = ['{0} = {1}[y, x]'.format(name, arg_lk[agg])
             for agg, name in local_lk.items()] + body
--- a/datashader/compiler.py
+++ b/datashader/compiler.py
@@ -166,17 +166,16 @@ def make_append(bases, cols, calls, glyph, categorical, antialias):
 
         where_reduction = len(bases) == 1 and isinstance(bases[0], where)
         if where_reduction:
-            arg_name = 'xxx'
-            args.append(arg_name)
+            update_index_arg_name = next(names)
+            args.append(update_index_arg_name)
 
-        call  = '{0}(x, y, {1})'.format(func_name, ', '.join(args))
-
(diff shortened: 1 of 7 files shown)


ianthomas23:
After discussion, we've decided to allow multiple `*_n` reductions only if they all have the same `n` value. This allows us to keep the new coordinate label as `n`.

jbednar (approved):
Looks good. Just need user docs, which can be a separate PR.

PR #1185, opened 2023-02-20
Correctly handle RaggedArray conversions to numpy arrays

### LINKED ISSUE #1158: Support numpy 1.24
Prior to `numpy` 1.24 creating an array from ragged nested sequences produced a `VisibleDeprecationWarning`. With 1.24 this is now a `ValueError`. This is OK currently as `numba` doesn't yet support `numpy` 1.24 but it needs to be fixed here before that happens, so it is quite urgent.

Thanks  to @Hoxbro for identifying this (https://github.com/holoviz/geoviews/pull/608).

ianthomas23:
Fixes #1158.

This removes all warnings caused by `numpy` conversions of ragged arrays which will be errors in `numpy` 1.24. In fact there weren't any problems in the library code itself as if you follow the docstrings you will create ragged arrays correctly, but some of the tests used shortcuts instead of the recommended way and these have been changed in this PR.

Either of these are correct ways to create a `DataFrame` series that is a ragged array to use in `datashader`:

```python
import pandas as pd
x = pd.array([[0, 1, 2], [4, 5, 6, 7, 8, 9]], dtype='Ragged[float32]')

from datashader.datatypes import RaggedArray
x = RaggedArray([[0, 1, 2], [4, 5, 6, 7, 8, 9]], dtype='float32')
```
The `dtype` is optional for `RaggedArray` as it is inferred.

The following worked in the past but are incorrect using `numpy` 1.24 onwards:
```python
x = np.asarray([[0, 1, 2], [4, 5, 6, 7, 8, 9]])

x = np.asarray([[0, 1, 2], [4, 5, 6, 7, 8, 9]], dtype=object)
```
The first approach will immediately fail, telling you to use the second `dtype=object` approach. This works for some but not all codepaths in `datashader` as it drops important dtype information. Hence avoid both.

Eventually the `RaggedArray` pandas extension array within `datashader` will be replaced by `awkward-array` and will simplify our code and make it more robust to future changes.

### DIFF
--- a/datashader/datatypes.py
+++ b/datashader/datatypes.py
@@ -638,6 +638,13 @@ Invalid indices for take with allow_fill True: {inds}""".format(
 
         return np.array([v for v in self], dtype=dtype, copy=copy)
 
+    def tolist(self):
+        # Based on pandas ExtensionArray.tolist
+        if self.ndim > 1:
+            return [item.tolist() for item in self]
+        else:
+            return list(self)
+
     def __array__(self, dtype=None):
         dtype = np.dtype(object) if dtype is None else np.dtype(dtype)
         return np.asarray(self.tolist(), dtype=dtype)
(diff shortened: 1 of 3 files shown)


jbednar (approved):
Looks good, once tests pass!

ianthomas23:
After rebasing this is passing CI.

PR #1187, opened 2023-02-23
Fix python 3.7 CI

ianthomas23:
Python 3.7 windows CI is failing as `mamba` cannot resolve the dependencies. A number of dependent packages have stopped support for 3.7. Here I am removing that particular CI run from the `pyctdev`/`holoviz_tasks` `conda`-based run and replaced them with a `pip` equivalent which is much easier to control.


ianthomas23:
Closing as superseded by #1188.

PR #1195, opened 2023-03-21
Switch to pyproject.toml and hatch

ianthomas23:
This is a very early stage proof of concept removing `pyctdev` and replacing it with use of `pyproject.toml` and `hatch` instead, modelled on holoviz/panel#4541.


ianthomas23:
This successfully builds and runs the test suite in github actions using `pip` for all 3 major OSes, and on Ubuntu and macOS using `conda`, but fails on Windows using `conda`.

PR #1196, opened 2023-03-22
Support max_n and min_n reductions on GPU

### LINKED ISSUE #1177: CUDA mutexes for complicated reductions

ianthomas23:
Closes #1177.

This adds support for `max_n` and `min_n` reductions on a GPU, both with and without `dask`. The key change is to add new CUDA mutex functionality to support CUDA `append` functions (i.e. individual pixel callbacks) that do more than a simple get/set operation. Because of the massively parallel nature of CUDA hardware multiple threads can access the same `canvas` pixel at the same time, and up until now we have been restricted to CUDA atomic operations (https://numba.readthedocs.io/en/stable/cuda/intrinsics.html#supported-atomic-operations) in `append` functions. With the new mutex we can lock access to a particular pixel to a single thread at a time and thus perform more complicated operations such as for `max_n` without any race conditions.

In implementation we need to get the mutex (a `cupy` array) to the CUDA `append` functions and this is achieved within the `expand_aggs_and_cols` framework by appending the mutex array in the `make_info` function which is where other arrays and/or dataframe columns are extracted and passed to `append` functions. This ensures that there is only ever a single shared mutex even if multiple reductions need it.

This implementation is limited by what is currently available in `numba` 0.56 which means we can only lock/unlock the mutex as a whole rather than individual elements/pixels of it so the performance will not be great. Numba PR https://github.com/numba/numba/pull/8790 will allow us to lock individual pixels, so when `numba` 0.57 is released I will write another PR to check use the fast route if that is available otherwise drop back to this slower one.

There is no support yet for `where(max_n)` on CUDA, but this will follow in another PR soon.

### DIFF
--- a/datashader/compiler.py
+++ b/datashader/compiler.py
@@ -146,7 +146,7 @@ def make_append(bases, cols, calls, glyph, categorical, antialias):
     if any_uses_cuda_mutex:
         # This adds an argument to the append() function that is the cuda mutex
         # generated in make_info.
-        inputs += [True]  # Doesn't matter what is in the list.
+        inputs += ["_cuda_mutex"]
     signature = [next(names) for i in inputs]
     arg_lk = dict(zip(inputs, signature))
     local_lk = {}
@@ -179,7 +179,7 @@ def make_append(bases, cols, calls, glyph, categorical, antialias):
             args.append("aa_factor")
 
         if uses_cuda_mutex:
-            args.append(signature[-1])
+            args.append(arg_lk["_cuda_mutex"])
 
         where_reduction = len(bases) == 1 and isinstance(bases[0], where)
         if where_reduction:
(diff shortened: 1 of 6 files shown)


jbednar:
Nice. Thanks!

PR #1197, opened 2023-03-31
pyproject.toml with pip and conda

ianthomas23:
This is a proof of concept of switching to `pyproject.toml` and using low-level commands such as `pip`, `conda` and `pytest`.


ianthomas23:
Closing this as I think there will be a `hatch` or `PDM` based solution instead.

PR #1198, opened 2023-04-05
x and y range attributes on returned aggregations

### LINKED ISSUE #1157: Expose computed bounds after aggregation
Currently the data bounds will computed during aggregation if the x or y ranges are not provided explicitely:
```
canvas = ds.Canvas(plot_width=300, plot_height=300, x_axis_type='linear', y_axis_type='linear')
aggregation = canvas.points(df, 'x', 'y', agg=ds.count())
```
However those computed ranges do not seem to get exposed and are thus not really accessible. Nevertheless they are quite important when displaying the data. Thus I propose to utilize the `attrs` field of the xarray.DataArray (which can hold arbitrary metadata) to provide the (computed) ranges.
Essentially I want to get the following:

```
x_range = aggregation.attrs["x_range"]
y_range = aggregation.attrs["y_range"]
```

ianthomas23:
Closes #1157.

This PR adds new attributes `x_range` and `y_range` to aggregations returned from datashader. Simple example:

```python
import datashader as ds
import pandas as pd

df = pd.DataFrame(dict(x=[1.1, 2.2, 3.3], y=[4.4, 5.5, 6.6]))
canvas = ds.Canvas(plot_height=5, plot_width=5)
agg = canvas.points(source=df, x="x", y="y")
print(agg)
```

Output produced:
```
<xarray.DataArray (y: 5, x: 5)>
array([[1, 0, 0, 0, 0],
       [0, 0, 0, 0, 0],
       [0, 0, 1, 0, 0],
       [0, 0, 0, 0, 0],
       [0, 0, 0, 0, 1]], dtype=uint32)
Coordinates:
  * x        (x) float64 1.32 1.76 2.2 2.64 3.08
  * y        (y) float64 4.62 5.06 5.5 5.94 6.38
Attributes:
    x_range:  (1.1, 3.3)
    y_range:  (4.4, 6.6)
```

so the attributes can be accessed using `agg.x_range` and similar.

The ranges are set regardless of whether they are specified by the user in the `Canvas` constructor, or determined from the data limits.

For situations that return an `xarray.Dataset` rather than an `xarray.DataArray`, e.g. if a `ds.summary()` is used, the attributes are copied to the `Dataset`. Hence they are always available as attributes of the top-level object returned from `Canvas` aggregation functions.

### DIFF
--- a/datashader/compiler.py
+++ b/datashader/compiler.py
@@ -265,7 +265,13 @@ def make_finalize(bases, agg, schema, cuda):
         def finalize(bases, cuda=False, **kwargs):
             data = {key: finalizer(get(inds, bases), cuda, **kwargs)
                     for (key, finalizer, inds) in calls}
-            return xr.Dataset(data)
+
+            # Copy x and y range attrs from any DataArray (their ranges are all the same)
+            # to set on parent Dataset
+            name = agg.keys[0]  # Name of first DataArray.
+            attrs = {attr: data[name].attrs[attr] for attr in ('x_range', 'y_range')}
+
+            return xr.Dataset(data, attrs=attrs)
         return finalize
     else:
         return agg._build_finalize(schema)
--- a/datashader/core.py
+++ b/datashader/core.py
@@ -1200,7 +1200,7 @@ x- and y-coordinate arrays must have 1 or 2 dimensions.
 
         coords = {xdim: xs, ydim: ys}
         dims = [ydim, xdim]
-        attrs = dict(res=res[0])
+        attrs = dict(res=res[0], x_range=self.x_range, y_range=self.y_range)
 
         # Find nodata value if available in any of the common conventional locations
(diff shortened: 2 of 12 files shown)


ianthomas23:
Test failures are because our CI conda environments are now using `pandas 2.0.0` which is incompatible with recent `xarray`, so the version of `xarray` installed is `0.19.0` from July 2021. `xarray` will fix this in due course, in the mean time I will try pinning `pandas < 2` to test this PR.

jbednar:
Looks like the ranges are slightly larger than the coordinates would suggest. Is that just due to the size of each array cell? If so, having the ranges explicitly listed is definitely useful.

ianthomas23:
Yes, linear coordinates are equally spaced and the end coordinates are half a cell width inside the ends. The ranges are therefore easy to calculate from the coordinates. But with logarithmic axes the maths is non-trivial so it is useful for the ranges to always be available.

PR #1203, opened 2023-04-25
Drop support for python 3.7

ianthomas23:
See proposed HEP1 at https://github.com/holoviz/holoviz/pull/362


ianthomas23:
I am self-merging this, otherwise I cannot progress with numba 0.57 RC which supports python 3.8+.

PR #1205, opened 2023-04-27
Do not pass masked arrays to numba

ianthomas23:
This is a WIP to support numba 0.57. We can no longer pass masked arrays to numba, but that is OK as we handle the masks separately anyway.

There is a separate problem that occurs on macOS but not linux, that I am still investigating.


ianthomas23:
There is a warning
```
  <snip>/site-packages/numba/np/ufunc/dufunc.py:84: NumbaDeprecationWarning: The 'nopython' keyword argument was not supplied to the 'numba.jit' decorator. The implicit default value for this argument is currently False, but it will be changed to True in Numba 0.59.0. See https://numba.readthedocs.io/en/stable/reference/deprecation.html#deprecation-of-object-mode-fall-back-behaviour-when-using-jit for details.
    dispatcher = jit(_target='npyufunc',
```
which can be ignored as it was fixed in https://github.com/numba/numba/pull/8909 and the fix will be included in the `numba 0.57` release.

ianthomas23:
I am happy with this now, all tests are passing with new `numba 0.57`, `numpy 1.24` and `python 3.11` as well as previous versions of all 3. The CI failures are on codecov which is the usual inability to determine coverage of `numba` code.

The only change that could be removed is downloading packages from the `numba` conda channel; I will remove this in a separate PR when `numba 0.57` is available on `conda-forge`.

PR #1212, opened 2023-05-10
Use fast cuda mutex available in numba 0.57

### LINKED ISSUE #1211: Implement fast CUDA mutex
Now that numba 0.57 has been released with support for CUDA atomic compare and swap functionality (numba/numba#8790), we need to use this to give us faster CUDA reductions such as `max_n`, building on top of #1196. It will use a check on the version of numba available to choose whether to use the new or old mutex functionality.

ianthomas23:
Fixes #1211.

For `numba >= 0.57` there is a CUDA mutex that is lockable on a per-pixel basis whereas earlier `numba` doesn't have the required functionality so there is a single lock shared between all pixels. Complicated CUDA reductions such as `max_n` need the mutex so that they work atomically; the new mutex allows the CUDA code to work in parallel very efficiently whereas the old solution forces datashader to run mostly serially.

There is very little code changed here. The `numba.__version__` is used to select the appropriate mutex lock and unlock functions, and the size of the `cupy` array used for the mutex. Other than that there is a new argument to the `info` function as the canvas shape is needed. Although this argument is only needed for CUDA code running a complicated reduction such as `max_n`, I have chosen to pass it for every call rather than dynamically determine at the point of calling whether it is needed or not.

As an illustration of performance, demo code:
```python
import cudf, datashader as ds, numba as nb, numpy as np, pandas as pd

n = 1_000_000
rng = np.random.default_rng(95123)
df = pd.DataFrame(dict(
    x = rng.random(size=n, dtype=np.float32),
    y = rng.random(size=n, dtype=np.float32),
    value = rng.random(size=n, dtype=np.float32),
))
df = cudf.DataFrame.from_pandas(df)

canvas = ds.Canvas(plot_height=300, plot_width=600)
agg = canvas.points(source=df, x="x", y="y", agg=ds.max_n("value", 3))
```

and time the `canvas.points` call, discarding the first timing as this includes `numba` compilation. This gives the following timings (using `numba 0.56.3` and `numba 0.57` for the `slow` and `fast` mutex):

Screenshot 2023-05-10 at 13 46 06

GPU on test system is an Nvidia Quadro T1000.

### DIFF
--- a/datashader/compiler.py
+++ b/datashader/compiler.py
@@ -32,7 +32,7 @@ def compile_components(agg, schema, glyph, *, antialias=False, cuda=False):
         Takes the aggregate shape, and returns a tuple of initialized numpy
         arrays.
 
-    ``info(df)``
+    ``info(df, canvas_shape)``
         Takes a dataframe, and returns preprocessed 1D numpy arrays of the
         needed columns.
 
@@ -45,7 +45,7 @@ def compile_components(agg, schema, glyph, *, antialias=False, cuda=False):
         Combine a list of base tuples into a single base tuple. This forms the
         reducing step in a reduction tree.
 
-    ``finalize(aggs)``
+    ``finalize(aggs, cuda)``
         Given a tuple of base numpy arrays, returns the finalized ``DataArray``
         or ``Dataset``.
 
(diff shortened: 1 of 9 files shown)


jbednar:
Excellent, thanks!

Issue #1207, opened 2023-05-02
Dask and CUDA support for first_n and last_n reductions


ianthomas23:
`first_n` and `last_n` reductions currently only work on the CPU without dask. They also need to work using CUDA and dask.

See https://github.com/holoviz/datashader/issues/1126.

ianthomas23:
Dask support is provided in PR #1214.

ianthomas23:
Fixed by #1219.

PR #1214, opened 2023-05-11
Dask support for first, last, first_n and last_n reductions

ianthomas23:
This fixes the Dask parts of issues #1182 and #1207. CUDA support will follow in a separate PR, this is already too big.

Full list of example reductions this supports on CPU with and without Dask:
- `first('value')` - returns single `value` per pixel
- `last('value')` - ditto
- `first_n('value', n=3)` - returns `n=3` `value`s per pixel
- `last_n('value', n=3)` - ditto
- `where(first('value'))` - returns single row index per pixel
- `where(last('value'))` - ditto
- `where(first_n('value', n=3))` - returns `n=3` row indexes per pixel
- `where(last_n('value', n=3))` - ditto
- `where(first('value'), 'other')` - returns single `other` per pixel
- `where(last('value'), 'other')` - ditto
- `where(first_n('value', n=3), 'other')` - returns `n=3` `other`s per pixel
- `where(last_n('value', n=3), 'other')` - ditto

Summary of changes:

1. Some `reduction` behaviour is now dependent on whether supplied data is partitioned (dask or CUDA) or not, so extra arguments `cuda` and `partitioned` are passed through the class hierarchy.

2. I have factored out some common `first` and `last` code into a private base class `_first_or_last` to simplify, and similar applies to `_first_n_or_last_n`.

3. The fundamental addition is that `first('value')` using Dask is implemented as `where(_min_row_index(), 'value')`. This takes advantage of the existing `where` machinery and a new class `_min_row_index` which is very similar to a `min` reduction except that it accepts a virtual row index column, not a real column supplied by the user. I could have extended the `min` class to cover this, but I thought it better to keep the new functionality in a new private class. It is possible to instantiate the private class of course, although no user code should do this, and I have done this in the test suite. There are equivalent new private classes for `last`, `first_n` and `last_n`.

I do need to recheck that trying to use CUDA with these reductions raises informative error messages not crashes.

### DIFF
--- a/datashader/reductions.py
+++ b/datashader/reductions.py
@@ -1866,7 +1866,8 @@ class _max_or_min_row_index(OptionalFieldReduction):
         # self.column is None but row index is passed as field argument to append functions
         # Doesn't yet support antialiasing
         if cuda:
-            return self._append_cuda
+            raise ValueError(f"'{type(self).__name__}' reduction is not supported on the GPU")
+            #return self._append_cuda
         else:
             return self._append
 
@@ -1948,7 +1949,8 @@ class _max_n_or_min_n_row_index(FloatingNReduction):
         # self.column is None but row index is passed as field argument to append functions
         # Doesn't yet support antialiasing
         if cuda:
-            return self._append_cuda
+            raise ValueError(f"'{type(self).__name__}' reduction is not supported on the GPU")
+            #return self._append_cuda
         else:
             return self._append
 


jbednar (approved):
Whew! That didn't look fun, but I'm very glad to see it. I didn't spot any issues, but I may or may not have glazed over when reviewing certain bits. :-)

ianthomas23:
> I may or may not have glazed over when reviewing certain bits. :-)

It hadn't occurred to me to be entertaining. I can try that approach, but it may come at the expense of code accuracy :grin:

ianthomas23:
I've rebased it on `main` to pick up the fast CUDA mutex PR, and added a couple of error messages. There are no functional code changes so I will merge then CI is green.

PR #1218, opened 2023-05-19
Fix numpy 1.24 warnings

### LINKED ISSUE #1216: RuntimeWarning: invalid value encountered in cast
Using `numpy 1.24` we are seeing
```
RuntimeWarning: invalid value encountered in cast
```
when running tests. For example
```
$ pytest -v datashader/tests/test_pandas.py::test_categorical_count_binning
datashader/tests/test_pandas.py::test_categorical_count_binning[df0] PASSED                                                                                    [100%]

datashader/tests/test_pandas.py::test_categorical_count_binning[df0]
  /Users/iant/github/datashader/datashader/reductions.py:202: RuntimeWarning: invalid value encountered in cast
    index = ((values - self.bin0) / self.binsize).astype(int)
```
The same warning also occurs in `transfer_functions`:
```
datashader/tests/test_transfer_functions.py: 306 warnings
  /home/runner/work/datashader/datashader/datashader/transfer_functions/__init__.py:310: RuntimeWarning: invalid value encountered in cast
    r = interp(data, span, rspan, left=255).astype(np.uint8)
```

Thanks to @Hoxbro for identifying this downstream.

ianthomas23:
Fixes #1216.

Summary of fixes:

- Conversion of float array containing NaN to integer:
    -  Simple uses now have a call to `np.nan_to_number` to convert `np.nan` to `0`, which is the number that has always been returned in this scenario.
    - The more complicated code in `category_binning` already determines a mask of the NaN values, so those NaN values are set to 0 (they could be any valid finite number) before the conversion to int.
    - In `Canvas.raster` we avoid the warning by using a `fill_value` of `0` for integer `dtype`s as that is what the docstring says is used in this scenario.

- Warnings about finding max or min of an "all-nan slice" are specifically ignored as they already return the expected value  of `nan` in this situation.

### DIFF
--- a/datashader/tests/test_raster.py
+++ b/datashader/tests/test_raster.py
@@ -11,6 +11,10 @@ try:
 except ImportError:
     rioxarray = None
 
+from dask.context import config
+
+config.set(scheduler='synchronous')
+
 open_rasterio_available = pytest.mark.skipif(rioxarray is None and rasterio is None, reason="requires rioxarray or rasterio")
 
 from os import path
--- a/datashader/core.py
+++ b/datashader/core.py
@@ -1070,6 +1070,8 @@ x- and y-coordinate arrays must have 1 or 2 dimensions.
             mask = array==nan_value
             array = np.ma.masked_array(array, mask=mask, fill_value=nan_value)
             fill_value = nan_value
+        elif np.issubdtype(source.dtype, np.integer):
+            fill_value = 0
         else:
             fill_value = np.NaN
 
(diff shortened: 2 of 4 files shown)


datashader/tests/test_raster.py
```diff
@@ -11,6 +11,10 @@
 except ImportError:
     rioxarray = None
 
+from dask.context import config
+
+config.set(scheduler='synchronous')
```
ianthomas23, line 16: This is consistent with what is in `test_dask.py`.

PR #1222, opened 2023-05-26
Make datashader.composite imports lazy

ianthomas23:
This PR moves the `import` of various `datashader.composite` functions into the functions that use them rather than at the top of the `transfer_functions` module. It makes the initial import of datashader much faster.

Thanks to @Hoxbro for identifying this downstream.


jbednar (approved):
Cool!

Issue #1210, opened 2023-05-04
Categorical support for where and <whatever>_n reductions


ianthomas23:
The extent to which this works already needs to be investigated, and whatever is missing needs to be implemented.

See #1126.

PR #1229, opened 2023-06-06
Support by(max_n) and by(min_n)

ianthomas23:
Support for categorical `max_n` and `min_n` reductions such as `ds.by("cat", ds.max_n("value", n=3))` on CPU and GPU both with and without dask.  This is the first part of issue #1210, support for categorical `first_n`, `last_n` and `where` to follow.

Example:
```python
import datashader as ds
import numpy as np
from numpy import nan
import pandas as pd

x = np.arange(2)
df = pd.DataFrame(dict(
    y_from = [0.0, 1.0, 0.0, 1.0, 0.0],
    y_to   = [0.0, 1.0, 1.0, 0.0, 0.5],
    value  = [1.1, 3.3, 5.5, 2.2, 4.4],
    cat    = ['a', 'b', 'a', 'b', 'a'],
))
df["cat"] = df["cat"].astype("category")

canvas = ds.Canvas(plot_height=2, plot_width=3)
agg = canvas.line(source=df, x=x, y=["y_from", "y_to"], axis=1,
                  agg=ds.by("cat", ds.max_n("value", n=3)))
print(agg)
```
which prints
```
xarray.DataArray (y: 2, x: 3, cat: 2, n: 3)>
array([[[[5.5, 4.4, 1.1],
         [nan, nan, nan]],

        [[1.1, nan, nan],
         [2.2, nan, nan]],

        [[1.1, nan, nan],
         [2.2, nan, nan]]],


       [[[nan, nan, nan],
         [3.3, 2.2, nan]],

        [[5.5, 4.4, nan],
         [3.3, nan, nan]],

        [[5.5, 4.4, nan],
         [3.3, nan, nan]]]])
Coordinates:
  * x        (x) float64 0.1667 0.5 0.8333
  * y        (y) float64 0.25 0.75
  * cat      (cat) <U1 'a' 'b'
  * n        (n) int64 0 1 2
Attributes:
    x_range:  (0, 1)
    y_range:  (0.0, 1.0)
```
Note that the returned DataArray has shape `(ny, nx, ncat, n)` which I think is more logical than the alternative possibility of `(ny, nx, n, ncat)`.

In terms of implementation, functions like `nanmax_n_in_place` now always accept a 4D array so that there is a single implementation for 3D (`max`) and 4D (`max_n`) arrays for each of CPU and GPU. Use of the combine function in `max` inserts the extra dimension of size 1 to change the shape without copying any data.

### DIFF
--- a/datashader/tests/test_pandas.py
+++ b/datashader/tests/test_pandas.py
@@ -25,11 +25,19 @@ df_pd = pd.DataFrame({'x': np.array(([0.] * 10 + [1] * 10)),
                       'plusminus': np.arange(20, dtype='f8')*([1, -1]*10),
                       'empty_bin': np.array([0.] * 15 + [np.nan] * 5),
                       'cat': ['a']*5 + ['b']*5 + ['c']*5 + ['d']*5,
+                      'cat2': ['a', 'b', 'c', 'd']*5,
                       'cat_int': np.array([10]*5 + [11]*5 + [12]*5 + [13]*5)})
 df_pd.cat = df_pd.cat.astype('category')
+df_pd.cat2 = df_pd.cat2.astype('category')
 df_pd.at[2,'f32'] = nan
 df_pd.at[2,'f64'] = nan
 df_pd.at[2,'plusminus'] = nan
+# x          0  0   0  0 0   0 0  0 0  0   1   1  1   1  1    1  1   1  1   1
+# y          0  0   0  0 0   1 1  1 1  1   0   0  0   0  0    1  1   1  1   1
+# i32        0  1   2  3 4   5 6  7 8  9  10  11 12  13 14   15 16  17 18  19
+# f32        0  1 nan  3 4   5 6  7 8  9  10  11 12  13 14   15 16  17 18  19
+# plusminus  0 -1 nan -3 4  -5 6 -7 8 -9  10 -11 12 -13 14  -15 16 -17 18 -19
+# cat2       a  b   c  d a   b c  d a  b   c   d  a   b  c    d  a   b  c   d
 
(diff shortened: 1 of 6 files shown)


datashader/tests/test_pandas.py
```diff
@@ -25,11 +25,19 @@
...
                       'empty_bin': np.array([0.] * 15 + [np.nan] * 5),
                       'cat': ['a']*5 + ['b']*5 + ['c']*5 + ['d']*5,
+                      'cat2': ['a', 'b', 'c', 'd']*5,
                       'cat_int': np.array([10]*5 + [11]*5 + [12]*5 + [13]*5)})
 df_pd.cat = df_pd.cat.astype('category')
+df_pd.cat2 = df_pd.cat2.astype('category')
 df_pd.at[2,'f32'] = nan
 df_pd.at[2,'f64'] = nan
 df_pd.at[2,'plusminus'] = nan
+# x          0  0   0  0 0   0 0  0 0  0   1   1  1   1  1    1  1   1  1   1
```
ianthomas23, line 35: I find myself creating this manual table whenever I need to check new tests, so here including it to make it easier to check new tests in future.

datashader/tests/test_pandas.py
```diff
@@ -652,7 +708,7 @@ def test_categorical_sum_binning(df):
 
 
 @pytest.mark.parametrize('df', dfs)
-def test_categorical_max(df):
+def test_categorical_max2(df):
```
ianthomas23, line 711: Keeping the existing categorical max test but renaming it so it doesn't overwrite the new one which is above it in this file.

jbednar (approved):
Thanks!

jbednar:
We'll need some docs at the Datashader level when you're done with all this, of course.

PR #1237, opened 2023-06-21
Add support for categorical where reductions

### LINKED ISSUE #1210: Categorical support for where and <whatever>_n reductions

ianthomas23:
Fixes #1210.

This adds support for categorical `where` reductions on CPU and GPU, with and without Dask.

An example is
```python
canvas = ds.Canvas(ny, nx)
agg = canvas.points(... agg=ds.by("cat", ds.where(ds.max_n("mass", n=3))))
```
This returns a 4D `xarray.DataArray` of shape `(ny, nx, ncat, n)` containing for each pixel and category the indexes of the 3 rows in the supplied `DataFrame` that have the maximum values of the `"mass"` column.

To return the values from another column instead of row indexes this would be
```python
agg = canvas.points(... agg=ds.by("cat", ds.where(ds.max_n("mass", n=3), "other")))
```

We can replace `max_n` in this example with `max`, `min`, `first`, `last`, `min_n`, `first_n`, or `last_n`. 

Support is also added for
```python
ds.by("cat", ds.first("value"))
```
and the `last`, `first_n` and `last_n` equivalents as these are implemented using `where` under certain circumstances (GPU and/or Dask).

### DIFF
--- a/datashader/reductions.py
+++ b/datashader/reductions.py
@@ -11,6 +11,7 @@ import xarray as xr
 from datashader.enums import AntialiasCombination
 from datashader.utils import isminus1, isnull
 from numba import cuda as nb_cuda
+from numba.typed import List
 
 try:
     from datashader.transfer_functions._cuda_utils import (
@@ -322,6 +323,11 @@ class Reduction(Expr):
     @property
     def inputs(self):
         return (extract(self.column),)
+
+    def is_where(self):
+        """Return ``True`` if this is a ``where`` reduction or directly wraps
+        a where reduction."""
+        return False
 
     def _antialias_requires_2_stages(self):
         # Return True if this Reduction must be processed with 2 stages,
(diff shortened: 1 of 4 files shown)


datashader/reductions.py
```diff
@@ -1793,32 +1818,34 @@ def _build_combine(self, dshape, antialias, cuda, partitioned):
         invalid = isminus1 if self.selector.uses_row_index(cuda, partitioned) else isnull
 
         @ngjit
-        def combine_cpu_2d(aggs, selector_aggs):
-            ny, nx = aggs[0].shape
+        def combine_cpu(aggs, selector_aggs):
```
ianthomas23, line 1821: There is a lot of similar but not quite identical code here that I am planning to refactor in a separate PR.

jbednar (approved):
Looks great, thanks! Is this the end of it? I.e., are there any combinations of by/where/reductions with cpu/gpu/dask that are still unsupported? Or is that entire cross product now covered somewhere?

ianthomas23:
After rebase tests are failing with some bokeh-panel incompatibility when running examples. That is nothing to do with this PR, so merging this and will deal with example problem separately.

PR #1249, opened 2023-07-10
Update readme to include python 3.11

thuydotm:
Added Python 3.11 in `Installation` section as datashader has officially supported it since version 0.15.0


ianthomas23 (approved):
Thanks @thuydotm!

PR #1259, opened 2023-07-28
Antialiased min and max row index reductions

ianthomas23:
This PR adds antialiased support for `_min_row_index` and `_max_row_index` reductions, including within `by` (categorical) reductions. It is part of the ongoing work to provide antialiased support for all reductions.


jbednar (approved):
Looks good, thanks!

PR #1276, opened 2023-09-08
Support pandas 2.1

ianthomas23:
Now that there is a SpatialPandas RC the supports Pandas 2.1 (https://pypi.org/project/spatialpandas/0.4.9rc1/#files and https://anaconda.org/pyviz/spatialpandas/files), we can remove the temporary pin of #1275.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -96,6 +96,9 @@ jobs:
       - name: doit develop_install
         run: |
           doit ecosystem=pip develop_install -o tests -o examples
+      - name: install spatialpandas RC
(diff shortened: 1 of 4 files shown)


ianthomas23:
It wasn't as easy as just removing the pandas pin. There were also problems in the edge bundling code when reading rows from a Pandas DataFrame; previously the code took whatever column order it was given which is not reliable when converting column names to deal with `edge_id`. Now we are using the recommended `df.itertuples()` and specifically extracting the columns in a particular order.

All of this code looks ripe for performance improvement eventually.

PR #1281, opened 2023-09-26
Update to latest holoviz_tasks

Hoxbro:
The changes in the install task are caching when a test fails. Beforehand, it will only do so if the test passed. 

The other change is saving the environment files from the different runs, making debugging easier. 

You can get the environment files here: https://github.com/holoviz/datashader/actions/runs/6309708107?pr=1281 under artifacts.


ianthomas23 (approved):
Thanks, this looks really useful.

Issue #1278, opened 2023-09-15
Support Canvas.lines of 2D xarray with common x coordinates


ianthomas23:
I would like to support the `xarray` equivalent of `LinesAxis1XConstant` which has multiple lines that all share the same `x` coordinates. The `xarray.DataArray` would be 2D, one dimension would be the `x` dimension and the other could be anything including a string/categorical dimension. It is best illustrated by an example:

```python
import datashader as ds
import numpy as np
import xarray as xr

xr_ds = xr.Dataset(
    data_vars=dict(
        name=(("channel", "x"), [[2, 1, 0, 1, 2], [1, 1, 1, 1, 1]]),
    ),
    coords=dict(
        channel=("channel", ['a', 'b']),
        x=("x", np.arange(5)),
    ),
)

canvas = ds.Canvas(plot_height=3, plot_width=5)
agg = canvas.line(source=xr_ds, x="x", y="name", agg=ds.count())
```

The `x` coordinates here have length 5 and the `DataArray` is of shape `(2, 5)`. The output aggregation here would be something like
```
<xarray.DataArray (y: 3, x: 5)>
array([[0, 0, 1, 0, 0],
       [1, 2, 1, 2, 1],
       [1, 0, 0, 0, 1]], dtype=uint32)
```
This should support both dask and CUDA backed xarrays, and the 2D dimensions could be in either order as we do not want to force the user to do an unnecessary array transpose.

We do support some use of `xarray` for lines at the moment, but it is restricted to 1D `x` and `y` values.

jbednar:
Sounds good to me!

ianthomas23:
Closed by #1282.

PR #1282, opened 2023-09-27
Implement lines using 2D xarray with common x coordinates

ianthomas23:
This is a work in progress to implement datashading lines from a 2D xarray that has common `x` coordinates, as described in issue #1278.

So far it only works for `count` and `any` reductions, with and without Dask but only on the CPU not GPU. Further work is underway to support other simple reductions, categorical  and `where` reductions, antialiasing, CUDA, etc.

There is a limitation when using dask chunking of the 2D array in the `x` direction in that the line segments between chunks are not rendered. This is the same limitation that occurs in other datashader line glyphs that similarly chunk along arrays, and no attempt will be made to address the limitation in this PR.

This work is part of a CZI Round 5 grant.

### DIFF
--- a/datashader/compiler.py
+++ b/datashader/compiler.py
@@ -140,7 +140,7 @@ def compile_components(agg, schema, glyph, *, antialias=False, cuda=False, parti
 
     create = make_create(bases, dshapes, cuda)
     append, any_uses_cuda_mutex = make_append(bases, cols, calls, glyph, antialias)
-    info = make_info(cols, any_uses_cuda_mutex)
+    info = make_info(cols, cuda, any_uses_cuda_mutex)
     combine = make_combine(bases, dshapes, temps, combine_temps, antialias, cuda, partitioned)
     finalize = make_finalize(bases, agg, schema, cuda, partitioned)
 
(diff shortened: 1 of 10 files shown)


ianthomas23:
Here is an example using `nchannel` lines of random data, each line offset from its neighbours.
```python
import colorcet as cc
import datashader as ds
import numpy as np
import xarray as xr

nx = 100
x = np.linspace(0.0, 1.0, nx)

nchannel = 26
channel = [chr(i) for i in range(65, 65+nchannel)]

rng = np.random.default_rng(8321)
data = rng.normal(loc=np.arange(nchannel).reshape((-1, 1)), scale=0.5, size=(nchannel, nx))

xr_ds = xr.Dataset(
    data_vars=dict(
        some_name=(("channel", "x"), data),
        value=("channel", np.arange(nchannel))
    ),
    coords=dict(
        channel=("channel", channel),
        x=("x", x),
    ),
)
print(xr_ds)

canvas = ds.Canvas(plot_height=600, plot_width=800)
agg = canvas.line(source=xr_ds, x="x", y="some_name", agg=ds.first("value"))
im = ds.transfer_functions.shade(agg, how="linear", cmap=cc.glasbey_cool)
```
which produces
```
<xarray.Dataset>
Dimensions:    (channel: 26, x: 100)
Coordinates:
  * channel    (channel) <U1 'A' 'B' 'C' 'D' 'E' 'F' ... 'U' 'V' 'W' 'X' 'Y' 'Z'
  * x          (x) float64 0.0 0.0101 0.0202 0.0303 ... 0.9697 0.9798 0.9899 1.0
Data variables:
    some_name  (channel, x) float64 0.5883 0.06726 -0.8486 ... 24.96 26.22 25.42
    value      (channel) int64 0 1 2 3 4 5 6 7 8 ... 17 18 19 20 21 22 23 24 25
```
and this image
lines_xarray_example

ianthomas23:
This now works for xarray DataArrays that are backed by

- numpy arrays
- dask arrays
- cupy arrays

but not for dask-cupy arrays.

ianthomas23:
Also note that the 2D `xr.DataArray` can have coordinates either way round, i.e. the common `x` coords can be dimension 0 or 1.

PR #1284, opened 2023-10-02
Vendor datashape

### LINKED ISSUE #1003: Remove dependency on `datashape`
[`datashape`](https://github.com/blaze/datashape) is unmaintained and hasn't had a release since the end of 2016. At this point it would probably be better to vendor the necessary parts.

ianthomas23:
Fixes #1003.

Datashape does not import on Python 3.12. The project is unmaintained, there hasn't been a release on PyPI since 2016 and the latest git commit was 4 years ago. This PR vendors the source code within Datashader and removes the install dependency.

The Datashape test suite is also included and will be automatically run in CI. It takes less than 3 seconds to run on my dev machine so it will not noticeably affect CI run times.


ianthomas23:
Passes CI and successfully builds packages and docs.

PR #1285, opened 2023-10-05
Support GeoPandas Polygons and MultiPolygons

### LINKED ISSUE #1006: Support GeoPandas GeoDataFrames
Context:

* Datashader requires plain arrays of coordinates ("ragged arrays" in case of lines/polygons, with additional offset/indices arrays) to efficiently visualize geometries (and this is also the representation that SpatialPandas uses under the hood).
* GeoPandas stores geometries as "opaque" Shapely objects (wrapping GEOS C++ object), and converting this to an array of coordinates is currently not always very efficient (although it's certainly possible, it's also what spatialpandas does in the `from_geopandas` conversion code eg [here](https://github.com/holoviz/spatialpandas/blob/4d679f97d750c6a69b49143c1a21e17bf6731d5d/spatialpandas/geometry/polygon.py#L37-L53))

With the latest release of PyGEOS, the conversion from geometries to (ragged) coordinate arrays can be done much more efficiently, though. 

<details>
<summary>Function using PyGEOS to convert array of GEOS geometries to arrays of coordinates / offsets (+ putting those in a spatialpandas array)</summary>

```python
def get_flat_coords_offset_arrays(arr):
    """
    Version for MultiPolygon data
    """
    # explode/flatten the MultiPolygons
    arr_flat, part_indices = pygeos.get_parts(arr, return_index=True)
    # the offsets into the multipolygon parts
    offsets1 = np.insert(np.bincount(part_indices).cumsum(), 0, 0)

    # explode/flatten the Polygons into Rings
    arr_flat2, ring_indices = pygeos.geometry.get_rings(arr_flat, return_index=True)
    # the offsets into the exterior/interior rings of the multipolygon parts 
    offsets2 = np.insert(np.bincount(ring_indices).cumsum(), 0, 0)

    # the coords and offsets into the coordinates of the rings
    coords, indices = pygeos.get_coordinates(arr_flat2, return_index=True)
    offsets3 = np.insert(np.bincount(indices).cumsum(), 0, 0)
    
    return coords, offsets1, offsets2, offsets3

def spatialpandas_from_pygeos(arr):
    """
    Create the actual spatialpandas MultiPolygonArray by putting the individual arrays
    into a pyarrow ListArray
    """
    coords, offsets1, offsets2, offsets3 = get_flat_coords_offset_arrays(arr)
    coords_flat = coords.ravel()
    offsets3 *= 2
    
    # create a pyarrow array from this
    _parr3 = pa.ListArray.from_arrays(pa.array(offsets3), pa.array(coords_flat))
    _parr2 = pa.ListArray.from_arrays(pa.array(offsets2), _parr3)
    parr = pa.ListArray.from_arrays(pa.array(offsets1), _parr2)
    
    return spatialpandas.geometry.MultiPolygonArray(parr)
```

</details>

With such a faster conversion available, it becomes more interesting for Datashader to directly support `geopandas.GeoDataFrame`, instead of requiring an up-front conversion to `spatialpandas.GeoDataFrame`. 
Currently, the spatialpandas requirement is hardcoded here (for `polygons()`):

https://github.com/holoviz/datashader/blob/1ae52b65ec8a79920e5db9c6c04487f254428553/datashader/core.py#L694-L701

Adding support for GeoPandas can be done, using the function I defined above, with something like (leaving aside imports of geopandas/pygeos):

```patch
    from spatialpandas import GeoDataFrame
    from spatialpandas.dask import DaskGeoDataFrame
    if isinstance(source, DaskGeoDataFrame):
        # Downselect partitions to those that may contain polygons in viewport
        x_range = self.x_range if self.x_range is not None else (None, None)
        y_range = self.y_range if self.y_range is not None else (None, None)
        source = source.cx_partitions[slice(*x_range), slice(*y_range)]
+   elif isinstance(source, geopandas.GeoDataFrame):
+      # Downselect actual rows to those for which the polygon is in viewport
+      x_range = self.x_range if self.x_range is not None else (None, None)
+      y_range = self.y_range if self.y_range is not None else (None, None)
+      source = source.cx[slice(*x_range), slice(*y_range)]
+      # Convert the subset to ragged array format of spatialpandas
+      geometries = spatialpandas_from_pygeos(source.geometry.array.data)
+      source = pd.DataFrame(source)
+      source["geometry"] = geometries
    elif not isinstance(source, GeoDataFrame):
        raise ValueError(
            "source must be an instance of spatialpandas.GeoDataFrame or \n"
```

This patch is what I tried in the following notebook, first using a smaller countries/provinces dataset from NaturalEarth, and then with a larger NYC building footprints dataset (similar to https://examples.pyviz.org/nyc_buildings/nyc_buildings.html). 

Notebook: https://nbviewer.jupyter.org/gist/jorisvandenbossche/3e7ce14cb5118daa0f6097d686981c9f

Some observations:

- This actually works nicely!
- Initial rendering with datashader is a bit slower when directly using geopandas.GeoDataFrame because of the extra conversion step. But, the conversion takes less time than the actual rendering, so it's only a relatively small slowdown.
- Zooming into small areas is really fast with the large dataset. And actually faster as using spatialpandas.GeoDataFrame (because I added a `.cx` spatial subsetting step in my patch above, filtering the data before rendering). For spatialpandas, such subsetting is only added for the dask version.

Gif of the notebook in action (the buildings dataset is fully loaded in memory, and not pararellized with dask, unlike the PyViz gallery example), interactively zooming into a GeoPandas dataframe with Datashader and Holoviews:

Peek 2021-06-08 13-45

(note this was done a bit manually with Holoviews DynamicMap and a callback with Datashader code, because the integrated datashade functionality of Holoviews/HvPlot wouldn't preserve the geopandas.GeoDataFrame with the current versions)

---

So, what's the way forward here? I think I showed that it can be useful for Datashader to directly support GeoPandas, and that it can also be done with a relatively small change to datashader. 
The big question, though, is about the **"GEOS -> ragged coordinate arrays -> spatialpandas array" conversion**. Where should this live / how should DataShader and GeoPandas interact?

Some initial thoughts about this:

- The quickest way to get this working is to do this conversion in DataShader (the above functions only rely on pygeos (which the user will already have when using GeoPandas) and pyarrow/spatialpandas (which are already requirements for this part of datashader)). But, long term, is this code that Datashader wants to maintain? Or is there a more logical place for this code?
- It could also live in SpatialPandas, since they already have code for such conversion of GeoPandas <-> SpatialPandas (and it would optimize its current implementation of that). But, should a user need to have SpatialPandas to plot GeoPandas with Datashader? (see also last bullet point)
- Alternatively, GeoPandas could add a function or method that converts its geometries into this required format, and then Datashader can call that method to get the data it needs. Long term, this might be the better solution (since other projects interacting with geopandas might also want to get the geometries in this format).
- How to communicate this data? In the current POC version included above, I first get the raw coordinate and offset arrays, and then convert them into a `pyarrow.ListArray` to then convert it to a spatialpandas `MultiPolygonArray`. But in the end, what Datashader needs is only the raw coordinates and offsets arrays.  
  For example, for rendering polygons, you access `.buffer_values` and `.buffer_offsets` of the MultiPolygonArray, which gives back the raw coordinate and offset arrays. 
  So in theory, this roundtrip through pyarrow and spatialpandas is not needed, and some method could convert GeoPandas geometries into coordinate/offset arrays, which could be directly handled by datashader as is. This would however require a bit more changes in datashader in the way that data gets passed down from `Canvas.polygons()` into the `glyph` rendering (as currently that uses the spatiapandas array as container for the coordinates/offsets).

One possible idea (relating to the third bullet point) is to standardize on some kind of `__geo_arrow_arrays__` interface (returning the coordinate + offset arrays), similarly to the existing `__geo_interface__` that returns the geometries in GeoJSON-like dictionary (and which can be used now for accepting any "geometry-like" object even from libraries you don't know).

ianthomas23:
This is a WIP to add support for rendering Polygons and MultiPolygons from GeoPandas GeoDataFrames. Fixes #1006.

The approach taken is to use `.cx` to obtain the subset of geometries within the rendered bounds and then use `shapely.io.to_ragged_array` to obtain contiguous arrays of coordinates and integer offsets. From here it is very similar to the existing SpatialPandas rendering code which uses numba to quickly traverse the geometries. There is no support for dask yet, nor any tests.

Obligatory pretty picture:
time_ragged_gpd

I have included three example notebooks:

1. `natural_earth.ipynb`. This uses "naturalearth.land" from geodatasets which is directly read using GeoPandas and converted to SpatialPandas. Outputs are the same for both and rendering times (after the first slower render that compiles the numba code) are about the same on my dev machine (M1 mac) at about 0.26 seconds for 600x1200 pixels.
2. `nyc_buildings.ipynb`. Uses the NYC buildings dataset which is used in SpatialPandas examples and has been pre-prepared for such. Download location and code to convert to GeoPandas parquet file is included at the top of the notebook. It demonstrates that the rendered outputs are the same. Time to render the whole dataset at 1000x1000 pixels is 1.6 s for GeoPandas and 0.7 s for SpatialPandas.
3. `nyc_buildings_zoom.ipynb`. The same as the previous notebook with a `zoom_factor` to render a zoomed area centred on the middle of the dataset. For a `zoom_factor` of 30 GeoPandas is faster at 27 ms compared to SpatialPandas at 39 ms.

Here is a table of the render times as a function of `zoom_level` for my dev machine.

| zoom_level  | GeoPandas | SpatialPandas |
| --- | --- | --- |
| 1 | 1.6 s | 0.7 s |
| 3 | 0.45 s | 0.23 ms |
| 10 | 65 ms | 61 ms |
| 30 | 27 ms | 39 ms |
| 100 | 16 ms | 30 ms |

So for this size of dataset we render GeoPandas slower than SpatialPandas at full resolution, but it scales better with zoom factor so that it is faster at high zoom levels.

I think this is explained by the two different approaches (SpatialPandas I understand, GeoPandas I don't yet). SpatialPandas reads the whole dataset from file as ragged arrays as that is how it is stored. This is fast. To render a subset of the data the ragged arrays are kept as they are and a boolean array is used to identify which polygons are within the bounds and need to be rendered. The rendering loop iterates over all of the polygons and uses the boolean flags to not bother rendering polygons that are not needed.

For GeoPandas and the shapely to_ragged_array conversion, the spatial subset is calculated before the ragged arrays are obtained, and the ragged arrays are dynamically generated. At full resolution the `to_ragged_array` takes over a second for me. When it comes to rendering there is no need for any boolean flags, we only have the polygons within the bounds and we render each and every one of them.

Pinging @jorisvandenbossche and @martinfleis to see if I have taken a poor approach here and to check if my analysis is correct. I did wonder if the format of saving the GeoPandas file is important here as I just used the default `to_parquet` options, whereas the SpatialPandas file is highly optimised for this use case.


datashader/glyphs/polygon.py
```diff
@@ -201,6 +242,62 @@ def draw_polygon(
     return draw_polygon
 
 
+def _build_extend_geopandas_polygon_geometry(
+        draw_polygon, expand_aggs_and_cols
+):
+    def extend_cpu(
+            sx, tx, sy, ty, xmin, xmax, ymin, ymax, geometry, *aggs_and_cols
+    ):
+        ragged = shapely.io.to_ragged_array(geometry)
```
jorisvandenbossche, line 251:
```suggestion
        ragged = shapely.to_ragged_array(geometry)
```

datashader/glyphs/points.py
```diff
@@ -72,11 +77,23 @@ def required_columns(self):
         return [self.geometry]
 
     def compute_x_bounds(self, df):
-        bounds = df[self.geometry].array.total_bounds_x
+        col = df[self.geometry]
+        if isinstance(col.dtype, gpd_GeometryDtype):
+            # geopandas
+            bounds = col.total_bounds[::2]
```
jorisvandenbossche, line 83: One small optimization would be to calculate the total_bounds only once (and not twice for x bounds and y bounds separately), if that's doable in the current code organisation

jorisvandenbossche:
Cool! 
I can't say I understand everything exactly, but generally the code that interacts with geopandas looks good. I think your analysis of the differences sounds correct (at least it also matches with what I wrote in the issue, and indeed using `cx` here ensures we first filter the geometries before converting to ragged arrays). 

> I did wonder if the format of saving the GeoPandas file is important here as I just used the default to_parquet options, whereas the SpatialPandas file is highly optimised for this use case.

For the timings it shouldn't matter, because once the data is read into memory, the representation in GeoPandas itself is the same independent from the format of the parquet file. 
For a full workflow of reading + visualizing, it can of course matter. But right now the defaults are fine, I think (that will store the geometries as WKB). With a project like https://github.com/geoarrow/geoarrow-python we can also read parquet file directly into geoarrow format (i.e. the ragged arrays). But for the initial GeoPandas support here, standard GeoParquet is fine.

martinfleis:
Thanks for looking into this!

I have tried timing the performance of `cx` versus `sindex.query` resulting in the same dataframe and it seems that at least with the NY buildings dataset you used, there's a space for improvements.

```py
from spatialpandas import io
import shapely

ny = io.parquet.read_parquet("/Users/martin/Downloads/nyc_buildings.parq").to_geopandas()

x = -8230000
y = 4960000
offset = 10

%%timeit
ny.cx[x-offset:x+offset, y-offset:y+offset]

%%timeit
ny.iloc[ny.sindex.query(shapely.box(x-offset,y-offset, x+offset, y+offset))]
```

The results:

| offset | cx | sindex | n_hits |
|--------|----|--------|---------|
| 10 | 29 ms ± 905 µs| 349 µs ± 41.6 µs | 3 |
| 50 | 28.1 ms ± 372 µs | 312 µs ± 9.32 µs | 29 |
| 250 | 28.8 ms ± 561 µs | 346 µs ± 2.37 µs | 572 |
| 1000 | 29.8 ms ± 714 µs | 997 µs ± 41.3 µs | 7740 |
| 5000 | 46.3 ms ± 130 µs | 11.6 ms ± 238 µs | 121858 |
| 10000 | 85.5 ms ± 1 ms | 37.4 ms ± 1.07 ms | 356570 |
| 50000 | 161 ms ± 3.61 ms | 120 ms ± 1.71 ms | 1157859 (all of them) |

All the way to the extent covering the whole array, `sindex` solution seems to outperform `cx`. I am not sure how generalisable this is as there were some experiments around this in geopandas (https://github.com/geopandas/geopandas/pull/1949, https://github.com/geopandas/geopandas/issues/2671) that did not prove this to be always a better solution but development on the GEOS side in the meantime could have changed the situation. It is surely worth exploring (also on the geopandas side, if we should not use it in `cx` internally after all).

jorisvandenbossche:
Interesting! For datashader, I would say that's something we should decide on the geopandas side, and then datashader can just keep using `.cx`? (I am not sure whether datashader would be able to make any smarter decision on what to use compared to geopandas)

jorisvandenbossche:
@martinfleis to be fully equivalent, you need a `predicate="intersects"` in the sindex.query call, and checking my previous benchmarks locally, that's actually making a significant difference for the case when you are selecting a large part of the data. At the same time, that's also a small difference that datashader might not care about, and so potentially a reason for here to use sindex.query instead of cx.

martinfleis:
True, there can be some minor differences as sindex just checks bounding boxes in this case. Though those are irrelevant for the usage in Datashader. And in the benchmark above, all the dfs are of equal shape.

datashader/glyphs/polygon.py
ianthomas23: Thanks.

datashader/glyphs/points.py
ianthomas23: Yes, I can do this.

ianthomas23:
Thanks for both of your comments. I'll deal with Joris' suggestions and continue with `dask-geopandas` support which I suspect should be relatively straightforward.

Overall this seems really promising and the code additions are simple enough to be really low risk. So I'd be inclined to finish this off as soon as possible and merge it, with `geopandas` and `dask-geopandas` being optional dependencies. Then people can use it and we can work on improving the performance over time.

Beyond that it makes sense to also provide direct support for other `geopandas` geometry types (lines, points, etc) as they will be similar (and simpler) than this polygon support. Outside of Datashader there is probably work to be done in other HoloViz projects (HoloViews, GeoViews, hvPlot) but I'll need to speak to other HoloViz maintainers about this.

jbednar:
This all sounds great! Thanks, everyone.

datashader/glyphs/polygon.py
ianthomas23: Done

datashader/glyphs/points.py
ianthomas23: Done

ianthomas23:
This is all working now, giving exactly the same results as SpatialPandas using either GeoPandas or Dask-GeoPandas. I have moved the imports from the top of `polygon.py` to where it is used (thanks to @maximlt and @Hoxbro for this recommendation).

I have kept the use of `.cx` for the spatial filtering rather than `sindex.query` but I am happy to be led by GeoPandas maintainers in this going forward.

There is no change to the project install dependencies, but I have added a new set of optional dependencies so that users will be able to `pip install datashader[geopandas]` to get both geopandas and dask-geopandas installed at the same time. I have explicitly pinned `shapely >= 2.0.0` so that `.to_ragged_array` is available and also added a runtime check just before we call it, in case an older version is installed.

I haven't yet added a new page to the docs about this. I was thinking that the best approach would be to add the Point and Line support and then write a new docs page about them all together.

There will indeed need to be some changes in HoloViews to support GeoPandas directly, as at the moment any attempt to pass a GeoPandas.GeoDataFrame from HoloViews to Datashader will force a conversion to SpatialPandas. Probably the best timing for this change is when the Point and Line support has been added to Datashader.

jorisvandenbossche:
> I have kept the use of `.cx` for the spatial filtering rather than `sindex.query` but I am happy to be led by GeoPandas maintainers in this going forward.

The upside of `.cx` is that it works for both geopandas and dask-geopandas at the same time. But if you want the fastest possible option for geopandas.GeoDataFrame, the `.sindex.query` approach will typically be a bit faster.

ianthomas23:
I have switched to using `.sindex.query` for geopandas and kept `.cx` for dask-geopandas. It is more complicated, partly because you cannot pass `None` for unspecified bounds but it seems that `-np.inf` and `+np.inf` are OK.

ianthomas23:
Merging as is. Support for lines and points to follow shortly.

PR #1290, opened 2023-10-13
Fix float16 being a floating type.

J08nY:
This basically applies the following: https://github.com/blaze/datashape/pull/241 to the vendored datashape.
Fixes an issue where float16 is not recognized as a floating type (and as a result as a real type, which makes it not work in plots).


ianthomas23 (approved):
Thanks @J08nY.

Issue #1274, opened 2023-08-17
Fix codecov reporting


ianthomas23:
The codecov report often claims a reduction in coverage, even for PRs that only touch the docs. As a result, PRs and the project README often report tests as failing when they haven't really. This could be easily improved.

ianthomas23:
Closed by #1292.

PR #1292, opened 2023-10-18
Update codecov configuration

ianthomas23:
Update codecov configuration so that reporting is more informative, and allow a slight reduction in coverage to still be classified as passing, otherwise PRs that update CUDA code will always be reported as failing coverage changes as CUDA code is not exercised by our Github Actions.


ianthomas23:
Closes #1274.

ianthomas23:
This improves the situation so merging.

PR #1295, opened 2023-10-20
Add pre-commit

ianthomas23:
This adds `pre-commit` to the repo and CI, following a similar approach to that in HoloViews. The initial setup is relatively simple, and we can extend it over time. It includes some of the standard pre-commit hooks as well as `codespell` and `ruff`.

It uses the default set of `ruff` checks, and I have fixed all reported errors. These are in the following categories:

1. Line length. I have set the max line length to 100 so that there aren't too many changes, and I expect in time the various HoloViz repos will converge on a common value.
2. Multiple statements on one line.
3. No bare `except`.
4. Ambiguous variable name `l`.
5. Module level imports not at top of file.
6. Use `def` rather than assigning a `lambda`.
7. Use `a not in b` rather than `not a in b`.

None of these changes should be controversial, but I accept that this PR is rather large so it may not be acceptable as it is.


jbednar (approved):
The general types of changes are fine. I would go blind if I tried to review each line and make sure that it hasn't changed, so I'll have to rely on your assurance that you have done so (or you can pick a victim to help review that for you. :-)). Thanks!

PR #1297, opened 2023-10-25
Geopandas example

ianthomas23:
This adds a new page to the User Guide in the docs to show use of GeoPandas. It uses one of the `geodatasets` and shows three images, one of which is (population density per US county in 1990):

Screenshot 2023-10-25 at 15 16 44

It is minimalistic at this stage as it uses pure Datashader. When we have support in HoloViews for this, we can add an interactive section to the end and hence show interactive zooming and panning, and show colorbars as well, to make it more interesting.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -23,6 +23,10 @@ repos:
     - id: codespell
       additional_dependencies:
         - tomli
+  - repo: https://github.com/hoxbro/clean_notebook
+    rev: v0.1.13
+    hooks:
+      - id: clean-notebook
 
 ci:
   autofix_prs: false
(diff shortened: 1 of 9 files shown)


jbednar (approved):
Looks good! Can you include a sentence or two in the new notebook about the tradeoffs between GeoPandas and SpatialPandas? Presumably something about one being faster, but the other being more convenient if you are already have your data in GeoPandas?

jbednar:
Looks good, thanks! I forget -- do we have antialiasing supported for polygons? If so it would be good to add that to the example.

ianthomas23:
> Looks good, thanks! I forget -- do we have antialiasing supported for polygons?

No, there is only antialiasing support for lines.

jbednar:
Oops; I wasn't clear. I meant, antialiasing for polygon _outlines_.  Those are just polylines, so I'd think they should be supported?

ianthomas23:
> Oops; I wasn't clear. I meant, antialiasing for polygon _outlines_. Those are just polylines, so I'd think they should be supported?

Yes, any use of `Canvas.line()` supports antialiasing regardless of the type of the `source`. Here it is with `line_width=2`:

Screenshot 2023-10-26 at 08 56 09

PR #1300, opened 2023-10-27
Improved antialiased mean reduction

ianthomas23:
This improves the output of `Canvas.line` using an antialiased `mean` reduction.

Example code:
```python
import datashader as ds
import numpy as np
import pandas as pd

x = np.asarray([0, 0.5, 1])
df = pd.DataFrame(dict(
    y0    = [0.0, 0.5, 1.0],
    y1    = [1.0, 0.0, 0.5],
    y2    = [0.0, 1.0, 0.2],
    value = [1.1, 2.2, 3.3],
))

canvas = ds.Canvas(plot_height=100, plot_width=150, x_range=(-0.1, 1.1), y_range=(-0.1, 1.1))
agg = canvas.line(source=df, x=x, y=["y0", "y1", "y2"], line_width=10, agg=ds.mean("value"), axis=1)
ds.transfer_functions.shade(agg, how="linear")
```
Image produced, before and after this PR:
temp_old temp_new

and zoomed in by a factor of 3 (before and after):
Screenshot 2023-10-27 at 15 52 42 Screenshot 2023-10-27 at 15 52 57

Before this PR a `mean` reduction was implemented by dividing a `sum` reduction by a `count` reduction. If you imagine `sum("value")` where all the value are 1 then the `sum` and `count` produce identical results with larger values at the middle of lines and smaller antialiased values at the edges. Dividing one by the other therefore loses the antialiasing.

This PR changes the `mean` to be a `sum` divided by a new reduction `_count_ignore_antialiasing` and the latter, as the name suggests, ignores the antialiasing factor. Hence the result produced is antialiased.

It isn't actually quite that simple as at the edges of joins artifacts were produced where the line goes over itself. Pixels where this occurred are visited twice, the second time with a larger antialiasing factor. Now when this occurs both of the antialiasing factors are passed to `Reduction.append` functions so that the reduction can better deal with it. There is no change to the low-level antialiasing code which has always calculated both values, but now both values are used rather than just one.

### DIFF
--- a/datashader/compiler.py
+++ b/datashader/compiler.py
@@ -392,7 +392,7 @@ def make_append(bases, cols, calls, glyph, antialias):
 
         args.extend([local_lk[i] for i in temps])
         if antialias:
-            args.append("aa_factor")
+            args += ["aa_factor", "prev_aa_factor"]
 
         if local_cuda_mutex and prev_local_cuda_mutex:
             # Avoid unnecessary mutex unlock and lock cycle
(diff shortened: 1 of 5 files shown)


jbednar (approved):
Looks great; thanks for following up on this!

PR #1306, opened 2023-11-06
Numpy 2.0 support

Hoxbro:
Found with the latest release of ruff and this command `ruff check . --preview --select=NPY201 --fix`


datashader/datashape/coretypes.py
```diff
@@ -779,7 +779,7 @@ def from_numpy_dtype(self, dt):
         elif np.issubdtype(dt, np.timedelta64):
             unit, _ = np.datetime_data(dt)
             return TimeDelta(unit=unit)
-        elif np.issubdtype(dt, np.unicode_):
+        elif np.__version__[0] != "2" and np.issubdtype(dt, np.unicode_):
```
Hoxbro, line 782:
```suggestion
        elif np.__version__[0] < "2" and np.issubdtype(dt, np.unicode_):
```

datashader/datashape/coretypes.py
```diff
@@ -779,7 +779,7 @@ def from_numpy_dtype(self, dt):
         elif np.issubdtype(dt, np.timedelta64):
             unit, _ = np.datetime_data(dt)
             return TimeDelta(unit=unit)
-        elif np.issubdtype(dt, np.unicode_):
+        elif np.__version__[0] < "2" and np.issubdtype(dt, np.unicode_):
```
philippjfr, line 782: Does numpy2 not have nay unicode dtypes?

Hoxbro: No, the recommendation is to use `np.str_` instead, but this is checked further down, which is why I did it this way.

https://numpy.org/devdocs/numpy_2_0_migration_guide.html

philippjfr (approved):

PR #1314, opened 2024-01-31
Fix test suite

Hoxbro:
This fixes the failing tests and removes warnings from the test CI. 

Also fixes a problem with Windows not solving the conda environment. 

Before this PR: `13 failed, 1581 passed, 62 skipped, 10 xfailed, 13569 warnings`
After this PR: `1581 passed, 65 skipped, 11 xfailed`

PR #1316, opened 2024-02-06
Remove redundant py2 helper code

alexander-beedie:
Some long-unnecessary py2 → py3 `int` and `str` helper stubs were lurking in the codebase. 

These can be removed to clean things up a bit (with the minor/bonus side-effect of making all the `isinstance` checks that referenced them ~1.5-2x faster, as the helper stubs were both single-element tuples).


datashader/datashape/predicates.py
```diff
@@ -28,7 +28,7 @@ def isscalar(ds):
     """
     if isinstance(ds, str):
         ds = dshape(ds)
-    if isinstance(ds, DataShape) and len(ds) == 1:
+    elif isinstance(ds, DataShape) and len(ds) == 1:
```
Hoxbro: Are you sure you can do this? For example, If the calculation above makes ds a DataShape with length 1.

datashader/datashape/coretypes.py
```diff
@@ -376,10 +372,10 @@ class String(Unit):
...
             fixlen, encoding = None, None
-        if len(args) == 1:
-            if isinstance(args[0], _strtypes):
+        elif len(args) == 1:
+            if isinstance(args[0], str):
                 fixlen, encoding = None, args[0]
-            if isinstance(args[0], _inttypes):
+            if isinstance(args[0], int):
                 fixlen, encoding = args[0], None
         if len(args) == 2:
```
Hoxbro, line 380:
```suggestion
        elif len(args) == 2:
```

To align with changes above.

Hoxbro (commented):
Thank you!

We vendored datashape this fall as it failed to install with Python 3.12, as it hadn't been updated for some time... That does not mean we should have Python 2 code in it. 

I haven't set over the test because https://github.com/holoviz/datashader/pull/1314 is needed for them to pass. I will set them over when the PR is merged.

datashader/datashape/predicates.py
Hoxbro: You reverted this before I could press the review button, and then it disappeared for me.

alexander-beedie: Yes, I realised I had been a little over-enthusiastic once I got some of the tests working ;)

alexander-beedie:
> We vendored datashape this fall as it failed to install with Python 3.12, as it hadn't been updated for some time... That does not mean we should have Python 2 code in it.

Makes sense! I was just starting to very gently poke around to see what it might take to get first-class Polars integration 🤔

datashader/datashape/predicates.py
Hoxbro: I am very familiar with that feeling.

Hoxbro:
Thank you for the PR :+1:

PR #1326, opened 2024-05-01
Add download data script

### LINKED ISSUE #1325: Polygon User Guide Long CSV Download Output
Under the "[Realistic Example](https://datashader.org/user_guide/Polygons.html#realistic-example)" in the Polygon User Guide, the CSV download is taking up the majority of the page.

image

It continues like this for about 80% of the entire page.

hoxbro:
Fixes #1325


Hoxbro:
I have built the documents again from this PR. Which fixes the problem in the issue.

hoxbro:
need to update to work with bokeh_sampledata

setup.py
```diff
@@ -40,6 +40,11 @@
     'spatialpandas',
 ]
 
+if sys.python_version_info[:2] >= (3, 10):
```
hoxbro, line 43:
```suggestion
if sys.version_info[:2] >= (3, 10):
```

philippjfr (approved):

philippjfr:
What's the longer term plan? Should the `pyct` commands in `datashader/__init__.py` be removed?

hoxbro:
> What's the longer term plan?

This only collects all the download data procedures in one place, it is heavily focused on the developer experience. `pyct` is still used.

I think longer term `pyct` should be replaced by `hvsampledata`.

Issue #1327, opened 2024-05-01
Visual artifacts when shading Polygons


philipc2:
I have a `spatialpandas.GeoDataFrame` with ~600k Polygons that I am attempting to render. There are no holes or gaps between any of the polygons. 

<img width="530" alt="image" src="https://github.com/holoviz/datashader/assets/67855069/19476238-4d6f-4ad3-a70a-4a30ac0919fc">

```Python
import datashader as ds
import datashader.transfer_functions as tf

cvs = ds.Canvas(plot_width=300, plot_height=300)
agg = cvs.polygons(gdf, geometry='geometry', agg=ds.mean('temperature_500hPa'))
tf.shade(agg)
```

<img width="293" alt="image" src="https://github.com/holoviz/datashader/assets/67855069/9eaa577e-75d2-4b63-befe-34f5d0a554e6">

The same issue continues when using HoloViews (which is what I started with initially. However, using different `pixel_ratio` values seems to change the behavior of these artifacts. 

```Python
hv_polygons = hv.Polygons(gdf, vdims=['temperature_500hPa'])
(hv.operation.datashader.rasterize(hv_polygons, dynamic=False, pixel_ratio=1.0) + 
 hv.operation.datashader.rasterize(hv_polygons, dynamic=False, pixel_ratio=2.0) + 
 hv.operation.datashader.rasterize(hv_polygons, dynamic=False, pixel_ratio=4.0) + 
 hv.operation.datashader.rasterize(hv_polygons, dynamic=False, pixel_ratio=8.0)).cols(2)
```

<img width="578" alt="image" src="https://github.com/holoviz/datashader/assets/67855069/1d23ac87-bc7d-4cee-bc43-da9fee28e436">

I have tried the following to attempt to isolate the issues, but it did not change anything
* Using `geopandas` instead of `spatialpandas`
* Using `matplotlib` as a backend

Version Info
* datashader                0.16.1
* holoviews                 1.18.3
* numba                     0.59.1
* spatialpandas             0.4.10

philipc2:
I can share the data upon request. Is there a preferred format for sharing a `GeoDataFrame` with Polygons. I assume a CSV should be good?

jbednar:
[...]

PR #1329, opened 2024-05-08
Remove artifact from Polygon rendering

hoxbro:
Fixes https://github.com/holoviz/datashader/issues/1327

image

image

Lets see how the CI handles this change.

Edit: Can still see some artifacts, if I use the holoviews example in the original issue. Can make them disappear, if I completely remove this if statement.


jbednar:
Thanks for taking a stab at this! It does seem to address the zoomed-out issue:

image

But zooming in, seems like it's not actually drawing the polygons any more?

image

jbednar:
For reference, here's what it did before the change, where the zoomed-out version has artifacts, but the zoomed-in version is fine:

image

Hoxbro:
image

datashader/glyphs/polygon.py
```diff
@@ -197,8 +200,10 @@ def draw_polygon(
                     # Reject edges that are above, below, or left of current pixel.
                     # Note: Edge skipped if lower vertex overlaps,
                     #       but is kept if upper vertex overlaps
-                    if (y0c >= yi or y1c < yi
-                            or (x0c < xi and x1c < xi)
+                    if (
+                        y0c >= yi
+                        or y1c < yi
+                        or (x0c < xi and x1c < xi)
```
jbednar, line 206: Can this reformatting be reverted?

jbednar:
Looks great! Panning and zooming with HoloViews, I can no longer detect any issues now for that dataset.  Thanks for following up on that @Hoxbro !  Can you motivate what the meaning of each of the changes is, and any risks?  

To me it seems like one risk would be if the polygon coordinates are near the `atol` value for `np.isclose`, which is 1e-08 by default.  E.g. if we were plotting nanometer-scale shapes on an integrated circuit wafer, would we have problems from the shape coordinates being considered "close"? Or are all such comparisons being done in pixel space, where 1e-08 is indeed a very small tolerance?

Hoxbro:
I have removed `np.isclose`, which I played around with when trying to find the problem. The coordinates is in pixel space, but I saw no difference with it. In theory, it could catch some of the cross-products, which are close to zero, but I don't feel it is worth the extra calculations for a case that right now is only theoretical. 

For risk related to the removal of  `+1` I can't say as this is pretty deep in the machinery combined with some theory I haven't dove deep into.

jbednar (approved):
Thanks for this. In the end, it's a very simple change, and seems to address the underlying issue (starting to draw one pixel too late).  I've studied the test cases that changed and I _think_ they make sense. The extent that's being drawn has changed slightly, but that seems to be what's required to fix the bug.  So as far as I can see, this deals with the reported bug. Thanks again!

.github/workflows/test.yaml
```diff
@@ -66,16 +66,26 @@ jobs:
...
           MATRIX=$(jq -nsc '{
               "os": ["ubuntu-latest", "macos-latest", "windows-latest"],
-              "python-version": ["3.9", "3.10", "3.11", "3.12"]
+              "python-version": ["3.9", "3.10", "3.11", "3.12"],
+              "exclude": [{
+                  "os": "windows-latest",
+                  "python-version": "3.9"
+              }]
           }')
+          # Windows won't solve on Windows + Python 3.9
```
hoxbro, line 75:
```suggestion
          # Won't solve on Windows + Python 3.9
```

.github/workflows/test.yaml
```diff
@@ -66,16 +66,26 @@ jobs:
...
```
hoxbro, line 88:
```suggestion
          # Won't solve on Windows + Python 3.9
```

PR #1338, opened 2024-05-22
Parallelize the test suite and fix a test polluted bug

hoxbro:
<s>I see this flaky test: `datashader/tests/test_quadmesh.py::test_raster_quadmesh_autorange_reversed[dask.array]` </s> 
Not flaky at all: `pytest 'datashader/tests/test_quadmesh.py::test_raster_quadmesh_autorange_reversed[dask.array]'` as this also fails. It is test pollution from the other array...

Want to merge it after my other prs.

### DIFF
--- a/datashader/data_libraries/dask_xarray.py
+++ b/datashader/data_libraries/dask_xarray.py
@@ -173,8 +173,8 @@ def dask_raster(glyph, xr_ds, schema, canvas, summary, *, antialias=False, cuda=
     src_y0, src_y1 = glyph._compute_bounds_from_1d_centers(
         xr_ds, y_name, maybe_expand=False, orient=False
     )
-    xbinsize = float(xr_ds[x_name][1] - xr_ds[x_name][0])
-    ybinsize = float(xr_ds[y_name][1] - xr_ds[y_name][0])
+    xbinsize = abs(float(xr_ds[x_name][1] - xr_ds[x_name][0]))
+    ybinsize = abs(float(xr_ds[y_name][1] - xr_ds[y_name][0]))
 
     # Compute scale/translate
     out_h, out_w = shape
--- a/datashader/datashape/coretypes.py
+++ b/datashader/datashape/coretypes.py
@@ -254,7 +254,7 @@ class DateTime(Unit):
         return np.dtype('datetime64[us]')
 
 
-_units = set(['ns', 'us', 'ms', 's', 'm', 'h', 'D', 'W', 'M', 'Y'])
+_units = ('ns', 'us', 'ms', 's', 'm', 'h', 'D', 'W', 'M', 'Y')
 
 
 _unit_aliases = {
--- a/datashader/tests/conftest.py
+++ b/datashader/tests/conftest.py
@@ -1,0 +1,34 @@
+CUSTOM_MARKS = ("benchmark",)
+
+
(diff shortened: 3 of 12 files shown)


datashader/tests/test_quadmesh.py
```diff
@@ -188,7 +188,8 @@ def test_raster_quadmesh_upsamplex_and_downsampley(array_module):
     assert_eq_xr(res, out)
 
 
-@pytest.mark.parametrize('array_module', array_modules)
+# FIXME: dask.array does not work for this test
```
hoxbro, line 191: I will open an issue, after this PR has been merged.

datashader/data_libraries/dask_xarray.py
```diff
@@ -173,8 +173,8 @@ def dask_raster(glyph, xr_ds, schema, canvas, summary, *, antialias=False, cuda=
     src_y0, src_y1 = glyph._compute_bounds_from_1d_centers(
         xr_ds, y_name, maybe_expand=False, orient=False
     )
-    xbinsize = float(xr_ds[x_name][1] - xr_ds[x_name][0])
-    ybinsize = float(xr_ds[y_name][1] - xr_ds[y_name][0])
+    xbinsize = abs(float(xr_ds[x_name][1] - xr_ds[x_name][0]))
+    ybinsize = abs(float(xr_ds[y_name][1] - xr_ds[y_name][0]))
```
hoxbro, line 177: These were causing the failing test, as they both were `-1`. I can't see any reason why these should not be positive.

datashader/datashape/coretypes.py
```diff
@@ -254,7 +254,7 @@ def to_numpy_dtype(self):
         return np.dtype('datetime64[us]')
 
 
-_units = set(['ns', 'us', 'ms', 's', 'm', 'h', 'D', 'W', 'M', 'Y'])
+_units = ('ns', 'us', 'ms', 's', 'm', 'h', 'D', 'W', 'M', 'Y')
```
hoxbro, line 257: `pytest-xdist` can't handle sets as the order are not consistent, see [here](https://pytest-xdist.readthedocs.io/en/stable/known-limitations.html#order-and-amount-of-test-must-be-consistent).

datashader/tests/conftest.py
```diff
@@ -0,0 +1,34 @@
+CUSTOM_MARKS = ("benchmark",)
```
hoxbro, line 1: Copied from HoloViews. Basically, it makes it so `benchmark` only runs when you pass in `--benchmark`. Benchmark does not work with `pytest-xdist` (it emits a warning if both are enabled at the same time, which is why I disable it with `--benchmark-skip`.

At some point, I would also like to add a GPU marker, so it is possible to run only the GPU tests and not all the tests with GPU tests enabled.

datashader/tests/test_quadmesh.py
hoxbro: I have made a fix in https://github.com/holoviz/datashader/pull/1338/commits/3c48b5929d834cfaeff5cb29173788d65528c821

datashader/data_libraries/dask_xarray.py
philippjfr: Makes sense.

philippjfr (approved):

PR #1340, opened 2024-05-23
Update test run

hoxbro:
- Remove the pip tests to reduce the maintenance burden
- Remove the pyctdev lint step, as this is already handled by pre-commit; this makes it so we can remove a couple of packages.
- Update `codecov` as it no longer works with their package.

Built on top of https://github.com/holoviz/datashader/pull/1338


maximlt:
> Remove the pip tests to reduce the maintenance burden

Did the pip tests fail more than the conda tests?

philippjfr (approved):

hoxbro:
> Did the pip tests fail more than the conda tests?

No, one of the primary reasons it was added to test Python 3.12, which is now also tested in the other test. 

Datashaders CI is also pretty intensive, testing every supported Python version on all platforms. I have noticed it can block the rest of our CI.

PR #1341, opened 2024-05-23
Add gpu marker and test both Classic/dask-expr Dask DataFrames

hoxbro:
By having it as markers it should be easy to test `--gpu` tests separately and not wait for the whole suite to pass. Running the non-GPU test continuously runs on the CI. 

image

The classic / dask-expr was mostly something I wanted to test out if I could get to work, and was "easy" to do when the rest of the test infrastructure was updated.


datashader/tests/conftest.py
```diff
@@ -25,7 +25,7 @@ def pytest_collection_modifyitems(config, items):
             skipped.append(item)
         elif empty:
             selected.append(item)
-        elif not empty and any(m in item.keywords for m in markers):
+        elif not empty and all(m in item.keywords for m in markers):
```
hoxbro, line 28:
```suggestion
        elif not empty and all(m in markers for m in item.keywords):
```

datashader/tests/test_dask.py
```diff
@@ -28,66 +29,72 @@
...
+    pytest.param(_dask_expr, id="dask-expr"),
+    pytest.param(_dask_cudf, marks=pytest.mark.gpu, id="dask-cudf"),
+]
+
+@pytest.fixture(params=_backends)
+def ddf(request):
+    return request.param()
+
+
+@pytest.fixture(params=[1, 2, 4])
```
hoxbro, line 56: I see no reason why to test for 1 to 4, so I removed 3.

philippjfr (approved):

PR #1345, opened 2024-06-13
Add fail-fast for datasets outside the visible extent

raphaelquast:
While trying to improve datashader integration with [EOmaps](https://github.com/raphaelquast/EOmaps) I ran into the following problem:

At the moment, `DSArtist.make_image()` does not check if the data is actually contained within the visible extent before attempting to create an image. 

This results in a huge performance penalty if a large number of datasets are present in a figure but only a few are actually in the visible extent. (e.g. during pan-zoom etc.)

This PR addresses this issue by evaluating the data-bbox on `DSArtist` initialization and then using it to fast-exit `DSArtist.make_image()` in case no data is in the extent.

### DIFF
--- a/datashader/mpl_ext.py
+++ b/datashader/mpl_ext.py
@@ -205,12 +205,22 @@ class DSArtist(_ImageBase):
         self.plot_height = plot_height
         self.width_scale = width_scale
         self.height_scale = height_scale
+        
+        x_col = glyph.x_label
+        y_col = glyph.y_label
+
+        self.bbox_df = Bbox(
+            np.array(
+                [[df[x_col].values.min(), df[y_col].values.min()], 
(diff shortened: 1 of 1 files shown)


philippjfr (approved):
Looks great to me!

raphaelquast:
Awesome! Thanks for the really quick response!

jbednar (approved):
Thanks so much for the contribution! I can't spot any reason not to merge this.

raphaelquast:
Nice! Thanks for having a look!

As i see it, the only drawback that might be relevant is that coordinate limits are now always evaluated while before one it was possible to circuumvent the evaluation by passing explicit values for `x_range` and `y_range`.

jbednar:
Seems like an OK tradeoff to me.

philippjfr (approved):

PR #1350, opened 2024-07-03
Remove Dask as a hard dependency of Datashader

### LINKED ISSUE #1349: Remove Dask as a hard dependency of Datashader
#### Is your feature request related to a problem? Please describe.
When using Datashader (either directly or interactively as a parameter in `hvPlot`) on a project that doesn't use Dask, you get a future warning of the form:
```
FutureWarning: Dask dataframe query planning is disabled because dask-expr is not installed.

You can install it with `pip install dask[dataframe]` or `conda install dask`.
This will raise in a future version.

  warnings.warn(msg, FutureWarning)
```
However, this warning is removed when Dask is added as a dependency of the project

#### Describe the solution you'd like
It will be nice to NOT have dask as a dependency for using datashader, especially when the specific case doesn't require the use of Dask.

#### Describe alternatives you've considered
I have temporarily added Dask as a dependency for the [current examples](https://github.com/holoviz-topics/examples/pull/400/commits/51079db73dccd2c7c3ba6b2e73bcf3dc3f765cd9) I've encountered it

#### Additional context
Example screenshot - Using `rasterize` in hvPlot:
image

hoxbro:
Resolves #1349 

TODO:
- Still need to look at the test


jbednar (commented):
Wow, when we discussed this on Discord yesterday I thought this would go onto the backlog and never get done, but here it is already being addressed! What a nice feeling! Looks good so far; just what I'd expect to see.

hoxbro:
I can see we need to update HoloViews at least in one place:

https://github.com/holoviz/holoviews/blob/eb79ab52414f53b5990f2b624edd4ba582a3fa6e/holoviews/operation/datashader.py#L5

philippjfr (approved):
This is great, thank you!

PR #1365, opened 2024-10-28
refactor!: Drop pillow as a required dependency

hoxbro:
I am not completely sure we should drop it, but we should at least postpone the import until it is needed.


datashader/tiles.py
```diff
@@ -292,11 +291,15 @@ def __init__(self, tile_definition, output_location, tile_format='PNG',
...
         self.post_render_func = post_render_func
 
+        if find_spec("PIL") is None:
+            raise ImportError('pillow is required to render tiles')
         # TODO: add all the formats supported by PIL
         if self.tile_format not in ('PNG', 'JPG'):
             raise ValueError('Invalid output format')
 
     def render(self, da, level):
+        from PIL import fromarray
```
hoxbro, line 301:
```suggestion
        from PIL.Image import fromarray
```

jbednar (approved):
Looks great to me. Happy to relax that requirement, as long as we clearly announce the change in the release notes. Thanks!

amaloney:
@amaloney note to look at dropping pillow or defer importing for now

amaloney (approved):
lgtm

PR #1366, opened 2024-10-29
ci: Move numba to top of dependencies

hoxbro:
Will not solve on `osx-64` because it is in a weird state between `numba` and `python 3.13`

image

### DIFF
--- a/pixi.toml
+++ b/pixi.toml
@@ -22,10 +22,10 @@ build = ["py311", "build"]
 lint = ["py311", "lint"]
 
 [dependencies]
+numba = "*"  # At top as this dictates Python version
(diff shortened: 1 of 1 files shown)


pixi.toml
```diff
@@ -22,10 +22,10 @@ build = ["py311", "build"]
 lint = ["py311", "lint"]
 
 [dependencies]
+numba = "*"  # At top as this dictates Python version
```
maximlt, line 25: The order of the dependencies has some influence?

hoxbro: This is because it would not solve on osx-64, see https://github.com/conda/rattler/issues/913.

PR #1368, opened 2024-11-06
refactor: Add NPY to ruff rules

hoxbro:
Enable Numpy rules: https://docs.astral.sh/ruff/rules/#numpy-specific-rules-npy

The changes are related to NPY002: https://docs.astral.sh/ruff/rules/numpy-legacy-random/#numpy-legacy-random-npy002

I'm not sure why `layout.py` is part of datashader.

### DIFF
--- a/datashader/layout.py
+++ b/datashader/layout.py
@@ -65,10 +65,10 @@ class random_layout(LayoutAlgorithm):
     def __call__(self, nodes, edges=None, **params):
         p = param.ParamOverrides(self, params)
 
-        np.random.seed(p.seed)
+        rng = np.random.default_rng(p.seed)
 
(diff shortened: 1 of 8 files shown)


jbednar (approved):
Looks good, thanks!

> I'm not sure why layout.py is part of datashader.

If these algorithms are available elsewhere in a usable form, I'd be happy to import them conditionally when using Datashader's network rendering. Otherwise, where would they go if not here, if Datashader is to support rendering networks?

hoxbro:
> Otherwise, where would they go if not here, if Datashader is to support rendering networks?

I'm not sure. I've never seen it as part of Datashader to support rendering networks. But I don't have any strong opinions about this.

PR #1373, opened 2024-12-02
compat: Numba 0.61 and Python 3.13

hoxbro:
Windows does not solve because of libxml2 pin and Python 3.13 because of zlib pin.

https://github.com/numba/llvmlite/issues/1105

test-gpu:

image


amaloney:
I must be doing something wrong as I am unable to solve an environment using pixi. I'm running `pixi shell` in order to create the default environment, but instead of creating a default one, it attempts to create all environments in the `environments` table and will not solve anything unless I add the platform to the `features.test-gpu` table.

hoxbro:
> I'm running pixi shell in order to create the default environment

This PR is still in draft; I just wanted to see if PyPi would solve it, which it wouldn't. What you are seeing is the solving of environments with PyPi dependencies, where it needs to download the PyPI package to solve. I have removed this and will wait on the conda-forge release.

hoxbro:
I don't think we can wait anymore for the release of numba and llvmlite on conda-forge for releasing datashader 0.17.0. Currently, it needs https://github.com/conda-forge/llvmlite-feedstock/pull/96 to be merged, followed by https://github.com/conda-forge/numba-feedstock/pull/150.

Most changes to `pixi.toml` can be reverted when numba is released on conda-forge. See `NOTE:` comments.

@amaloney: I see two options for this PR: merge the PR as is or do a new PR with the updates to `README.md` and `pyproject.toml` classifiers. I'm leaning toward the latter.

hoxbro:
superseded by #1403

PR #1377, opened 2024-12-09
doc: remove topics page replaced by a link to examples.holoviz.org

maximlt:
The topics page wasn't very pretty and was basically a bunch of links to examples.holoviz.org, except *Amazon.com center distance* that was linking to https://anaconda.org/defusco/amz_centers/notebook that is no longer available. With this PR it's replaced by an external link to examples.holoviz.org that has even more examples using datashader than originally listed on the topics page.

image

This introduces `sphinx-reredirects` as a docs dependency instead of sphinx rediraffe as the latter doesn't allow redirecting to an external URL unfortunately.


amaloney:
Does `examples.holoviz.org` contain all the examples that were originally on the topics page?

amaloney (approved):
lgtm

PR #1379, opened 2024-12-12
chore: Update pre-commit and add COVERAGE_CORE = "sysmon"

hoxbro:
For sysmon see https://github.com/holoviz/holoviews/pull/6465

Mac + Python 3.12 goes down from 11 min and 32 sec to 8 min and 2 sec.


pyproject.toml
```diff
@@ -86,7 +86,6 @@ select = [
     "W",
 ]
 ignore = [
-    "UP027",  # unpacked-list-comprehension (deprecated)
```
hoxbro, line 89: Has been removed

Issue #1082, opened 2022-05-04
dsshow savefig error with matplotlib pdf and svg export. Shading intensity variation in png export.


rotheconrad:
#### ALL software version info
redhat:enterprise_linux:7.9
Linux 3.10.0-1160.49.1.el7.x86_64
Python 3.8.12
Matplotlib 3.2.2
datashader 0.14.0
pandas 1.1.3
numpy 1.21.6

#### Description of expected behavior and the observed behavior
The datashader plotted points should be the same size as the ax and export well but the scale is off and changes with dpi export settings. Additionally, the shading intensity is variable for png export.

#### Complete, minimal, self-contained example code that reproduces the issue
```
import matplotlib
matplotlib.use('Agg')
import matplotlib.pyplot as plt
import datashader as ds
from datashader.mpl_ext import dsshow
import pandas as pd
import numpy as np

# Fake data for testing
x = np.random.normal(size=100000)
y = x * 3 + np.random.normal(size=100000)

fig, ax = plt.subplots()
dsartist = dsshow(
                df,
                ds.Point("xs", "ys"),
                ds.count(),
                vmin=0,
                vmax=100,
                norm="linear",
                aspect="auto",
                ax=ax
                )
plt.title('300 dpi')
fig.savefig(test_300dpi.pdf, dpi=300)
fig.savefig(test_300dpi.png, dpi=300)
fig.savefig('test_300dpi.svg', dpi=300)
plt.close()

fig, ax = plt.subplots()
dsartist = dsshow(
                df,
                ds.Point("xs", "ys"),
                ds.count(),
                vmin=0,
                vmax=100,
                norm="linear",
                aspect="auto",
                ax=ax
                )
plt.title('default dpi')
fig.savefig(test_defaultdpi.pdf)
fig.savefig(test_defaultdpi.png)
fig.savefig('test_defaultdpi.svg')
plt.close()
```
#### Screenshots or screencasts of the bug in action
<img width="838" alt="image" src="https://user-images.githubusercontent.com/12850144/166799907-5c5f1b0b-89fd-4d45-b322-456686a9a119.png">
<img width="1237" alt="image" src="https://user-images.githubusercontent.com/12850144/166800028-992dfbff-0fc1-421f-b859-ea3b6f55e0e6.png">
<img width="1053" alt="image" src="https://user-images.githubusercontent.com/12850144/166800223-4b836158-ab77-4d0e-9791-53d14f84ce0e.png">
[...]

PR #1382, opened 2025-01-03
Fix: PDF Scaling Bug

thomas-reimonn:
Datashader currently has a known bug where, upon saving figures to PDF, the rasterized portion of the Datashader output is scaled differently from the vector-based elements (such as axes, labels, etc.). This PR fixes the issue and demonstrates it by ensuring the Datashader images render correctly within a single PDF figure.


nvictus:
This fixes the `dsshow` issue #1082

amaloney:
There are two problems discussed in the linking issue #1082.

1. Data being transformed without explicit user settings
2. Color variations due to the type of image being saved (bitmap vs vector)

### 1. Data being transformed without explicit user settings

I was able to reproduce the unexpected data scaling based on the type of image being saved. The environment and modified MRE are shown below in the code section. To get the expected behavior, uncomment out the code below the note.

#### Unexpected data scaling (no explicit plot dimensions given for vector images)

<p align="center">
  
  
</p>
<p align="center">
  
  
</p>
<p align="center">
  
  
</p>

#### Expected behavior (explicit plot dimensions given for vector images)

<p align="center">
  
  
</p>
<p align="center">
  
  
</p>
<p align="center">
  
  
</p>

`datashader` is applying an affine transformation when it outputs to a pdf or svg (vector image). It does not apply a transformation when a png (bitmap) is the final output. The transformation behavior is left to the end user to suss out on their own based on the image being drawn, which I think is bad user experience. The docstring for the `make_image` method discusses an affine transformation being applied if the kwarg `unsampled` is `True`.

https://github.com/holoviz/datashader/blob/6802220257a56166d685536798c128a08414f82c/datashader/mpl_ext.py#L255-L272

`datashader` will switch this value and apply a transformation if the final image size is not explicitly given, and the user is saving as a vector image.

https://github.com/holoviz/datashader/blob/6802220257a56166d685536798c128a08414f82c/datashader/mpl_ext.py#L282-L288

I need to investigate further, but I think the call to `make_image` is happening when a user saves the image to disk. Unfortunately when the user is saving to a vector image, the resulting image size is unbounded and the `unsampled` kwarg is being switched, thus a transformation on the data occurs.

### 2. Color variations due to the type of image being saved (bitmap vs vector)

Pixel color values are computed based on a combination of

1. type of glyph used;
2. the glyph geometry;
3. the glyph size;
4. the space available for the final image; and
5. the overlap of glyphs within the final image's area.

You can see this behavior in the following movie where I modify the final image's total area by making it progressively smaller using the supplied notebook, but not using the given PR changes. This behavior is expected, but may not be documented well enough.

[](https://github.com/user-attachments/assets/8b87b145-29bb-4379-b0a5-6552b9fcd868)

@nvictus and @thomas-reimonn the PR modifies default `datashader` behavior, and I am still fairly new to the codebase so I do not think I can approve it without more discussions with @jbednar.

---

#### Code

Install the environment and run the MRE.

```bash
mamba env create environment.yaml && mamba activate datashader-issue1082
python example.py
```

```yaml
# environment.yaml
name: datashader-issue1082
channels:
    - conda-forge
dependencies:
    - python ==3.8.12

    # Package managers
    - pip

    # Dependencies
    - datashader ==0.14.0
    - matplotlib ==3.2.2
    - numpy ==1.21.6
    - pandas ==1.1.3
```

```python
# example.py
import datashader as ds
import numpy as np
import pandas as pd
from datashader.mpl_ext import dsshow

EXTENSIONS = ["pdf", "png", "svg"]
SIZE = 100_000
DPIS = ["300", "default"]


def main():
    # Fake data for testing
    x = np.random.normal(size=SIZE)
    y = x * 3 + np.random.normal(size=SIZE)
    df = pd.DataFrame(columns=["xs", "ys"], data=np.array([x, y]).T)

    for dpi in DPIS:
        for extension in EXTENSIONS:
            fig, ax = plt.subplots()
            dsartist = dsshow(
                df=df,
                glyph=ds.Point("xs", "ys"),
                aggregator=ds.count(),
                vmin=0,
                vmax=100,
                norm="linear",
                aspect="auto",
                ax=ax,
                # NOTE: Uncomment without the PR changes to get similarly scaled images.
                # plot_width=1920,
                # plot_height=1440,
            )
            plt.title(f"{dpi} dpi {extension}")
            if dpi == "300":
                fig.savefig(f"test_{dpi}_dpi.{extension}", dpi=300)
            else:
                fig.savefig(f"test_{dpi}_dpi.{extension}")
            plt.close()


if __name__ == "__main__":
    main()
```

nvictus:
@amaloney not sure yet if this addresses your concerns, but here is an explanation of the change:

The `DSArtist.make_image` artist method is called by matplotlib during rendering. If you look at what [matplotlib does](https://github.com/matplotlib/matplotlib/blob/main/lib/matplotlib/image.py#L576-L604), it depends on the rendering backend:

1. If a renderer supports affine transforms (e.g. in a notebook), matplotlib passes `unsampled=True` to `make_image` and passes an affine transform `trans` to `draw_image`.

2. If a renderer does NOT support affine transforms (like PDF), matplotlib calls `make_image` without passing `unsampled` (assuming it is False), and then calls `draw_image` without passing in an affine transform. In this case, I believe matplotlib rescales the image using a sampling method, but maybe @tacaswell could clarify.

Notably, matplotlib has the **expectation that the default value of unsampled is False**. It was my mistake to set the default value to True in my original implementation of the DSArtist, as I wanted to avoid additional manipulation by matplotlib. This didn't matter in scenario 1, so it only surfaces when using a matplotlib renderer that does not support affine transformation.

amaloney (changes requested):
Overall, this LGTM. I don't think the `datashader-image.pdf` file, nor the notebook are necessary since they are both there as examples showing the bug and the fix.

thomas-reimonn:
Thank you for the feedback on this! I removed the demo notebook and image.

jbednar:
Thanks for catching that!

PR #1383, opened 2025-01-07
Speed up edge bundling

lmcinnes:
The hammer edge bundling is fantastic, but can be quite time consuming for large graphs (100,000+ edges and upward). I spent some time  benchmarking the code, and then doing some profiling to determine how the time was being spent, and then attempted to make some minor improvements. I quickly discovered that, at least on the machines and setups I tried, the dask usage actually made it significantly *slower*. I have therefore made dask optional, with a param ``use_dask`` which defaults to ``False``. I also spent a while trying to wring the most I could out of numba for many of the core or frequently called functions. Primarily this involved adding more annotations to the decorators, and a careful rewrite of the distance function which is called extremely often. The remainder of the work was re-juggling when and where various computations were done to avoid duplicate work, or move more loops inside numba where possible. Lastly I rewrote ``_convert_edge_segments_to_dataframe`` to use a single large numpy allocation followed by ``zip`` and ``chain`` rather than a generator with many many small allocations (the code is a little less readable, but significantly faster for very large graphs).

After all these changes a typical use case for me (a knn-graph) went from 1h20m to 15s, and scaled up versions of the random graph examples from the docs (with ``n=500`` and  ``m=100000``) went from 1h 13min for circular layout and 1h 39min for force-directed layout to 1min 38s and 60s respectively. This would make edge-bundling and graph drawing with datashader (which I love!) far more practical for a much wider range of datasets.

I'll be happy to discuss the changes as required -- some are more important than others, but I went down the optimization rabbit hole and did all the things I could.

### DIFF
--- a/datashader/bundling.py
+++ b/datashader/bundling.py
@@ -198,31 +198,23 @@ def smooth(edge_segments, tension, idx, idy):
         smooth_segment(segments, tension, idx, idy)
 
 @nb.jit(
-    nb.void(nb.float32[:,::1], nb.float32[:,::1], nb.float32[:,::1], nb.float64, nb.int64, nb.int64),
+    nb.float32[:,::1](nb.float32[:,::1], nb.float32[:,::1], nb.float32[:,::1], nb.int64, nb.float64, 
+                      segment_length_type, nb.uint64, nb.uint64, nb.int64),
     nopython=True,
     nogil=True,
     fastmath=True,
-    locals={"x": nb.uint16, "y": nb.uint16}
-)
-def advect_segments(segments, vert, horiz, accuracy, idx, idy):
-    for i in range(1, len(segments) - 1):
-        x = np.uint16(segments[i, idx] * accuracy)
-        y = np.uint16(segments[i, idy] * accuracy)
-        segments[i, idx] += horiz[x, y] / accuracy
-        segments[i, idy] += vert[x, y] / accuracy
-        segments[i, idx] = max(0, min(segments[i, idx], 1))
-        segments[i, idy] = max(0, min(segments[i, idy], 1))
-
-
-@nb.jit(
-    nb.float32[:,::1](nb.float32[:,::1], nb.float32[:,::1], nb.float32[:,::1], nb.int64, nb.float64, segment_length_type, nb.int64, nb.int64, nb.int64),
-    nopython=True,
-    nogil=True,
-    locals={'it': nb.uint8}
+    locals={'it': nb.uint8, "i": nb.uint16, "x": nb.uint16, "y": nb.uint16}
 )
 def advect_and_resample(vert, horiz, segments, iterations, accuracy, squared_segment_length, idx, idy, ndims):
     for it in range(iterations):
-        advect_segments(segments, vert, horiz, accuracy, idx, idy)
+        for i in range(1, len(segments) - 1):
+            x = np.uint16(segments[i, idx] * accuracy)
+            y = np.uint16(segments[i, idy] * accuracy)
+            segments[i, idx] += horiz[x, y] / accuracy
+            segments[i, idy] += vert[x, y] / accuracy
+            segments[i, idx] = max(0, min(segments[i, idx], 1))
+            segments[i, idy] = max(0, min(segments[i, idy], 1))
+
         if it % 2 == 0:
             segments = resample_edge(segments, squared_segment_length, ndims)
     return segments
--- a/datashader/bundling.py
+++ b/datashader/bundling.py
@@ -194,7 +194,6 @@ def smooth_segment(segments, tension, idx, idy):
                             (tension*(previous[idx] + next_point[idx]) / 2))
             current[idy] = (((1-tension)*current[idy]) +
                             (tension*(previous[idy] + next_point[idy]) / 2))
-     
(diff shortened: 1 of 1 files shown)


jbednar:
That all sounds really promising. Thanks for the contribution! We'll review and let you know, but as @amaloney suggests, posting some benchmarking examples in the associated issue would be really useful.

amaloney:
I've had the opportunity to add the `parallel=True` flag to spots in the module where numba can parallelize loops, as well as adding type signatures to the other parts of the code that have the jit decorator. I am not able to squeeze out any more optimization without diving really deep, which should not hold up this PR as it functions very well right now.

@lmcinnes I would suggest making new PRs if you have other ideas or thoughts on how to optimize this module further. I plan to take what you have done and apply it to other areas of the codebase, as I find using the explicit flags inside the `numba.jit` decorator **much** easier to read than the nonstandard `@ngjit` currently being used.

datashader/bundling.py
```diff
@@ -35,36 +36,64 @@ def func(*args, **kwargs):
...
 
-@ngjit
+
+@nb.jit(
+    nb.float32(nb.float32[::1], nb.float32[::1]),
+    nopython=True,
+    nogil=True,
+    fastmath=True,
+    locals={"result": nb.float32, "diff": nb.float32},
+)
```
amaloney, line 54: I'm a big fan of the explicit nature of this decorator. I think we should use this as a good example of why being terse (_e.g._ using `@ngjit`) is not always better.

Also, I like the usage of the trailing comma

datashader/bundling.py
```diff
@@ -35,36 +36,64 @@ def func(*args, **kwargs):
...
-    return (((a[0] - b[0]) ** 2) + ((a[1] - b[1]) ** 2))**(0.5)
-
-
-@ngjit
-def resample_segment(segments, new_segments, min_segment_length, max_segment_length, ndims):
+    diff = a[0] - b[0]
+    result = diff * diff
+    diff = a[1] - b[1]
+    result += diff * diff
+    return result
```
amaloney, line 61: The previous code took the square root to find the Euclidean distance, while the new addition does not. Computationally it doesn't matter, but we should update the docstring to let the reader know it doesn't matter, in case someone has never come across this before. I'm thinking about future readers that may consider it a bug.

datashader/bundling.py
```diff
@@ -35,36 +36,64 @@ def func(*args, **kwargs):
...
+    fastmath=True,
+    locals={
+        'next_point': nb.float32[::1],
+        'current_point': nb.float32[::1],
+        'step_vector': nb.float32[::1],
+        'i': nb.uint16,
+        'pos': nb.uint64,
+        'index': nb.uint64,
+        'distance': nb.float32
+    }
```
amaloney, line 76: superb usage of locals, that I am going to use in the codebase everywhere

datashader/bundling.py
```diff
@@ -77,96 +106,127 @@ def resample_segment(segments, new_segments, min_segment_length, max_segment_len
...
     total += 1
-    return any_change, total
-
-
-def resample_edge(segments, min_segment_length, max_segment_length, ndims):
-    change, total_resamples = calculate_length(segments, min_segment_length, max_segment_length)
+    return any_change, total, n_points_to_add
+
+@nb.jit(
+    nb.float32[:,::1](nb.float32[:,::1], segment_length_type, nb.int64),
```
amaloney, line 156: This is a good example of why we may want to include formatting rules

```suggestion
    nb.float32[:, ::1](nb.float32[:, ::1], segment_length_type, nb.int64),
```

datashader/bundling.py
```diff
@@ -77,96 +106,127 @@ def resample_segment(segments, new_segments, min_segment_length, max_segment_len
...
-                        max_segment_length, segment_class):
+@nb.jit(
+    nb.float32[:,::1](nb.float32[:,::1], nb.float32[:,::1], nb.float32[:,::1], nb.int64, nb.float64,
+                      segment_length_type, nb.uint64, nb.uint64, nb.int64),
+    nopython=True,
+    nogil=True,
+    fastmath=True,
+    locals={'it': nb.uint8, "i": nb.uint16, "x": nb.uint16, "y": nb.uint16}
+)
+def advect_and_resample(vert, horiz, segments, iterations, accuracy, squared_segment_length,
```
amaloney, line 210: `accuracy` should be `nb.int16`, see line 521. To me this says we might want to incorporate static type checking, but that's on us to implement for the future.

datashader/bundling.py
```diff
@@ -472,6 +544,17 @@ def __call__(self, nodes, edges, **params):
...
+            resample_edges_fn = delayed(resample_edges)
+            draw_to_surface_fn = delayed(draw_to_surface)
+            get_gradients_fn = delayed(get_gradients)
+            advect_resample_all_fn = delayed(advect_resample_all)
+        else:
+            resample_edges_fn = resample_edges
+            draw_to_surface_fn = draw_to_surface
+            get_gradients_fn = get_gradients
+            advect_resample_all_fn = advect_resample_all
+
```
amaloney, line 557: nice

amaloney (approved):
lgtm

Issue #1392, opened 2025-01-21
drop Python 3.9


amaloney:
See the references

- https://holoviz.org/about/heps/hep1.html
- https://scientific-python.org/specs/spec-0000/

PR #1399, opened 2025-01-28
enh!: Drop Python 3.9

hoxbro:
resolves https://github.com/holoviz/datashader/issues/1392


hoxbro:
The missing coverage is related to the spatialpandas test no longer being run. They are skipped because they only support the legacy implementation. 

Ref issue: https://github.com/holoviz/spatialpandas/issues/146

amaloney (approved):
lgtm

PR #1403, opened 2025-01-30
ci: Add numba channel and test python 3.13

hoxbro:
We can drop the numba channel when numba is released on conda-forge.

image

### DIFF
--- a/pixi.toml
+++ b/pixi.toml
@@ -1,6 +1,6 @@
 [project]
 name = "datashader"
-channels = ["pyviz/label/dev", "conda-forge"]
+channels = ["pyviz/label/dev", "numba", "conda-forge"]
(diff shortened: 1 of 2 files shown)


pixi.toml
```diff
@@ -14,15 +14,16 @@ PYTHONIOENCODING = "utf-8"
...
 test-gpu = ["py312", "test-core", "test-gpu"]
 docs = ["py311", "example", "doc"]
 build = ["py311", "build"]
 lint = ["py311", "lint"]
 
 [dependencies]
-python = "<3.13"
-numba = "*"
+numba = { version = "*", channel = "numba" }
+llvmlite = { version = "*", channel = "conda-forge" }
```
amaloney, line 26: :+1:

amaloney (approved):
Thanks for this. I'm a fan of how we can change channels in the `pixi.toml` file.

PR #1408, opened 2025-02-12
naive jit implementation for cooling

dmiracle:
Tried out a simple numba implementation for layout cooling and got more than a 10x speedup. I made some questionable choices (`matrix = matrix.toarray()`) but just putting it up to say that the approach gave a good speedup.


amaloney (approved):
Great work, thanks again for the contribution

PR #1415, opened 2025-04-09
chore: bump python version

amaloney:


hoxbro:
To give some background, this should have been done in the last release, 0.17.0, but it was not. Since so much time has passed, we should not yank the release. 

The best way (at least to me) to handle this is to include it in the next release. I suggest 0.18.0, with a small note in the release notes explaining this.

hoxbro (approved):

PR #1419, opened 2025-05-14
refactor: wip attempt at lazy import modules

hoxbro:
Trying to see if I could lazy import modules, so we don't get such a heavy import cost of datashader. 

This is heavily inspired by the work done in https://github.com/holoviz/holoviews/pull/6476, also here an import hook for when the module is imported. Though, I have chosen to have the import here in a centralized file as this matches better with how datashader is structured. 

Nowhere near finishing in the transition ~~, also currently dask stalls and don't know why...~~

--- 
# Benchmark
### Main
``` bash
❯ hyperfine "python -c ''" "python -c 'import datashader'" --warmup 5
Benchmark 1: python -c ''
  Time (mean ± σ):      10.0 ms ±   0.8 ms    [User: 6.8 ms, System: 3.0 ms]
  Range (min … max):     7.9 ms …  12.7 ms    263 runs

Benchmark 2: python -c 'import datashader'
  Time (mean ± σ):      1.099 s ±  0.017 s    [User: 2.510 s, System: 0.164 s]
  Range (min … max):    1.077 s …  1.125 s    10 runs

Summary
  python -c '' ran
  109.99 ± 8.70 times faster than python -c 'import datashader'

```

### This branch ([7648773](https://github.com/holoviz/datashader/pull/1419/commits/7648773d331b71386106a54d2a9f49d7ae7ca329))
``` bash
❯ hyperfine "python -c ''" "python -c 'import datashader'" --warmup 5
Benchmark 1: python -c ''
  Time (mean ± σ):      10.0 ms ±   0.8 ms    [User: 7.0 ms, System: 2.9 ms]
  Range (min … max):     7.8 ms …  11.8 ms    257 runs

Benchmark 2: python -c 'import datashader'
  Time (mean ± σ):     558.2 ms ±   7.5 ms    [User: 2047.4 ms, System: 96.5 ms]
  Range (min … max):   544.8 ms … 568.1 ms    10 runs

Summary
  python -c '' ran
   55.59 ± 4.45 times faster than python -c 'import datashader'
```

## Tuna
```
❯ python -X importtime -c 'import datashader' 2> tuna.log && tuna tuna.log
```

### Main
image

### This branch ([7648773](https://github.com/holoviz/datashader/pull/1419/commits/7648773d331b71386106a54d2a9f49d7ae7ca329))

image

### DIFF
--- a/datashader/_dependencies.py
+++ b/datashader/_dependencies.py
@@ -68,6 +68,8 @@ class _LazyModule:
         self.__module_name = module_name
         self.__package_name = package_name or module_name
         self.__bool_use_sys_modules = bool_use_sys_modules
+        if module_name in sys.modules:
+            self._module
 
     @property
     def _module(self):
@@ -113,8 +115,13 @@ class _LazyModule:
         return self.__module and self.__module.__version__ or _get_version(self.__package_name)
 
 
-def register_import_hook(module:  _LazyModule, fn: Callable):
-    _module_hooks[module].append(fn)
+def register_import_hook(module: _LazyModule, fn: Callable):
+    # NOTE: We cannot use module.__module,
+    # because it is overwritten by __getattr__
+    if module.__dict__["_LazyModule__module"]:
+        fn()
+    else:
+        _module_hooks[module].append(fn)
 
 
 TYPE_CHECKING = False
--- a/datashader/_dependencies.py
+++ b/datashader/_dependencies.py
@@ -1,0 +1,155 @@
+import re
+import sys
+from collections import defaultdict
+from functools import lru_cache
+from importlib import import_module
+from importlib.metadata import PackageNotFoundError, version
+from importlib.util import find_spec
+from collections.abc import Callable
+
+_re_no = re.compile(r"\d+")
+
+
+class VersionError(Exception):
+    """Raised when there is a library version mismatch."""
+
+    def __init__(self, msg, version=None, min_version=None, **kwargs):
+        self.version = version
+        self.min_version = min_version
+        super().__init__(msg, **kwargs)
+
+
+@lru_cache
+def _is_installed(module_name):
+    return find_spec(module_name) is not None
+
+
+@lru_cache
+def _get_version(package_name):
+    try:
+        return version(package_name)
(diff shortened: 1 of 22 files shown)


jbednar:
Cool, thanks. Note that there are at least two sources of long startup time for Datashader, i.e. imports and Numba compilation. I think only once Numba can be precompiled and/or cached (which is _supposed_ to be possible already) will the startup time be reasonable.

hoxbro:
> Cool, thanks. Note that there are at least two sources of long startup time for Datashader, i.e. imports and Numba compilation. I think only once Numba can be precompiled and/or cached (which is supposed to be possible already) will the startup time be reasonable.

Yes, it could be great also to tackle that, but just by looking at the profiler, I can see, e.g., `dask.dataframe` accounted for half of the import time, and pandas a little below  2/10.  

If I lazy load pandas (and therefore xarray and comment out some ragged extension inheritance for now), I can get it down to the following: 

``` python
❯ hyperfine "python -c ''" "python -c 'import datashader'" --warmup 5
Benchmark 1: python -c ''
  Time (mean ± σ):       9.6 ms ±   0.7 ms    [User: 6.6 ms, System: 2.9 ms]
  Range (min … max):     7.7 ms …  11.8 ms    277 runs

Benchmark 2: python -c 'import datashader'
  Time (mean ± σ):     298.7 ms ±  14.9 ms    [User: 1847.7 ms, System: 57.0 ms]
  Range (min … max):   284.5 ms … 331.5 ms    10 runs

Summary
  python -c '' ran
   31.19 ± 2.67 times faster than python -c 'import datashader'
```

image

hoxbro:
> What I was bringing up was the time to the first datashaded result, not the time for an import that isn't used.

I understand that this is the main problem, but it is outside the scope of this PR. But something we should improve at some point.

> Are we often importing datashader without actually using it?

I don't think so, but when we do import it, we import a lot of stuff we don't necessarily need. For example, if we want to rasterize a pandas DataFrame, we try to import dask.DataFrame. This, at least for me, causes unnecessary slowdown, because I have Dask installed in my development environment.

PR #1425, opened 2025-06-18
fix: refactor failing tests

amaloney:
- Updates fixture data
- Skips tests based on pandas errors

### DIFF
--- a/datashader/tests/test_datatypes.py
+++ b/datashader/tests/test_datatypes.py
@@ -716,10 +716,8 @@ class TestRaggedGetitem(eb.BaseGetitemTests):
     def test_getitem_series_integer_with_missing_raises(self, data, idx):
         pass
 
-    @pytest.mark.xfail(reason="pytest empty string warning")
+    @pytest.mark.filterwarnings("ignore::pytest.PytestWarning")
     def test_take_pandas_style_negative_raises(self, data, na_value):
-        # Backport may not happen.
-        # https://github.com/pandas-dev/pandas/issues/61557
         super().test_take_pandas_style_negative_raises(data, na_value)
 
 


datashader/tests/test_geopandas.py
```diff
@@ -58,26 +58,26 @@
 ])
 
 nybb_points_sol = np.array([
```
hoxbro, line 62:
```suggestion
# This data is for checking repeatability in our tests, not correctness.
# GeoPandas 1.1.0 updated the sampling of points in https://github.com/geopandas/geopandas/pull/3471
nybb_points_sol = np.array([
```

The first line should maybe also be added above `nybb_lines_sol`

datashader/tests/test_datatypes.py
```diff
@@ -841,6 +850,12 @@ def test_duplicated(self, data):
         # https://github.com/pandas-dev/pandas/pull/55255
         super().test_duplicated(data)
 
+    @pytest.mark.xfail(reason="pytest empty string warning")
+    def test_argmax_argmin_no_skipna_notimplemented(self, data_missing_for_sorting):
+        # Backport may not happen.
+        # https://github.com/pandas-dev/pandas/issues/61557
+        super().test_argmax_argmin_no_skipna_notimplemented(data_missing_for_sorting)
```
hoxbro:
```suggestion
    @pytest.mark.filterwarnings("ignore::pytest.PytestWarning")
    def test_argmax_argmin_no_skipna_notimplemented(self, data_missing_for_sorting):
        super().test_argmax_argmin_no_skipna_notimplemented(data_missing_for_sorting)
```

I think we should still run it and ignore this warning, same with the other one.

datashader/tests/test_datatypes.py
```diff
@@ -730,8 +736,7 @@ def test_groupby_extension_transform(self):
     def test_groupby_extension_agg(self):
         pass
 
-    @pytest.mark.skip(
-        reason="numpy.ndarray unhashable and buffer wrong number of dims")
+    @pytest.mark.skip(reason="numpy.ndarray unhashable and buffer wrong number of dims")
     def test_groupby_extension_apply(self):
         pass
```
hoxbro:
```suggestion
    @pytest.mark.xfail(reason="numpy.ndarray unhashable and buffer wrong number of dims")
    def test_groupby_extension_apply(self):
        super().test_groupby_extension_apply()
```

Let's just run it; it could be that we solve this by accident at some point. 

Also, is this a new extension test added in pandas 2.3.0? I know `test_array_interface_copy` is.

amaloney: I believe so

hoxbro: Can you verify it?

datashader/tests/test_datatypes.py
```diff
@@ -716,6 +716,12 @@ def test_getitem_invalid(self, data):
     def test_getitem_series_integer_with_missing_raises(self, data, idx):
         pass
 
+    @pytest.mark.xfail(reason="pytest empty string warning")
+    def test_take_pandas_style_negative_raises(self, data, na_value):
```
hoxbro: Forgot to update this.

datashader/tests/test_datatypes.py
amaloney: It was not added in 2.3.0 and I'm not sure when it was, but it ends with a `TypeError: Series.name must be a hashable type` from `pandas.core.dtypes.common.py::validate_all_hashable(*args, error_name: str | None = None) -> None:`. The error name is `Series.name` and the args of the method is a NumPy array. I'm not sure if we are doing anything to get this error, but it seems like knowing it will fail is fine.

hoxbro:
Thank you :+1:

datashader/tests/test_datatypes.py
hoxbro: Just noticed that your change was only linting. This will always fail because the `test_groupby_extension_apply` needs two arguments.

PR #1440, opened 2025-08-27
Use numba for faster estimation of curvilinear bounds

dcherian:
Disclosure: written by Claude mostly

1. Unit	tests pass locally
2. Our internal test suite passes too

if you'd like i can add a hypothesis property test checking that this numba version is equivalent to the old numpy version run twice

### DIFF
--- a/datashader/resampling.py
+++ b/datashader/resampling.py
@@ -997,7 +997,7 @@ def infer_interval_breaks_2d(coord):
     """
     Optimized Numba version for 2D arrays.
     Equivalent to applying infer_interval_breaks sequentially on axis=1 then axis=0.
-    ~2.1x faster than NumPy with true parallelism (nogil=True).
+    Combines both passes for interior points to avoid intermediate array.
 
     Parameters:
     -----------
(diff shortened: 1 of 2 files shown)


datashader/resampling.py
```diff
@@ -990,3 +990,75 @@ def infer_interval_breaks(coord, axis=0):
...
+    Returns:
+    --------
+    ndarray
+        2D array with interval boundaries, shape (m+1, n+1)
+    """
+    coord_float = coord.astype(np.float64)
+    m, n = coord_float.shape
+    
+    if m == 0 or n == 0:
+        return np.empty((m + 1, n + 1), dtype=coord_float.dtype)
```
hoxbro: Shouldn't something like this be `np.zeros`? I only think it is OK to use `np.empty` if we fill it up afterwards.

datashader/resampling.py
```diff
@@ -990,3 +990,75 @@ def infer_interval_breaks(coord, axis=0):
...
+    -----------
+    coord : ndarray
+        2D coordinate array representing grid centers
+
+    Returns:
+    --------
+    ndarray
+        2D array with interval boundaries, shape (m+1, n+1)
+    """
+    coord_float = coord.astype(np.float64)
```
hoxbro: Not sure why this conversion is needed... 

I think we should use the logic as in `infer_interval_breaks`:

``` python
    if cupy and isinstance(coord, cupy.ndarray):
        # leave cupy array as-is
        pass
    else:
        coord = np.asarray(coord)
```

datashader/resampling.py
```diff
@@ -990,3 +990,75 @@ def infer_interval_breaks(coord, axis=0):
...
+    m, n = coord_float.shape
+    
+    if m == 0 or n == 0:
+        return np.empty((m + 1, n + 1), dtype=coord_float.dtype)
+
+    # Step 1: Process along axis=1 (expand columns)
+    temp = np.empty((m, n + 1), dtype=coord_float.dtype)
```
hoxbro: If we end up keeping `coord_float` as is, this should just be `np.float64`.

datashader/resampling.py
```diff
@@ -990,3 +990,75 @@ def infer_interval_breaks(coord, axis=0):
...
                       for n in range(coord.ndim))
     return np.concatenate([first, coord[trim_last] + deltas, last], axis=axis)
+
+
+@ngjit
+def infer_interval_breaks_2d(coord):
+    """
+    Optimized Numba version for 2D arrays.
+    Equivalent to applying infer_interval_breaks sequentially on axis=1 then axis=0.
+    ~2.1x faster than NumPy with true parallelism (nogil=True).
```
hoxbro: Have you measured this yourself?

datashader/resampling.py
hoxbro: Thinking a bit about it. I don't think we can use cupy.arrays, so maybe we should add an if statement before running this optimized version.

dcherian:
Yes, now up to 3.5x with the single-pass version in the latest commit.
```
Microbenchmark Results (500x500 array, 10 iterations):
NumPy version:  0.0010s per call
Numba version:  0.0003s per call
Speedup:        3.60x
🚀 Numba is 3.60x faster!

--- Testing 1000x1000 array ---
Running NumPy benchmark (10 iterations)...
Running Numba benchmark (10 iterations)...

Microbenchmark Results (1000x1000 array, 10 iterations):
NumPy version:  0.0052s per call
Numba version:  0.0014s per call
Speedup:        3.79x
🚀 Numba is 3.79x faster!

--- Testing 5000x5000 array ---
Running NumPy benchmark (3 iterations)...
Running Numba benchmark (3 iterations)...

Microbenchmark Results (5000x5000 array, 3 iterations):
NumPy version:  0.1614s per call
Numba version:  0.0472s per call
Speedup:        3.42x
🚀 Numba is 3.42x faster!
```

<details>

```python
#!/usr/bin/env python3
import sys
import os
import numpy as np
import timeit

from datashader.resampling import infer_interval_breaks, infer_interval_breaks_2d

if __name__ == "__main__":
    print("Testing Numba version correctness...")
    test_arr = np.arange(12.0).reshape(3, 4)
    print(f"Input array:\n{test_arr}")
    
    result_numpy = infer_interval_breaks(infer_interval_breaks(test_arr, axis=0), axis=1)
    print("NumPy result:")
    print(result_numpy)
    
    result_numba = infer_interval_breaks_2d(test_arr)
    print("Numba result:")
    print(result_numba)
    
    try:
        np.testing.assert_array_almost_equal(result_numpy, result_numba, decimal=10)
        print("✓ Numba version matches NumPy perfectly!")
    except AssertionError:
        print("✗ Results don't match!")
        print(f"Max difference: {np.max(np.abs(result_numpy - result_numba))}")

    sizes = [(500, 500), (1000, 1000), (5000, 5000)]
    for size in sizes:
        rows, cols = size
        print(f"\n--- Testing {rows}x{cols} array ---")
        
        test_array = np.arange(rows * cols, dtype=np.float64).reshape(rows, cols)
        _ = infer_interval_breaks_2d(test_array)
        
        def numpy_version():
            return infer_interval_breaks(infer_interval_breaks(test_array, axis=0), axis=1)
        
        def numba_version():
            return infer_interval_breaks_2d(test_array)
        
        iterations = 10 if rows <= 1000 else 3
        
        print(f"Running NumPy benchmark ({iterations} iterations)...")
        time_numpy = timeit.timeit(numpy_version, number=iterations)
        
        print(f"Running Numba benchmark ({iterations} iterations)...")
        time_numba = timeit.timeit(numba_version, number=iterations)
        
        print(f"\nMicrobenchmark Results ({rows}x{cols} array, {iterations} iterations):")
        print(f"NumPy version:  {time_numpy/iterations:.4f}s per call")
        print(f"Numba version:  {time_numba/iterations:.4f}s per call")
        print(f"Speedup:        {time_numpy/time_numba:.2f}x")
        
        if time_numba < time_numpy:
            print(f"🚀 Numba is {time_numpy/time_numba:.2f}x faster!")
        else:
            print(f"⚠️  Numba is {time_numba/time_numpy:.2f}x slower")
```

</details>

datashader/resampling.py
```diff
@@ -990,3 +990,71 @@ def infer_interval_breaks(coord, axis=0):
...
+        2D coordinate array representing grid centers
+
+    Returns:
+    --------
+    ndarray
+        2D array with interval boundaries, shape (m+1, n+1)
+    """
+    m, n = coord.shape
+
+    assert m > 1
```
dcherian, line 1014: afaict quadmesh expects at least size-2 dimensions even though no clear error message is raised saying this

dcherian:
Actually I am going to close this one and iterate some more. The correct single pass version is quite messy.

FWIW I don't really understand how calling `infer_interval_breaks` twice gets the right answer

PR #1441, opened 2025-08-28
fix: Handle descending coordinates in QuadMeshRaster downsampling

### LINKED ISSUE #1439: QuadMeshRectilinear returning all NaN in some cases
This one is pretty weird :/, either broadcasting out coordinates or sorting the y-coordinate makes it work.

(this library, plus any other relevant software, e.g. bokeh, python, notebook, OS, browser, etc should be added within the dropdown below.)

<details>

```plaintext
datashader 0.18.2
macos
numba 0.61.2
```
</details>

```python
import datashader as dsh
import numpy as np
import xarray as xr

west = 3125000.0
south = 3250000.0
east = 4250000.0
north = 4375000.0

da = xr.DataArray(
    np.random.random((940, 940)),
    dims=("x", "y"),
    coords={
        "x": np.linspace(3123580.0, 4250380.0, 940),
        "y": np.linspace(4376200.0, 3249400.0, 940),
    },
    name="foo",
)
cvs = dsh.Canvas(256, 256, x_range=(west, east), y_range=(south, north))
# uncommenting this line makes it work
#da = da.assign_coords(dict(zip(("x", "y"), xr.broadcast(da["x"], da["y"]))))
# uncommenting this line makes it work, but with different results
#da = da.isel(y=slice(None, None, -1))
cvs.quadmesh(da.transpose("y", "x"), x="x", y="y")

```

#### Screenshots or screencasts of the bug in action

dcherian:
Disclaimer: generated by Claude, but seems right to me. That said, I don't really know the code 🤷🏾‍♂️ 

-----

Closes #1439

Fixes issue where quadmesh would return all NaN values when y coordinates are descending. The problem occurred in downsample_cpu and downsample_cuda functions where negative scale factors caused src_j0 > src_j1, resulting in empty ranges. Added checks to swap indices when scale is negative.

Includes regression test to prevent this issue from reoccurring.

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/datashader/tests/test_quadmesh.py
+++ b/datashader/tests/test_quadmesh.py
@@ -30,7 +30,7 @@ except ImportError:
 try:
     import cudf
     import cupy
-    array_modules.append(cupy)
+    array_modules.append(pytest.param(cupy, marks=pytest.mark.gpu))
 except ImportError:
     cudf = None
     cupy = None
@@ -916,4 +916,4 @@ def test_raster_quadmesh_descending_coords_2(array_module):
     cvs = ds.Canvas(256, 256, x_range=(west, east), y_range=(south, north))
     actual = cvs.quadmesh(da.transpose("y", "x"), x="x", y="y")
     expected = cvs.quadmesh(da.isel(y=slice(None, None, -1)).transpose("y", "x"), x="x", y="y")
-    np.testing.assert_allclose(expected, actual)
+    assert_eq_xr(expected, actual, close=True)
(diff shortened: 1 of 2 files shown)


datashader/tests/test_quadmesh.py
```diff
@@ -30,7 +30,7 @@ class dask:
 try:
     import cudf
     import cupy
-    array_modules.append(cupy)
+    array_modules.append(pytest.param(cupy, marks=pytest.mark.gpu))
```
hoxbro, line 33: (Not related to your PR)

PR #1442, opened 2025-08-28
Fix rendering viewport fully contained in one cell

### LINKED ISSUE #1438: QuadMeshRectilinear returning all NaN when plot_range is smaller than cell
(this library, plus any other relevant software, e.g. bokeh, python, notebook, OS, browser, etc should be added within the dropdown below.)

<details>

```plaintext
datashader 0.18.2
numba 0.61.2
macos
```
</details>

The code below ends up using the QuadMeshRectilinear path?

If I add broadcasted coordinates, then I get the expected result (the QuadMeshCurvilinear path).

```python
import datashader as dsh
import numpy as np
import xarray as xr

west = 111445.1872398071
south = 10018715.95288047
east = 111483.4057539478
north = 10018754.171394613

da = xr.DataArray(
    np.array(
        [
            [-0.4246922, -0.41608012, -0.40739873],
            [-0.4381327, -0.42964128, -0.42107907],
            [-0.45110574, -0.4427344, -0.43429095],
        ],
        dtype=np.float32,
    ),
    dims=("latitude", "longitude"),
    coords={
        "latitude": np.array([9936481.84638604, 10005932.69392196, 10076083.88816376]),
        "longitude": np.array([83489.61809496, 111319.49079327, 139149.36349159]),
    },
    name="foo",
)
cvs = dsh.Canvas(256, 256, x_range=(west, east), y_range=(south, north))
# uncommenting this line makes it work
# da = da.assign_coords(dict(zip(da.dims, xr.broadcast(da["latitude"], da["longitude"]))))
cvs.quadmesh(da, x="longitude", y="latitude")
```

#### Screenshots or screencasts of the bug in action

dcherian:
Disclaimer: generated by Claude, and this one I haven't spent the time to really wrap my head around it

-----

Closes #1438

The QuadMeshRectilinear glyph was incorrectly checking if coordinate break points were within canvas bounds, instead of checking if coordinate intervals overlap the canvas bounds. This caused quadmesh to return all NaN values when coordinate centers were outside the canvas but intervals still overlapped.

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/datashader/glyphs/quadmesh.py
+++ b/datashader/glyphs/quadmesh.py
@@ -193,10 +193,10 @@ class QuadMeshRectilinear(_QuadMeshLike):
             # An interval [xscaled[i], xscaled[i+1]] overlaps [0,1] if:
             # max(xscaled[i], xscaled[i+1]) >= 0 and min(xscaled[i], xscaled[i+1]) <= 1
             # This handles both ascending and descending coordinate orders
-            xinds = np.where((np.maximum(xscaled[1:], xscaled[:-1]) >= 0) &
-                           (np.minimum(xscaled[1:], xscaled[:-1]) <= 1))[0]
-            yinds = np.where((np.maximum(yscaled[1:], yscaled[:-1]) >= 0) &
-                           (np.minimum(yscaled[1:], yscaled[:-1]) <= 1))[0]
+            xin0, xin1 = xscaled >= 0, xscaled <= 1
+            yin0, yin1 = yscaled >= 0, yscaled <= 1
+            xinds = np.where((xin0[:-1] | xin0[1:]) & (xin1[:-1] | xin1[1:]))[0]
+            yinds = np.where((yin0[:-1] | yin0[1:]) & (yin1[:-1] | yin1[1:]))[0]
 
 
             if len(xinds) == 0 or len(yinds) == 0:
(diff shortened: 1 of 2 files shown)


dcherian:
I think I understand this failure and fix now; I've added a test for both ascending and descending coordinates.

datashader/glyphs/quadmesh.py
```diff
@@ -189,8 +189,16 @@ def extend(aggs, xr_ds, vt, bounds, x_breaks=None, y_breaks=None):
...
-            xinds = np.where((xscaled >= 0) & (xscaled <= 1))[0]
-            yinds = np.where((yscaled >= 0) & (yscaled <= 1))[0]
+            # Find intervals that overlap the canvas bounds [0,1]
+            # An interval [xscaled[i], xscaled[i+1]] overlaps [0,1] if:
+            # max(xscaled[i], xscaled[i+1]) >= 0 and min(xscaled[i], xscaled[i+1]) <= 1
+            # This handles both ascending and descending coordinate orders
+            xinds = np.where((np.maximum(xscaled[1:], xscaled[:-1]) >= 0) &
+                           (np.minimum(xscaled[1:], xscaled[:-1]) <= 1))[0]
+            yinds = np.where((np.maximum(yscaled[1:], yscaled[:-1]) >= 0) &
+                           (np.minimum(yscaled[1:], yscaled[:-1]) <= 1))[0]
```
hoxbro:
```suggestion
            xin0, xin1 = xscaled >= 0, xscaled <= 1
            yin0, yin1 = yscaled >= 0, yscaled <= 1
            xinds = np.where((xin0[:-1] | xin0[1:]) & (xin1[:-1] | xin1[1:]))[0]
            yinds = np.where((yin0[:-1] | yin0[1:]) & (yin1[:-1] | yin1[1:]))[0]
```

Would something like this work? This way, we avoid maximum/minimum calculations, which seem like heavier operations.

dcherian:
tests pass locally!

PR #1443, opened 2025-08-29
Use numba for faster estimation of curvilinear bounds

dcherian:
The first commit is the "cleanest" and gives ~4X speed up over the numpy version: https://github.com/holoviz/datashader/commit/59d09fddbec1e7103b8c8edc71f4a9fdd471b55b

The second commit avoids some useless data copying to give ~6X speed up over the numpy version: https://github.com/holoviz/datashader/commit/6f82702ecd11de16f46f2028729351daebce02ff

Benchmark script:
<details>

```python
#!/usr/bin/env python3
import sys
import os
import numpy as np
import timeit

from datashader.resampling import infer_interval_breaks, infer_interval_breaks_2d

if __name__ == "__main__":
    print("Testing Numba version correctness...")
    test_arr = np.arange(12.0).reshape(3, 4)
    print(f"Input array:\n{test_arr}")
    
    result_numpy = infer_interval_breaks(infer_interval_breaks(test_arr, axis=0), axis=1)
    print("NumPy result:")
    print(result_numpy)
    
    result_numba = infer_interval_breaks_2d(test_arr)
    print("Numba result:")
    print(result_numba)
    
    try:
        np.testing.assert_array_almost_equal(result_numpy, result_numba, decimal=10)
        print("✓ Numba version matches NumPy perfectly!")
    except AssertionError:
        print("✗ Results don't match!")
        print(f"Max difference: {np.max(np.abs(result_numpy - result_numba))}")

    sizes = [(500, 500), (1000, 1000), (5000, 5000)]
    for size in sizes:
        rows, cols = size
        print(f"\n--- Testing {rows}x{cols} array ---")

        test_array = np.arange(rows * cols, dtype=np.float64).reshape(rows, cols)
        _ = infer_interval_breaks_2d(test_array)
        
        def numpy_version():
            return infer_interval_breaks(infer_interval_breaks(test_array, axis=0), axis=1)
        
        def numba_version():
            return infer_interval_breaks_2d(test_array)
        
        iterations = 10 if rows <= 1000 else 3
        
        print(f"Running NumPy benchmark ({iterations} iterations)...")
        time_numpy = timeit.timeit(numpy_version, number=iterations)
        
        print(f"Running Numba benchmark ({iterations} iterations)...")
        time_numba = timeit.timeit(numba_version, number=iterations)
        
        print(f"\nMicrobenchmark Results ({rows}x{cols} array, {iterations} iterations):")
[...]

PR #1446, opened 2025-09-02
Interval bounds updates

dcherian:
1. Use `parallel=True` (this gets us to 8-9X over numpy vs 6X on the previous PR branch)
2. Support for CUDA
3. Property tests

Follow up to #1443


datashader/tests/test_quadmesh.py
```diff
@@ -811,3 +814,50 @@ def test_rectilinear_quadmesh_bbox_smaller_than_grid(array_module):
...
+        dtype=np.float64,
+        shape=st.tuples(
+            st.integers(min_value=2, max_value=50),
+            st.integers(min_value=2, max_value=50)
+        ),
+        elements=st.floats(
+            min_value=0.1,  # Positive spacings to ensure monotonic coordinates
+            max_value=10.0,
+            allow_nan=False,
+            allow_infinity=False
```
dcherian, line 830: does this seem right for the coordinate arrays?

datashader/glyphs/quadmesh.py
```diff
@@ -432,6 +432,10 @@ def compute_bounds_dask(self, xr_ds):
         return self.compute_x_bounds(xr_ds), self.compute_y_bounds(xr_ds)
 
     def infer_interval_breaks(self, centers):
+        if cupy and isinstance(centers, cupy.ndarray):
```
hoxbro: Should we also do this for `dask.Arrays`, or does the new function work with them?

pyproject.toml
```diff
@@ -48,7 +48,7 @@ Source = "https://github.com/holoviz/datashader"
 HoloViz = "https://holoviz.org/"
 
 [project.optional-dependencies]
-tests = ["pytest"]
+tests = ["pytest", "hypothesis"]
```
hoxbro, line 51: Also needed to add to `pixi.toml`, I would add it to `feature.test-core.dependencies`

datashader/tests/test_quadmesh.py
hoxbro: I think so.

datashader/glyphs/quadmesh.py
dcherian: It will eagerly compute them; but now i've sent dask arrays through the old code path; nominally we can add support but i don't fully understand how those coordinates are handled

.github/workflows/test.yaml
```diff
@@ -121,6 +121,13 @@ jobs:
       - uses: holoviz-dev/holoviz_tasks/pixi_install@v0
         with:
           environments: ${{ matrix.environment }}
+      - name: Restore cached hypothesis directory
+        uses: actions/cache@v4
+        with:
+          path: .hypothesis/
```
hoxbro, line 127: Curious, why is this needed? Do you have a link where I can read about it?

dcherian: hypothesis saves failing examples to a "database" and then retries those examples in future runs. So if CI detects a failure, hypothesis will write to the database, and we want to cache and restore it.

They've redone the docs and I can't find a good explanation anymore :( but see
1.https://hypothesis.readthedocs.io/en/latest/tutorial/replaying-failures.html
2. https://hypothesis.readthedocs.io/en/latest/reference/api.html#database 

Seems like they have a vastly more complex solution now:
https://hypothesis.readthedocs.io/en/latest/reference/api.html#hypothesis.database.GitHubArtifactDatabase

(I iterated on this stuff a year and a half ago, and copied over the code from Xarray's CI)

hoxbro: Thank you for the detailed info :+1:

PR #1448, opened 2025-09-04
fix: Improve `compute_chunksize` for 

hoxbro:
Previously, the following code would, depending on the dask version, either kill the process or kill my computer. 

I haven't done any profiling to see if this affects performance, but at least it doesn't crash my computer anymore. If it does, we can maybe move the functionality into `resample_2d_distributed`.

``` python
import numpy as np
import dask.array as da
import datashader as ds
import xarray as xr
import dask

print(dask.__version__)

# create large dask array
N = 100_000
dask_array = da.random.random((N, N), chunks=(1000, 1000))  # .compute()
# convert to dasked xarray
dask_xarray = xr.DataArray(
    dask_array,
    dims=["x", "y"],
    coords={"x": np.arange(N), "y": np.arange(N)},
    name="example_data",  # Name of the data variable
)
# create plot using Datashader
arr = ds.Canvas(plot_height=300, plot_width=300).raster(dask_xarray)
arr.compute()
```

PR #1450, opened 2025-09-05
test: Add benchmark to CI

hoxbro:
I think it is unfortunate that the benchmark script is never run (not in CI and likely very few times locally). This moves the benchmark from `pytest-benchmark` to `pytest-codspeed`, which will give us feedback directly in PRs. 

I hope this can help with improving the quadmesh performance, and I hope it shows no performance regressions in https://github.com/holoviz/datashader/pull/1448.


datashader/tests/benchmarks/__init__.py
```diff
@@ -1,3 +1,26 @@
+import itertools
 import pytest
+import numpy as np
 
 pytestmark = pytest.mark.benchmark
+
+array_modules = [np]
+sizes = [256, 512, 1024]
```
hoxbro: Not sure if we should add more sizes here.

hoxbro: Added more: [256, 512, 1024, 2048, 4096, 8192]

datashader/tests/benchmarks/test_quadmesh.py
hoxbro, line 1: @dcherian, haven't looked over all your benchmark scripts, but if you have some more benchmark scripts, I would love to add them to this PR.

datashader/tests/benchmarks/__init__.py
hoxbro: Maybe this is taking too long with this many configurations.

hoxbro: I now keep the sizes but only run with numpy.

datashader/tests/benchmarks/test_quadmesh.py
```diff
@@ -0,0 +1,41 @@
...
+    data = data.assign_coords({"lon": lon_coord, "lat": lat_coord})
+
+    return data, (west, east), (south, north)
+
+
+@pytest.mark.benchmark(group="quadmesh")
+def test_quadmesh(benchmark, quadmesh_data):
+    def func():
+        data, x_range, y_range = quadmesh_data
+        cvs = ds.Canvas(256, 256, x_range=x_range, y_range=y_range)
```
dcherian:
```suggestion
        cvs = ds.Canvas(1024, 1024, x_range=x_range, y_range=y_range)
```

We should get some upsampling cases too

datashader/tests/benchmarks/test_quadmesh.py
```diff
@@ -0,0 +1,41 @@
...
+    )
+    data = data.isel(y=slice(None, None, -1))
+    lon_coord, lat_coord = xr.broadcast(data.x, data.y)
+    data = data.assign_coords({"lon": lon_coord, "lat": lat_coord})
+
+    return data, (west, east), (south, north)
+
+
+@pytest.mark.benchmark(group="quadmesh")
+def test_quadmesh(benchmark, quadmesh_data):
```
dcherian:
```suggestion
def test_quadmesh_curvilinear(benchmark, quadmesh_data):
```

datashader/tests/benchmarks/test_raster.py
```diff
@@ -0,0 +1,30 @@
...
+
+
+sizes = [256, 512, 1024, 2048, 4096, 8192]
+
+
+@pytest.fixture(params=sizes)
+def raster_data(request, rng):
+    size = request.param
+    data = xr.DataArray(
+        rng.random((size, size)),
```
hoxbro: Note to self: convert back to dask for #1448

datashader/tests/benchmarks/test_quadmesh.py
dcherian: See #1451

hoxbro: Perfect, just ping me when it is ready to merge into this one (if it isn't already).

PR #1457, opened 2025-09-10
test(benchmark): Add seed and make dask single-threaded

hoxbro:
It could be the cause of this flakiness.


dcherian:
WDYM by flakiness? IME running benchmark measurements on github actions machines is basically noise below 50%-ish. https://labs.quansight.org/blog/github-actions-benchmarks

PR #1459, opened 2025-09-12
Fix bug in QuadMeshRectilinear

dcherian:
Fixes a rendering bug in QuadmeshRectilinear.  I think this was introduced in https://github.com/holoviz/datashader/pull/1442

### before (notice the thin lines)

image

### after
image

### DIFF
--- a/datashader/glyphs/quadmesh.py
+++ b/datashader/glyphs/quadmesh.py
@@ -218,26 +218,13 @@ class QuadMeshRectilinear(_QuadMeshLike):
                 yscaled = yscaled[::-1]
                 xr_ds = xr_ds.isel({y_name: slice(None, None, -1)})
 
-            xm0 = max(np.searchsorted(xscaled, 0, 'right') - 1, 0)
+            xm0 = max(np.searchsorted(xscaled, 0, "right") - 1, 0)
             xm1 = np.searchsorted(xscaled, 1, "left")
-            ym0 = max(np.searchsorted(yscaled, 0, 'right') - 1, 0)
+            ym0 = max(np.searchsorted(yscaled, 0, "right") - 1, 0)
             ym1 = np.searchsorted(yscaled, 1, "left")
 
-            if xm0 == xm1:
-                xm1 += 1
-            if ym0 == ym1:
-                ym1 += 1
-
-            # xin0, xin1 = xscaled >= 0, xscaled <= 1
-            # yin0, yin1 = yscaled >= 0, yscaled <= 1
-            # xinds, = np.where((xin0[:-1] | xin0[1:]) & (xin1[:-1] | xin1[1:]))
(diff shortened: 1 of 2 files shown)


hoxbro:
Have a hard time wrapping my head around this. Can you help me understand why the `xm1` and `ym1` logic is wrong? 

Do you have some examples where your diff passes and my diff fails?  

Also, wouldn't there be a case for having decreasing values for the x-axis?

datashader/glyphs/quadmesh.py
```diff
@@ -220,18 +220,18 @@ def extend(aggs, xr_ds, vt, bounds, x_breaks=None, y_breaks=None):
                 # Nothing to do
                 return
 
-            xm0, xm1 = max(xinds.min() - 1, 0), xinds.max() + 1
-            ym0, ym1 = max(yinds.min() - 1, 0), yinds.max() + 1
+            xm0, xm1 = xinds.min(), xinds.max() + 2
```
hoxbro: I think we should add a comment here. 

My reasoning for this change is that when we look at the original implementation, we wanted the value just before the minimum and the value right after the maximum index (an assumption on my part). When we did the diff-like operation in #1442, we needed to add one to the minimum and maximum found. This is what is currently written: `(min - 1 + 1, max + 1 + 1)`. 

Though the removal of `+1` in the slice below, I'm a bit iffy about. I think it has been done to include the last index in a slice. The tests pass both with and without it... Need to think a bit more about it.

dcherian: yes we want the last cell to the left of `xscaled=0` and the first cell to the right of `xscaled=1`. We also need to handle the case where a cell edge lands on 0, 1. That's probably the test case that's missing.

This is why I found the `searchsorted` approach easier to reason about.

>  I think it has been done to include the last index in a slice.

:+1:

hoxbro: Let's add the +1 in the slice back. 

> That's probably the test case that's missing.

Likely, do you have a rough idea of how to create the test for this?

dcherian: Yeah this is indeed the problem:

If the canvas lines up exactly with cell edges; we grab an extra cell on either side, so now if the last cell within the canvas on either side is all NaN but cell right next to them has data; we'd get a stripe just like the image above. I'll push a test case.
```
(Pdb) xscaled[xm0:xm1]
array([-0.54545455,  0.        ,  0.5       ,  1.        ,  1.5       ])
(Pdb) 
array([-0.54545455,  0.        ,  0.5       ,  1.        ,  1.5       ])
(Pdb) yscaled[ym0:ym1]
array([-0.33333333,  0.        ,  0.33333333,  0.66666667,  1.        ,
        1.33333333])
```

dcherian: I pushed a test and reverted back to the searchsorted approach. I left the old approach in if you wanted to experiment.

PR #1460, opened 2025-09-15
test: Merge hypothesis cache

hoxbro:
Merge all the cache together in a separate job to avoid:

image

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -126,6 +126,7 @@ jobs:
         with:
           path: .hypothesis/
           key: cache-hypothesis-*
+          enableCrossOsArchive: true
       - name: Test Unit
         run: |
           pixi run -e ${{ matrix.environment }} test-unit $COV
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yaml
```diff
@@ -139,11 +140,32 @@ jobs:
         with:
           token: ${{ secrets.CODECOV_TOKEN }}
       - name: Save hypothesis directory
+        uses: actions/cache/save@v4
+        if: always()
+        with:
+          path: .hypothesis/
+          key: cache-hypothesis-${{ github.run_id }}-${{ github.run_attempt }}-${{ matrix.environment }}-${{ matrix.os }}
```
dcherian: sorry i forgot why this was needed and deleted it when porting to datashader.

This is what we do in xarray
https://github.com/pydata/xarray/blob/ccb34374564de32c8df305285a7233dd892cdce5/.github/workflows/hypothesis.yaml#L98-L105

PR #1462, opened 2025-09-26
chore: replace direct download of nyc_taxi with hvsampledata

Azaya89:
This PR replaces the direct download of the large `nyc_taxi` data with the version from hvsampledata.

**Testing with an editable install of hvsampledata**
```python
import hvsampledata as hvs
import datashader as ds

df = hvs.nyc_taxi_remote("pandas", engine_kwargs={"columns": ['dropoff_x', 'dropoff_y']})
print(df.head())

agg = ds.Canvas().points(df, 'dropoff_x', 'dropoff_y')
print(agg.shape)
```
```
   dropoff_x  dropoff_y
0 -8234835.5  4975627.0
1 -8237020.5  4976875.0
2 -8232279.0  4986477.0
3 -8238124.0  4971127.0
4 -8238107.5  4974457.0
(600, 600)
```


datashader/__init__.py
```diff
@@ -17,19 +17,18 @@
 if Version(pandas_version) >= Version('0.24.0'):
     from . import datatypes  # noqa (API import)
 
-# make pyct's example/data commands available if possible
+# make pyct's example commands available if possible
 from functools import partial
 try:
-    from pyct.cmd import copy_examples as _copy, fetch_data as _fetch, examples as _examples
```
hoxbro, line 23: Completely remove pyct.

examples/getting_started/1_Introduction.ipynb
```diff
@@ -24,22 +24,13 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "import datashader as ds, pandas as pd, colorcet as cc\n",
+    "import datashader as ds, colorcet as cc\n",
+    "import hvsampledata as hvs\n",
     "\n",
-    "df = pd.read_csv('../data/nyc_taxi.csv', usecols=['dropoff_x', 'dropoff_y'])\n",
+    "df = hvs.nyc_taxi(\"pandas\", engine_kwargs={\"columns\": ['dropoff_x', 'dropoff_y']})\n",
```
hoxbro:
```suggestion
    "df = hvs.nyc_taxi_remote(\"pandas\", engine_kwargs={\"columns\": ['dropoff_x', 'dropoff_y']})\n",
```

Also, mention that this is the first time this cell is run.

pixi.toml
```diff
@@ -38,6 +38,7 @@ scipy = "*"
 setuptools = "*" # distutils for pyct
 toolz = "*"
 xarray = "*"
+fastparquet = "*"
```
hoxbro: Please check that this has not indirectly pinned anything like pandas.

scripts/download_data.py
```diff
@@ -1,11 +1,6 @@
 from contextlib import suppress
-
-import pyct.cmd
 from packaging.version import Version
 
-pyct.cmd.fetch_data(name="data", path="examples", datasets="datasets.yml")
```
hoxbro, line 6: You should add the hvsampledata command here.

datashader/__init__.py
```diff
@@ -11,6 +13,18 @@
 from . import transfer_functions as tf                   # noqa (API import)
 from . import data_libraries                             # noqa (API import)
 
+with suppress(ImportError):
```
hoxbro: This is not the right place it should be in __main__.py.

Azaya89: OK!

pyproject.toml
```diff
@@ -34,7 +34,6 @@ dependencies = [
     'numpy',
     'pandas',
     'param',
-    'pyct',
```
maximlt, line 37: I don't think pyct should be removed yet from the dependencies.

Azaya89: It's not removed totally. Just moved to optional dependencies.

maximlt: With these changes if I install datashader and run `datashader example` it breaks. So that's a breaking change. The goal is to deprecate it so people get a chance to stop using `datashader example`. Then `pyct` can be removed entirely from the dependencies.

scripts/download_data.py
```diff
@@ -21,4 +16,51 @@
 
     gds.get_path("geoda.natregimes")
     gds.get_path("nybb")
-    gds.get_path('geoda health')
+    gds.get_path("geoda health")
+
+
+with suppress(ImportError):
```
hoxbro, line 28:
``` python
hvs.nyc_taxi_remote("download-only")  # return None or path
```

I think we should add something like this in hvsampledata to avoid all of this boilerplate. This will likely also make it so we can move pyarrow back in `pixi.toml`

pyproject.toml
```diff
@@ -119,4 +119,6 @@ filterwarnings = [
     "ignore:The 'shapely.geos' module is deprecated, and will be removed in a future version:DeprecationWarning",
     # 2025-09
     "ignore:Signature .* for <class 'numpy.longdouble'>.*:UserWarning",
+    # 2025-10
+    "ignore:The 'pyct' package bundled as a datashader dependency is deprecated since version 0.19 and will be removed in version 0.20.*" # https://github.com/holoviz/datashader/pull/1462
```
hoxbro: I think we should remove this filtering. Running datashader in Python (and not CLI) should never show this warning.

scripts/download_data.py
maximlt: What is the goal of this code?

hoxbro: The goal of the script is to pre-download the data, so it doesn't fail in CI tests later on, and to avoid download bars in examples.

The goal of the code was likely to download it, without the latest hvsampledata.

examples/getting_started/1_Introduction.ipynb
```diff
@@ -25,9 +25,10 @@
    "outputs": [],
    "source": [
     "import datashader as ds, colorcet as cc\n",
-    "import hvsampledata as hvs\n",
+    "import hvsampledata as hvs, pandas as pd\n",
     "\n",
-    "df = hvs.nyc_taxi_remote(\"pandas\", engine_kwargs={\"columns\": ['dropoff_x', 'dropoff_y']})\n",
```
hoxbro: Let us keep this.

pixi.toml
```diff
@@ -38,6 +38,7 @@ scipy = "*"
 setuptools = "*" # distutils for pyct
 toolz = "*"
 xarray = "*"
+pyarrow = "*"
```
hoxbro: This should be added back to where it previously was.

scripts/download_data.py
```diff
@@ -1,9 +1,12 @@
...
-import pyct.cmd
 from packaging.version import Version
 
-pyct.cmd.fetch_data(name="data", path="examples", datasets="datasets.yml")
+with suppress(ImportError):
+    import pyct.cmd
+    from datashader import _warn_pyct_deprecated
+
+    _warn_pyct_deprecated(stacklevel=1)
+    pyct.cmd.fetch_data(name="data", path="examples", datasets="datasets.yml")
```
hoxbro, line 7: The scripts directory is for internal use only; please remove this.

datashader/__init__.py
```diff
@@ -18,12 +18,35 @@
...
 
 # make pyct's example/data commands available if possible
-from functools import partial
+import warnings
+from functools import partial, wraps
+
+
+def _warn_pyct_deprecated(stacklevel=2):
+    warnings.warn(
+        "The 'pyct' package is deprecated since version 0.19 "
```
hoxbro: If a user saw this warning, do you think they would understand it? 

I will also move warnings import into the function.

datashader/__init__.py
```diff
@@ -18,12 +18,35 @@
...
+        category=FutureWarning,
+        stacklevel=stacklevel,
+    )
+
+
+def _deprecated_pyct_wrapper(func):
+    """Wrapper to add deprecation warning to pyct functions."""
+    @wraps(func)
+    def wrapper(*args, **kwargs):
+        _warn_pyct_deprecated(stacklevel=2)
```
hoxbro: I'm not sure if these stack levels are set correctly, but they are likely still emitted because the category is a FutureWarning. 

You can test this out by setting the `CategoryWarning` to `DeprecationWarning`.

datashader/__init__.py
Azaya89: > If a user saw this warning, do you think they would understand it?

Reads simple enough for me, yeah.

> I will also move warnings import into the function.

OK.

hoxbro: Let me rephrase it: what does the pyct package do, and why is it affecting the deprecated functions? Without looking at the code, it seems very confusing.

datashader/__init__.py
Azaya89: You're right. Thanks. The correct stacklevel is 3 in this instance. I tested it via
```python
import datashader as ds

ds.fetch_data("datashader")
```

datashader/__init__.py
Azaya89: Hmm, OK I get it now. A user wouldn't have used `pyct` directly and would have been confused seeing the mention of it in the message. Will rephrase to mention the actual functions being called.

scripts/download_data.py
hoxbro: You did not remove this...

examples/getting_started/1_Introduction.ipynb
```diff
@@ -35,9 +36,8 @@
    "metadata": {},
    "source": [
     ":::{note}\n",
-    "The file `nyc_taxi.csv` used above is located at\n",
-    "[nyc_taxi.csv](https://github.com/holoviz/datashader/blob/main/examples/data/.data_stubs/nyc_taxi.csv) in the Datashader repository. When running this example, make sure the file is available locally and update the path accordingly.\n",
-    ":::\n"
+    "The first time this cell is run, it may take some time for the NYC taxi dataset to be downloaded remotely (about 260MB).\n",
```
hoxbro:
```suggestion
    "The first time this cell is run, it will download the NYC taxi dataset (about 260MB).\n",
```

datashader/__init__.py
```diff
@@ -18,12 +18,36 @@
     from . import datatypes  # noqa (API import)
 
 # make pyct's example/data commands available if possible
-from functools import partial
+from functools import partial, wraps
```
hoxbro, line 21: Let's `del wraps` as we did with the others.

datashader/__main__.py
```diff
@@ -1,4 +1,12 @@
...
+
+
 def main(args=None):
+    with suppress(ImportError):
+        import pyct  # noqa: F401
+
+        from . import _warn_pyct_deprecated
+        _warn_pyct_deprecated(stacklevel=3)
     try:
         import pyct.cmd
```
hoxbro:
```suggestion
    try:
        import pyct.cmd
        from . import _warn_pyct_deprecated
        _warn_pyct_deprecated(stacklevel=3)
```

pixi.toml
```diff
@@ -80,6 +80,7 @@ scikit-image = "*"
 shapely = ">=2.0.0"
 spatialpandas = "*"
 streamz = "*"
+hvsampledata = ">=0.1.4"
```
hoxbro:
```suggestion
hvsampledata = ">=0.1.5a3"
```

Let's use a dev version to clearly indicate which version works here. We can remove the dev version once 0.1.5 is released.

datashader/__init__.py
```diff
@@ -18,12 +18,36 @@
...
+
+
+def _warn_pyct_deprecated(stacklevel=2):
+    import warnings
+
+    warnings.warn(
+        "The 'fetch_data()', 'copy_examples()', and 'examples()' functions are "
+        "deprecated since version 0.19 and will be removed in version 0.20. "
```
hoxbro, line 29: With the release candidacy of datashader, I feel like 0.20 would be released at least 6 months after 0.19.

Seems like each minor release has been a year in between (ignoring 0.17.0, which did not correctly drop a Python version).

Azaya89: Ok, so do you suggest we remove it in 0.19.xx, and modify the message to that effect?

hoxbro: No. I suggest 0.20, and not 0.21. We would _never_ drop something in a patch release.

Azaya89: OK, the message is correct then. It says 0.20 already.

hoxbro: Yes, but still want to highlight it as we normally do two minor releases.

maximlt: FYI, as per HEP2, there's no requirement to indicate the removal version:

image


https://holoviz.org/about/heps/hep2.html is worth a read if you haven't done it yet.

maximlt (approved):

PR #1495, opened 2026-04-21
fix: Improve edge detection for polygons

### LINKED ISSUE #1494: error when rendering polygons
This is a Claude-derived reproducer of a bug I encountered when trying to render healpix cells. Notice how the left edge is missing 2 cells for both polygons

Image

Here is claude's claimed diagnosis:
> Pixel centers that fall exactly on a polygon edge are left unfilled: the scan-line winding-number algorithm in _build_draw_polygon cancels the on-edge contribution against the opposite edge's contribution. Shows up as transparent seams when rendering HEALPix cells (polygons meeting at a shared vertex whose incident edges lie on pixel-center lines).

<details>

```plaintext
[Include version information here](datashader: main)
```
</details>

```python
def test_polygons_shared_vertex_no_seam_gap():
    r"""Two diamonds meeting at (0, 0) should cover every interior pixel.

    With this 20x10 canvas the shared vertex lands at canvas x=9.5, so column
    xi=10 sits on the right diamond's lower edge (0,0)->(1,-1). The rasterizer
    leaves that column NaN instead of claiming it for the right diamond.

    Geometry (data coords, y up)::

             (-1, 1)     (1, 1)
                *           *
               / \         / \
              /   \       /   \
             /     \     /     \
            *       * = *       *
         (-2, 0)  (0, 0)      (2, 0)
                  shared
             \     /     \     /
              \   /       \   /
               \ /         \ /
                *           *
             (-1,-1)     (1,-1)

    Rendered (20x10 canvas, middle rows; '.' = NaN, '1'/'2' = diamond id)::

        col:  0         1         2
              0123456789012345678901
        row4: .111111111.222222222.   <- xi=10 is the seam gap (bug)
        row5: .111111111.222222222.
    """
    df = GeoDataFrame({
        "polygons": pd.Series(
            [
                [[[-2.0, 0.0, -1.0, -1.0, 0.0, 0.0, -1.0, 1.0, -2.0, 0.0]]],
                [[[0.0, 0.0, 1.0, -1.0, 2.0, 0.0, 1.0, 1.0, 0.0, 0.0]]],
            ],
            dtype="MultiPolygon[float64]",
        ),
        "v": [1.0, 2.0],
    })
    canvas = ds.Canvas(
        plot_width=20, plot_height=10, x_range=(-2.0, 2.0), y_range=(-1.0, 1.0)
    )
    arr = canvas.polygons(df, geometry="polygons", agg=ds.mean("v")).values

    # Middle rows cross the widest part of each diamond; every interior column
    # should be filled.
    for yi in (4, 5):
        row = arr[yi]
        missing = np.where(np.isnan(row[1:19]))[0]
        assert missing.size == 0, (
            f"row {yi} has NaN at columns {missing + 1}: {row!r}"
        )
```

hoxbro:
## Description

Fixes #1494

The following test has been changed because of this change, see my comments below for my reasoning to why this is okay. I don't think we can expect non horizontal and vertical lines to not have overlap when we are working in finite space (somewhat like the [staircase paradox](https://en.wikipedia.org/wiki/Staircase_paradox)). The changes also make the output of the test 90 deg rotational symmetric.

``` 
datashader/tests/test_polygons.py::test_no_overlap[GeoDataFrame]
datashader/tests/test_polygons.py::test_no_overlap[dask_GeoDataFrame]
datashader/tests/test_polygons.py::test_no_overlap_agg[GeoDataFrame]
datashader/tests/test_polygons.py::test_no_overlap_agg[dask_GeoDataFrame]
```

## How Has This Been Tested?

CI + Reasoning about failing test locally

## AI Disclosure

- [x] This PR contains AI-generated content.
  - [x] I have tested all AI-generated content in my PR.
  - [x] I take responsibility for all AI-generated content in my PR.

Tools and Models: Claude Code + Opus 4.6. Redirected it to the function, asked for it to find an off by one error in `draw_polygon`

### AI Summary
Change: `yincreasing[ei]` → `yincreasing[ei] > 0` at line 224.

Why: Original tiebreak count on-edge case for any non-zero yincreasing (both +1 and -1 truthy). Cause asymmetry — left edge (yinc=-1) and right edge (yinc=+1) of CCW polygon both "count" when point on edge. At shared-vertex between two touching polygons, two edges meet at same pixel column, contributions cancel → winding=0 → NaN column.

How to apply: Convention now "include right boundary, exclude left" (CCW: yinc=+1 = right edge). On-edge pixel claimed exactly once by polygon to its left. Touching polygons at shared vertex → left polygon claim the column.

Quadmesh.py use stricter `bxa * yincreasing < 0` (no on-edge at all) — polygon.py need on-edge inclusion for pixel-grid-aligned vertex (e.g. unit-diamond lower tip), so need tiebreak, but now direction-consistent.

## Checklist

- [x] Tests added and are passing


hoxbro:
Looking at one of the failing tests example

``` python
import holoviews as hv
import spatialpandas as spd
import pandas as pd
from holoviews.operation.datashader import rasterize

hv.extension("bokeh")

df = spd.GeoDataFrame(
    {
        "polygons": pd.Series(
            [
                [[1, 1, 2, 2, 1, 3, 0, 2, 1, 1],[0.5, 1.5, 0.5, 2.5, 1.5, 2.5, 1.5, 1.5, 0.5, 1.5]],
                # [[0.5, 1.5, 1.5, 1.5, 1.5, 2.5, 0.5, 2.5, 0.5, 1.5]],
                # [[0, 1, 2, 1, 2, 3, 0, 3, 0, 1, 1, 1, 0, 2, 1, 3, 2, 2, 1, 1]],
            ],
            dtype="Polygon[float64]",
        ),
    }
)

poly = hv.Polygons(df).opts(alpha=0.4)
rpoly = rasterize(poly, dynamic=False, width=16, height=16).opts(alpha=0.4)
(poly * rpoly).opts(width=800, height=800)
```

This is how it looks on main:
main

And on this branch:

branch

datashader/tests/test_geopandas.py
```diff
@@ -86,26 +85,26 @@
 
 
 nybb_polygons_sol = np.array([
```
hoxbro, line 87: Same data as before.

datashader/tests/test_geopandas.py
```diff
@@ -35,28 +34,28 @@
 except ImportError:
     spatialpandas = None
 
-
+_ = np.nan  # done for visual clarity in test data
 nybb_lines_sol = np.array([
```
hoxbro, line 38: Same data as before

hoxbro:
The other failing test shows similar inconsistency. My take is this is the branch has the correct behavior and overlapping is expected when working with finite precision and lines which has angle which is not horizontal or vertical. 

``` python
import holoviews as hv
import spatialpandas as spd
import pandas as pd
import datashader as ds
from holoviews.operation.datashader import rasterize

hv.extension("bokeh")

df = spd.GeoDataFrame({
    'polygons': pd.Series([
       [[1, 1, 2, 2, 1, 3, 0, 2, 1, 1], [0.5, 1.5, 0.5, 2.5, 1.5, 2.5, 1.5, 1.5, 0.5, 1.5]],
       [[0.5, 1.5, 1.5, 1.5, 1.5, 2.5, 0.5, 2.5, 0.5, 1.5]],
       [[0, 1, 2, 1, 2, 3, 0, 3, 0, 1, 1, 1, 0, 2, 1, 3, 2, 2, 1, 1]]
    ], dtype='Polygon[float64]'),
    'v': range(3)
})

poly = hv.Polygons(df, vdims="v").opts(alpha=0.4)
rpoly = rasterize(poly, dynamic=False, width=16, height=16, aggregator=ds.sum("v")).opts(alpha=0.4)
(poly * rpoly).opts(width=800, height=800)
```
main
main

this branch
branch

dcherian:
Certainly looks better to me!

dcherian:
Can we get this merged and released please? It's a great improvement for my use cases

jbednar (commented):
Thanks for looking into this. I can't tell which branch is better by looking at the code, but the plots definitely seem like better behavior to me, and apparently it addresses the original complaint. 

I do wonder about the convention of including the right edge rather than the left edge as the AI is reporting:

> How to apply: Convention now "include right boundary, exclude left" (CCW: yinc=+1 = right edge). On-edge pixel claimed exactly once by polygon to its left. Touching polygons at shared vertex → left polygon claim the column.

For regularly gridded datasets in [HoloViews](https://holoviews.org/user_guide/Continuous_Coordinates.html), we are very explicit about including the left and lower edge and excluding the upper and right edge, to avoid double counting:

> Indexing in continuous coordinates always returns the value closest to the requested value, given the available resolution. Note that in the case of coordinates truly on the boundary between array elements (as for -0.2,0.4), the bounds of each array cell are taken as right exclusive and upper exclusive, and so (-0.2,0.4) returns array index (3,0).

That convention dates back to at least 2003 (in Topographica) and matches conventions in Python array indexing, so I worry about having the opposite convention for polygons.

> Merging this PR will degrade performance by 23.46%

I'm guessing that report is an error? I don't see why changing this line of code would affect the graph layout code or quadmesh code, which seem like they would be in  entirely different parts of the codebase.

hoxbro:
> I do wonder about the convention of including the right edge rather

This is only for overlapping behavior, where a pixel center falls exactly on a polygon edge. Before two boundary edges would cancel each other out, we now chose the right edge as also defined in the comments below.

> I'm guessing that report is an error?

Yes, these test are very flaky.

Issue #1496, opened 2026-04-28
Add support for narwhals eager backends


cakiki:
First step in supporting Arrow tables in general, and any container that uses arrow tables in the background.

edit: renamed the issue to narrow the scope to narwhals eager backends like discussed on discord.

PR #1502, opened 2026-05-26
Add support fo eager Narwhals dataframes

### LINKED ISSUE #1199: Support for polars dataframes
As [Polars](https://github.com/pola-rs/polars) is becoming more and more popular I am wondering if there are any consideration to support its dataframes. Given the performance oriented focus of this package, it might be solid contender.

cakiki:
This is a first attempt to add support for narwhals eager dataframes.

I tried as best I could to copy the pandas path.

Closes #1199 and #1496


cakiki:
@hoxbro this is ready I think

PR #1505, opened 2026-06-14
Add missing docstrings to reductions

### LINKED ISSUE #1422: Add docstrings to all the reductions
In https://github.com/holoviz/datashader/pull/1420 I've attempted to add a few missing reductions to the API ref. This didn't work for all of them and I assume that it is because some (like `max_n`) do not have a docstring.

cakiki:
Closes #1422


cakiki:
cc @hoxbro

hoxbro (approved):

PR #1507, opened 2026-07-18
Fix segment_end clip propagation; reproduce direction-dependent antialiasing bug in Canvas.line

SiggyF:
## Summary

- Propagates `clipped_end` into `segment_end` in `draw_segment` (`datashader/glyphs/line.py`), mirroring the existing `segment_start = segment_start or clipped_start`. Without this, `_full_antialias` treats a canvas-edge-clipped endpoint as a midpoint (intensity ~0.5) instead of an end cap (intensity ~0.293), over-brightening pixels at tile/canvas boundaries in tiled rendering pipelines.
- Adds `datashader/tests/test_line_direction_invariance.py`, reproducing a **related but distinct** bug found while validating the fix above against real-world tiled rendering, filed separately as #1508: rendering the same physical 2-point line segment gives a different result depending on which endpoint is listed first, even with no clipping involved on an interior endpoint.

See #1508 for full details on the direction-dependence bug, root cause (`flip_order` in `_full_antialias`, ~line 863), and the minimal repro.

## Context

Surfaced via [ais-shader](https://github.com/TUDelft-CITG/ais-shader), a tiled AIS vessel-traffic heatmap renderer, where it caused a visible "seam" of inflated `transit_count`/mean-speed values at tile boundaries -- neighboring tiles render the same boundary-crossing segments, but AIS-ping chronological digitization order differs arbitrarily per segment, so the direction-dependent discrepancy doesn't cancel out across the tile mosaic.

## Test plan

- [x] Manual verification of the `segment_end` clip-propagation fix (see commit `84c12e3`)
- [x] New `test_line_direction_invariance.py` added, documents current failing behavior (52 tests fail, demonstrating the #1508 bug) plus the canonicalization workaround (passing)
- [ ] Would appreciate maintainer input on whether the `flip_order` branch itself should be fixed for numerical symmetry, or whether canonicalizing endpoint order at a higher level (e.g. in `Canvas.line`) is preferred for #1508


SiggyF:
Still WIP I will submit this again later.

